PR #2, opened 2020-03-10
Woutermeek/upstream

woutermeek:


pwielders (approved):
Tested on Linux and Windows.

PR #14, opened 2020-03-13
[core] corrected the xutility inclusion.

anjalirajan:


Source/core/Rectangle.h
```diff
@@ -21,7 +21,9 @@
 #define __RECTANGLE_H
 
 #include <algorithm>
+#ifdef WIN32
```
MFransen69: Didn't try to build it but had a quick check at the code, is the xutility needed anyway for the windows build?

MFransen69 (changes requested):

Source/core/Rectangle.h
pwielders: Also since the windows build is capable of running 32 bits and 64 bits, make sure to use the proper pre-processor statement (__WINDOWS__) unless this header file is only required on 32 Bits builds and not on 64 buts, which I doubt :-).

pwielders (approved):
Removal of lines of code is always good :-)

MFransen69 (approved):

PR #23, opened 2020-03-24
Development/automated proxy stub generation

pwielders:
1) Cached Release, with this release we cater for all possible scenarios, which led to a potentially "Bartjes" case. The rewrite of this clear showed some race conditions which in turn required some code generator rewrites. Getting the ProxyStub under the hood maintenance object had to do the ref count for that interface in 1 go. In the old generated code this was done in two steps. Issue resolved now!
2) Using COMRPC between multiple platforms revealed a portability issue. Here we where communicating between a 64bit platform and a 32bit platform. As you can image if I compile the COMRPC code on a 32 bit platform, the space for the “actual implementation address” of the 64bit platform did not fit and let to issues. Hence why now you can define the biggest denominator for the interoperability of the platform. In case of 32/64 bits platforms this would be 64. This allows the testcase COMRPC build on you 64 bits laptop to communicate with the RPI platform to work properly. This Infrastructure is now part of the COMRPC support library AND the code generators.
3) There where still some manually generated P/S in the system the transition work was already done but this was not yet properly tested. These where still manually generated because:
A: they where typeset templates (IStringIterator,IValueIterator)
B: not all methods (IShell) run over a Process Boundary s as NDS this could not generate code for a method. Needed a new keyword.
4) As debugging under Windows is still my favourite environment to tackle race conditions/deadlocks we had to make some minor modifications to PS generator and code.


Source/com/IUnknown.h
```diff
@@ -334,6 +334,9 @@ namespace ProxyStub {
 
             _adminLock.Unlock();
 
+            if (result != Core::ERROR_NONE) {
+                delete &_parent;
```
MFransen69, line 338: Couldn't we better make _parent a pointer instead of a reference? Makes the ownership more clear and then we could also set it to nullptr here

MFransen69 (commented):
See remark

Source/com/IUnknown.h
pwielders: Nope, This is a linked through reference counted object (composit of the parent). If it deletes the parent it deletes this one as well. Lifetime between the two objects (as this is a composit of the parent) are equal. Do not want to rasie the suggestion that it is an aggregate or that the two can live without each other. So the reference has exactly the same lifetime as the parent (you can not event construct this object without the parent)

MFransen69: Clear!

MFransen69 (approved):

PR #26, opened 2020-03-24
Increased RPC communication timeout to 20 secs

goruklu:
We have observed that in some cases DRM initialization takes longer than 10 seconds causing OpenCDM update() to fail with a timeout. Increasing the timeout to 20 seconds reduces the frequency of these errors.

### DIFF
--- a/Source/com/Administrator.h
+++ b/Source/com/Administrator.h
@@ -36,7 +36,7 @@ namespace RPC {
 #ifdef __DEBUG__
     enum { CommunicationTimeOut = Core::infinite }; // Time in ms. Forever
 #else
-    enum { CommunicationTimeOut = 10000 }; // Time in ms. 10 Seconden
+    enum { CommunicationTimeOut = 20000 }; // Time in ms. 20 Seconden
 #endif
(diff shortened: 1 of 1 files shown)


pwielders:
That is a bit scary. It means that the communication channel (since it is a synchronous invoke) is blocked for potentially 20S. It looks like the initialization of the OCDM should be decoupled and report back after some time that it initialized correctly. 
I assume this happens during the loading of the OCDM plugin ?

goruklu:
we are seeing this when creating a new DRM session with Playready 2.5. I'll see if we can increase the timeout value only in that particular case instead of increasing the generic timeout value.

pwielders:
If it is on creating a pLAYrEADY 2.5 (software drm I assume) session, the answer is probably even easier :-) Currently I know that the callback to report the generated License message is part of the Creation of the plugin (different from WideVine where this is really decoupled) I think this is a good time and reason to also in PlayReady decouple it. I guess this will drop the invoke time more than enough.

goruklu:
will try to find another way to handle the slow DRM_initialize() cases.

PR #28, opened 2020-03-26
[STATE TRIGGER] Added typecast for comparing strongly typed enums.

anjalirajan:
To compare the strongly typed enums (replaced the  old-style C++ enums ), added a  typecast.


Source/core/StateTrigger.h
```diff
@@ -110,7 +110,7 @@ namespace Core {
         m_AdminLock.Lock();
 
         // Now wait till our state is reached.
-        while (((a_State & m_State) == 0) && (blTimeOut == false)) {
+        while (((a_State & static_cast<uint32_t>(m_State)) == 0) && (blTimeOut == false)) {
```
MFransen69: Although I guess uint32_t and unsigned int are the same at the moment to prevent any problems and cast to the exact same type "unsigned int" (also below of course).

MFransen69 (changes requested):

MFransen69 (approved):

PR #32, opened 2020-03-26
add opencdm adapter for BCOM SVP with gstreamer secure buffer metadata 

goruklu:


pwielders (approved):
Nice, added functionality for people that want to use the secbuf of BRCM. However, I am missing, I think, a CMakefile in the adapter that adds this new source file ?

PR #33, opened 2020-03-26
Add callsign support for Thunder plugins

goruklu:

### DIFF
--- a/cmake/project.cmake.in
+++ b/cmake/project.cmake.in
@@ -94,6 +94,10 @@ macro(write_config)
                 endforeach()
             endif()
 
+            if (NOT ${callsign} STREQUAL "")
+                map_append(${plugin_config} callsign ${callsign})
+            endif()
+
             if (NOT ${autostart} STREQUAL "")
                 map_append(${plugin_config} autostart ${autostart})
             endif()


pwielders (changes requested):
I think this will break a lot of functionality. The Callsign (in case the config file is a separate entity) is equal to the name of this config file, without the json extension. You could, potentially, override this in the config by inserting this keyword. If you would now add this patch all plugins will get the same callsign, since typically we do not want to overwrite this callsign name (already depicted by the filename).
Why is this patch created, what was the usecase ?

anand-ky:
Pierre - The use case is to be able to create fully qualified names for plugin callsigns like "org.rdk.DisplaySettings". We felt it would be easier to set this in the config file instead of changing the file name. If this parameter is not set in the config, it would use the name of the file as the callsign.

pwielders:
> Pierre - The use case is to be able to create fully qualified names for plugin callsigns like "org.rdk.DisplaySettings". We felt it would be easier to set this in the config file instead of changing the file name. If this parameter is not set in the config, it would use the name of the file as the callsign.
Oke, that makes sense, and do agree to do it like that. Let me re-evaluate if we would than run into issues.

pwielders:
Oke, that makes sense, and do agree to do it like that. Let me re-evaluate if we would than run into issues.

pwielders (approved):

pwielders:
@sergiygladkyy can you please sign the CLA so I can merge this pull request ?

sergiygladkyy:
Signed!

PR #34, opened 2020-03-26
install package config for websocket

goruklu:


pwielders (changes requested):
Can we make the description a bit broader. Now it suggests that this is only for the JSONRPC headers but it also installs HTTP protocol (RFC2616), WebSocket (RFC6455) and JSONRPC functionality.

pwielders (approved):

PR #35, opened 2020-03-26
added playbackStopped method in opencdm interface

goruklu:
We need to know when playback has stopped but CDM session is still active. Some apps keep the CDM session alive even when it's no longer active. 

This is useful when we are enforcing HDCP 2.2 requirement for example.

### DIFF
--- a/Source/interfaces/IDRM.h
+++ b/Source/interfaces/IDRM.h
@@ -289,6 +289,8 @@ public:
         const uint32_t f_cbClearContentOpaque,
         uint8_t* f_pbClearContentOpaque)
         = 0;
+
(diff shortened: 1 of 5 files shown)


pwielders (changes requested):
I do not think that Player behaviour should be forwarded to a decryption module. If you want to stop playback, I suggest to stop playback at the player level and Release" the session that is associated with it. This means that we are using the regular lifetime interfaces on any object.
What would be the action on the PlaybackStopped with respect to the Decryption?

goruklu:
@pwielders based on our discussion, I've renamed the function playbackStopped() to resetOutputProtection() to make it more clear what it does.

pwielders (approved):

PR #45, opened 2020-04-05
Development/hdcp improvements

MFransen69:


pwielders (approved):
Wow.... even a windows implementation 👍

PR #46, opened 2020-04-06
64 bit fix for downcasting issues in Communicator

sergiygladkyy:
reinterpret_cast<unsigned int> from void* fails when data types have different lengths (e.g. on x86_64)

### DIFF
--- a/Source/com/Communicator.cpp
+++ b/Source/com/Communicator.cpp
@@ -430,7 +430,7 @@ namespace RPC {
         ASSERT(BaseClass::IsOpen() == false);
         _announceEvent.ResetEvent();
 
-        instance_id impl = reinterpret_cast<instance_id>(implementation);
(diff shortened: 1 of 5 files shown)


pwielders:
Preprocessor usaga has the disadvantage of not obeying the namespace solutions, hence why I think using the preprocessor to fix this issue is not preferred. 
Further more I think it is good to channel all conversion calls through methods that have a _cast postfic (const_cast, static_cast, reinterpret_cast, dynamic_cast and in thunder proxy_cast and now instance_cast), this allows for easy highlighting of potential intersting code :-)
I created a counter proposal, to be found here:
https://github.com/rdkcentral/Thunder/tree/development/PlatformPortability
I hope this is sufficient to address your issue, if not let me know the issue.

pwielders:
Currently added https://github.com/rdkcentral/Thunder/pull/50 that should resolve the same issue, which meand we should be able to close this pullrequest.

PR #49, opened 2020-04-07
Leak OpenCDMSession::Sink to avoid crashes on late SessionCallbackStubsMethods callbacks

varumugam123:
Hi,

We have been getting crashes with the below stack trace
```
[ 0 ] WPEFramework::ProxyStubs::<lambda(WPEFramework::Core::ProxyType<WPEFramework::Core::IPCChannel>&, WPEFramework::Core::ProxyType<WPEFramework::Core::IPCMessageType<2u, WPEFramework::RPC::Data::Input, WPEFramework::RPC::Data::Output> >&)>::_FUN
[ 1 ] WPEFramework::ProxyStub::UnknownStubType<OCDM::ISession::ICallback, ((void (**)(WPEFramework::Core::ProxyType<WPEFramework::Core::IPCChannel>&, WPEFramework::Core::ProxyType<WPEFramework::Core::IPCMessageType<2u, WPEFramework::RPC::Data::Input, WPEFramework::RPC::Data::Output> >&))(& WPEFramework::ProxyStubs::SessionCallbackStubMethods))>::Handle
[ 2 ] WPEFramework::RPC::Administrator::Invoke
[ 3 ] WPEFramework::RPC::InvokeServerType<4u, 1u>::Info::Dispatch
[ 4 ] WPEFramework::Core::ThreadPoolType<WPEFramework::RPC::InvokeServerType<4u, 1u>::Info, 1u, 4u>::ThreadUnitType<WPEFramework::RPC::InvokeServerType<4u, 1u>::Info>::Worker
[ 5 ] WPEFramework::Core::Thread::StartThread
[ 6 ] start_thread
[ 7 ] clone
```

On analysis it is found that when the callbacks are received by client post OpenCDMSession destruction, SessionCallbackStubsMethods tries to call one of its stubs with stale pointer to OpenCDMSession::Sink. The commit leaks OpenCDMSession::Sink and ensures that any late callbacks doesn't lead to crash.

### DIFF
--- a/Source/core/SystemInfo.cpp
+++ b/Source/core/SystemInfo.cpp
@@ -233,7 +233,10 @@ namespace Core {
             uint64_t DeltaTickCount = CurrentTickCount - previousTickCount;
             uint64_t DeltaIdleTime = CurrentIdleTime - previousIdleTime;
 
-            SystemInfo::m_cpuload = ((DeltaTickCount - DeltaIdleTime) * 100) / DeltaTickCount;
+            if (DeltaTickCount == 0)
+                SystemInfo::m_cpuload = 0;
+            else
+                SystemInfo::m_cpuload = ((DeltaTickCount - DeltaIdleTime) * 100) / DeltaTickCount;
 
             // Store current tick statistics for next cycle
             previousTickCount = CurrentTickCount;


pwielders (changes requested):
This sounds like resolving issues caused by incorrect usage.
If the session is destrcuted (a c call to destruct is done) the sink *MUST* be removed prior to that action.
So suggest to fix this at the root and not the collateral damage that is caused by usage of OCDM. Do you have a reproduction scenario ?

varumugam123:
Hi @pwielders, This crash happens when OCDM plugin takes more time to respond through callbacks. @goruklu shared a mechanism to induce a crash earlier today, by forcing MediaKeySession creation fail with "DRM_E_TEST_OPL_MISMATCH" after 30 seconds by which time the OpenCDMSession is destroyed. So when the callback reaches the stub after 30 seconds, it tries to call OnKeyStatusUpdate() with the stale pointer and eventually crashes

pwielders:
I indeed discusssed this with @goruklu and the idea is to properly solve this by decoupling the callback in the drm, so that the initializations takes no more than 3 seconds. 
I think if you destroy the session, the sink should also be decoupled and destructed so the callback should never happen.
From your remark, I think we need to investigate "why" the DRM_E_TEST_OPL_MISMATCH is still being received aftre 30 Seconds, since the full session should have been desturcted alread=ty and the server should not send a callback anymore... So I think the root cause might be in the OCDM Server...

pwielders:
Sorry wrong button

pwielders:
Sorry, twice the same mistake :-)

varumugam123:
> From your remark, I think we need to investigate "why" the DRM_E_TEST_OPL_MISMATCH is still being received aftre 30 Seconds, since the full session should have been desturcted alread=ty and the server should not send a callback anymore... So I think the root cause might be in the OCDM Server...

My theory:
1) Client makes a call say MakeKeySession() and returns expecting an async response (through callbacks)
2) Server receives the request and processing but got stuck somewhere (this goes for 30 seconds)
3) Client request timesout after 20 seconds. And client initiates Session destruction
4) Server returns from the hang (previous API call) and sends the response through callback
5) Server proceeds to handling session destruction (this is where actually the Revoke() happens)

Why OCDM hangs for 30 seconds is again a question. But the fact is, client can receive callbacks when the object is no more alive. WPEFramework has any mechanism to handle this scenario @pwielders ?

goruklu:
@pwielders Playready OCDM server taking too long during session creation is something we will be looking for sure. 
but it seems like there is general issue in which :

- client (wpe-webkit) calls OCDM createSession().
- in the OCDM server, session creation either takes a long time (> 10 seconds) and/or  fails to initialize DRM module.
- client receives a timeout from session create() call and destroys the session.
- on the OCDM server, generateKeyRequest() is called which then calls one of the callback functions; onKeyStatusUpdate() with an error or onKeyMessage() with the license request
- In the ProxyStubs_OpenCDMI, the callback causes a crash since the sink object is no longer valid.

There needs to be some protection for callbacks when the parent sink is no longer valid. Also, I don't see a good way to indicate to the client that something went wrong in session creation (no way to return an error code)

pwielders:
How about:

-  client (wpe-webkit) calls OCDM createSession(). 

- in the OCDM server, session creation either takes a long time (> 10 seconds) and/or fails to initialize DRM module.

- client receives a timeout from session create() call and destroys the session.

[PW] Lets pause here, the fact that the client, destroys the session means a release should be send to the erver of the "drop" of this session. since this is the last reference still available to this session, the release will be send over the COMRPC connection to the server, tehre the server must destruct the session and cleanup any still in progress being initialization of this session and thus revoke the Sink that was given to the session (no callbacks should happen anymore after tjhis destruction) than if the whole sesssion is destructed (remeber COMRPC is synchronopus and thus the release will only return from the server if the full stuff has been torn down, the client side sink can be safely destructed. the next steps in your sequence..
I think here might be a bug (the root cause) if this does not happen and I think that is what we should fix.
 

[PW] The client released the session and now releases (and thus deletes) the sink, this would cause your crash after the next line you suggested would be executed.

- on the OCDM server, generateKeyRequest() is called which then calls one of the callback functions; onKeyStatusUpdate() with an error or onKeyMessage() with the license request

[PW] This will not take place as the release in the previous step destructed the whole session, at least it should have, and maybe this is the real bug.

- In the ProxyStubs_OpenCDMI, the callback causes a crash since the sink object is no longer valid.

[PW] This will not take placethe previous step could already not take place since everyhting is destructed

varumugam123:
> [PW] Lets pause here, the fact that the client, destroys the session means a release should be send to the erver of the "drop" of this session. since this is the last reference still available to this session, the release will be send over the COMRPC connection to the server, tehre the server must destruct the session and cleanup any still in progress being initialization of this session and thus revoke the Sink that was given to the session (no callbacks should happen anymore after tjhis destruction) than if the whole sesssion is destructed

```
    virtual ~OpenCDMSession()
    {
        OpenCDMAccessor* system = OpenCDMAccessor::Instance();

        system->RemoveSession(_sessionId);

        if (IsValid()) {
           _session->Revoke(&_sink);
        }

        if (_session != nullptr) {
            Session(nullptr);
        }

        system->Release();
        TRACE_L1("Destructed the Session Client side: %p", this);
    }
```
**So here the problem is since the CreateSession() itself is timingout, OpenCDMSession does not hold a valid _session, as you can see, no Revoke happens during the destruction from client for the callback**

Also I saw sometimes, the system->Release() ends up in waiting further
```
#2  __pthread_cond_wait_common (abstime=0xbecc5ff8, mutex=0xbecc6034, cond=0xbecc6050) at /usr/src/debug/glibc/2.24-r0/git/nptl/pthread_cond_wait.c:533
#3  __pthread_cond_timedwait (cond=0xbecc6050, mutex=0xbecc6034, abstime=0xbecc5ff8) at /usr/src/debug/glibc/2.24-r0/git/nptl/pthread_cond_wait.c:667
#4  0xb58ab882 in __pthread_cond_timedwait (cond=cond@entry=0xbecc6050, mutex=mutex@entry=0xbecc6034, abstime=abstime@entry=0xbecc5ff8) at /usr/src/debug/glibc/2.24-r0/
git/nptl/forward.c:162
#5  0xb38d4e38 in WPEFramework::Core::Event::Lock (this=this@entry=0xbecc6030, nTime=nTime@entry=20000) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git
/Source/core/Sync.cpp:900
#6  0xb46eab90 in WPEFramework::Core::IPCChannelType<WPEFramework::Core::SocketPort, WPEFramework::Core::Void>::IPCTrigger::Wait (waitTime=20000, this=0xbecc6028) at /u
sr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/core/IPCConnector.h:909
#7  WPEFramework::Core::IPCChannelType<WPEFramework::Core::SocketPort, WPEFramework::Core::Void>::Execute (this=0x165ae0, command=..., waitTime=20000) at /usr/src/debug
/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/core/IPCConnector.h:1102
#8  0xb46ed010 in WPEFramework::Core::IPCChannel::Invoke<WPEFramework::Core::IPCMessageType<2u, WPEFramework::RPC::Data::Input, WPEFramework::RPC::Data::Output> > (wait
Time=20000, command=..., this=0x165ae0) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/core/IPCConnector.h:762
#9  WPEFramework::ProxyStub::UnknownProxy::Invoke (this=this@entry=0x18acbc, message=..., waitTime=waitTime@entry=20000) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+
0105e8f8ed-r0/git/Source/com/IUnknown.h:266
#10 0xb46ef59a in WPEFramework::ProxyStub::UnknownProxy::Release (this=this@entry=0x18acbc) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/com/
IUnknown.h:199
#11 0xb46ef5e4 in WPEFramework::ProxyStub::UnknownProxyType<OCDM::IAccessorOCDM>::Release (this=0x18acb8) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/g
it/Source/com/IUnknown.h:414
#12 0xb46ea18c in OpenCDMAccessor::~OpenCDMAccessor (this=this@entry=0x13f658, __in_chrg=__in_chrg@entry=2, __vtt_parm=__vtt_parm@entry=0x0) at /usr/src/debug/wpeframew
ork/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm_impl.h:118
#13 0xb46ea21c in OpenCDMAccessor::~OpenCDMAccessor (this=this@entry=0x13f658) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm_imp
l.h:127
#14 0xb46ea23a in OpenCDMAccessor::~OpenCDMAccessor (this=0x13f658) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm_impl.h:127
#15 0xb46e905a in OpenCDMAccessor::Release (this=<optimized out>) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm_impl.h:145
#16 0xb46eba1e in OpenCDMSession::~OpenCDMSession (this=0x2b40d8, __in_chrg=<optimized out>) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/oc$m/open_cdm_impl.h:551
#17 0xb46eba72 in OpenCDMSession::~OpenCDMSession (this=0x2b40d8, __in_chrg=<optimized out>) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/oc$m/open_cdm_impl.h:553
#18 0xb46e865c in OpenCDMSession::Release (this=0x2b40d8) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm_impl.h:561
#19 opencdm_destruct_session (session=0x2b40d8) at /usr/src/debug/wpeframework/1.99+gitAUTOINC+0105e8f8ed-r0/git/Source/ocdm/open_cdm.cpp:236
#20 0xb638d7f4 in WebCore::SessionDeleter::operator() () at /usr/src/debug/wpe-webkit/0.4.4+gitAUTOINC+5f899bc2e0-r0/git/Source/WebCore/platform/graphics/gstreamer/eme$CDMOpenCDM.h:45
#21 std::unique_ptr<OpenCDMSession, WebCore::SessionDeleter>::~unique_ptr () at /usr/include/c++/6.4.0/bits/unique_ptr.h:239
#22 WebCore::CDMInstanceOpenCDM::Session::~Session () at /usr/src/debug/wpe-webkit/0.4.4+gitAUTOINC+5f899bc2e0-r0/git/Source/WebCore/platform/graphics/gstreamer/eme/CD$
OpenCDM.cpp:361
#23 0xb638f5c0 in WTF::ThreadSafeRefCounted<WebCore::CDMInstanceOpenCDM::Session>::deref () at /usr/src/debug/wpe-webkit/0.4.4+gitAUTOINC+5f899bc2e0-r0/git/Source/WTF/$
tf/ThreadSafeRefCounted.h:71
#24 WebCore::CDMInstanceOpenCDM::requestLicense(WebCore::CDMSessionType, WTF::AtomicString const&, WTF::Ref<WebCore::SharedBuffer>&&, WTF::Ref<WebCore::SharedBuffer>&&$
 WTF::Function<void (WTF::Ref<WebCore::SharedBuffer>&&, WTF::String const&, bool, WebCore::CDMInstance::SuccessValue)>) () at /usr/src/debug/wpe-webkit/0.4.4+gitAUTOIN$
+5f899bc2e0-r0/git/Source/WTF/wtf/Ref.h:58
```

varumugam123:
I was trying to prevent callbacks reaching the ProxyStubs_OCDM.cpp by validating the impl existance in **_channelReferenceMap** something like this
```
bool Administrator::StubImplFind(const Core::ProxyType<Core::IPCChannel>& channel, Core::ProxyType<InvokeMessage>& message) {
// check _channelReferenceMap for the existance of message->Parameters().InterfaceId() & message->Parameters().Implementation()
}

void Administrator::Invoke(Core::ProxyType<Core::IPCChannel>& channel, Core::ProxyType<InvokeMessage>& message)
{
   uint32_t methodId(message->Parameters().MethodId());
   uint32_t interfaceId(message->Parameters().InterfaceId());
   instance_id implementation(message->Parameters().Implementation());

   std::map<uint32_t, ProxyStub::UnknownStub*>::iterator index(_stubs.find(interfaceId));
   if (index != _stubs.end()) {
       if(StubImplFind(channel, message))
           index->second->Handle(methodId, channel, message);
    } else {
        // Oops this is an unknown interface, Do not think this could happen.
        TRACE_L1("Unknown interface. %d", interfaceId);
    }
}
```
But again to validate this CreateSession() should have completed (so that RegisterUnknownInterface() would have been made through Administrator::AddRef()). But since **callbacks are sent as part of CreateSession (not post CreateSession)** I can't go with this. Somehow we need to validate incoming message against existence of an interface implementation (instance).

pwielders:
I think indeed you found the rootcause. The fact that the session is not valid prevents the sink from being Revoked. Good point. That brings me back to the initial statement we need to decouple the message sending of the first license message (currently done on the Create Session Invoke for the PlayReady Implementation), which causes the "long" RPC invoke initially and I assume the time out of the create that in turn than reports a "invalid" session.
I think the intention of the OCDM Frauenhofer design of this interface would be to make the "initialization" short and limited e.g. < 1S, because the creation of the licensing request is non-deterministic, it needs to occure n the callback (currently the PlayReady implementation (and maybe the factory around it) do not implement it this way, right now...
I think this is reatedto: https://github.com/rdkcentral/Thunder/pull/26 Where the request was to "extend" the COMRPC timeout. I think decoupling the callback from the CreateSession, solves both :-)

varumugam123:
@pwielders As you mentioned, I was trying to not to send callbacks from OCDM plugin's CreateSession() and enforce validation of Invoke messages from Administrator. The above validation on Administration imposes that any calls to Stubs will be allowed only after the implementations are registered with Administrator.

Please find the OCDM Plugin changes here : [send_callbacks_after_createsession_completion.txt](https://github.com/rdkcentral/Thunder/files/4503198/send_callbacks_after_createsession_completion.txt)

PR #50, opened 2020-04-07
[PORTABILITY] Seems that mixing 32/64 bit platforms the void* can not…

pwielders:
… always be casted.


sebaszm (approved):
Looks alright to me. However perhaps just using uintptr_t instead of void* would be a simpler solution?

PR #63, opened 2020-04-14
[ProcessContainers] Integration of crun backend

mipo57:
Added a crun backend for process containerization in Thunder.
Also added missing license headers from other backends


pwielders (changes requested):
I think lifetime management is already absttracted by an interface called IReferenceCounted (See https://github.com/rdkcentral/Thunder/blob/master/Source/core/Portability.h#L587-L593)
One of the joys of this interface is that it fits the ProxyType structures and it forces the smae method for lifetime everywhere (COMRPC objects inherit from this interface as well) So please align the AddRef and Release methods with this interface, than You will already see that there are two different types of Releases used in your code (return value differntaiation)

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.h
```diff
@@ -0,0 +1,86 @@
...
+
+        std::string Name() const override;
+        uint32_t NumIPs() const override;
+        std::string IP(uint32_t id) const override;
+    };
+
+    class CRunContainer : public IContainer 
+    {
+    public:
+        CRunContainer(string name, string path, string logPath);
```
pwielders: If it is reference counted, usggest making the Constructor private, so you can only get instances of this interface/Object through a Create method, this way the object could not be potentially placed on a stack

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.h
```diff
@@ -0,0 +1,86 @@
...
+    class CRunContainerAdministrator : public IContainerAdministrator 
+    {
+        friend class CRunContainer;
+    public:
+        IContainer* Container(const string& id, 
+                                IStringIterator& searchpaths, 
+                                const string& logpath,
+                                const string& configuration) override; //searchpaths will be searched in order in which they are iterated
+
+        CRunContainerAdministrator();
```
pwielders, line 78: Sama as above, if it is refcounted, make sure the object can not be created on the stack, add a static Create method (or I think here it is also assumed a singleton ??? If so create a static Instance method. Preferably returning the interface, i.s.o. the object.

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+    {
+        libcrun_error_t error = nullptr;
+        int ret = 0;
+
+        // Find containers 
+        libcrun_container_list_t *list;
+        ret = libcrun_get_containers_list (&list, "/run/crun", &error);
+        if (ret < 0) {
+            TRACE_L1("Failed to get containers list. Error %d: %s", error->status, error->msg);
+            return Core::ERROR_UNAVAILABLE;
```
pwielders: Minor, but still for the next time, for maintainability reasons, the coding guidelines only allow for 1 return statement in a method.

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    void CRunContainer::AddRef() const 
+    {
+        _refCount++;
+    };
+
+    uint32_t CRunContainer::Release()
+    {
+        if (--_refCount == 0) {
```
pwielders: One of the reasons for having a basic implementation of IReferenceCounted is to avoid race conditions. _refCount id defined as a uint32_t which is not an atomic. However reference counting might get messed up if this is not a locked increment/decrement or atomic.

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+        result = libcrun_read_container_status (&status, _context.state_root, _name.c_str(), &error);
+        if (result != 0) {
+            TRACE_L1("Failed to get PID of container %s", _name.c_str());
+        }
+
+        return status.pid;
+    }
+
+    CRunContainer::MemoryInfo CRunContainer::Memory() const
+    {
```
pwielders: Memory measurements/aabstractions have already been part of the Core, why not use the core? This way enhancement (like sander did on the ProcessInfo object, will roll in for free)

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+    uint32_t CRunContainerAdministrator::Release()
+    {
+        --_refCount;
+
+        return (Core::ERROR_NONE);
+    }
+    
+    // Container
+    // ------------------------------------
+    CRunContainer::CRunContainer(string name, string path, string logPath)
```
pwielders: Would const string& not be better suited here ?

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    CRunContainerAdministrator::ContainerIterator CRunContainerAdministrator::Containers()
+    {
+        return ContainerIterator(_containers);
+    }
+
+    void CRunContainerAdministrator::AddRef() const
+    {
+        _refCount++;
```
pwielders: atomic or InterLockedIncrement/Decrement (prefer the latter)
Interesting, I do not think the reference count has any meaning as this is just singleton :-)

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+    CRunContainerAdministrator::CRunContainerAdministrator()
+        : _refCount(1)
+    {
+
+    }
+
+    CRunContainerAdministrator::~CRunContainerAdministrator()
+    {
+        if (_containers.size() > 0) {
```
pwielders: Since it is a singleton, lets make sure that there is only 1 reference left if we are destructed. The reference we hold on our selves. So assert (_refCount == 1) might be interesting here. Than you will see that adding the reference count (addRef on the instance, will balance the lifetime managament correctly.

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+        return result;
+    }
+
+    bool CRunContainer::Stop(const uint32_t timeout /*ms*/)
+    {
+        bool result = true;
+        libcrun_error_t error = nullptr;
+
+        if (libcrun_container_delete (&_context, NULL, _name.c_str(), true, &error) != 0) {
```
pwielders: I am missing the point here where the container would be deleted from the CRunAdministrator ?

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+    void CRunContainer::AddRef() const 
+    {
+        _refCount++;
+    };
+
+    uint32_t CRunContainer::Release()
+    {
+        if (--_refCount == 0) {
+            if (_created == true) {
```
pwielders: I think this container will never reach a reference count of 0 as the CRunAdministrator will always hold the last refernce..... It will only be cleared if the administrator is destructed. Is that as designed ? Should the container only stop if the ContainerAdministrator is destructed?

pwielders (changes requested):

Source/processcontainers/implementations/CRunImplementation/CRunImplementation.h
pwielders: also maybe use const string& as parameters ?

MFransen69 (changes requested):
We'll discuss of line

Source/processcontainers/ProcessContainer.h
```diff
@@ -27,71 +27,71 @@ namespace ProcessContainers {
...
-
-        // Lifetime management:
-        // ------------------------------------
-        void AddRef();
-        void Release();
-        
-        // Implementation-dependent
-        // -----------------------------------
-
+    class INetworkInterfaceIterator : public Core::IIterator
```
MFransen69: Very good to derive from the IIterator but I think we now lost the refcount interface (although still implemented in NetworkInfoUnimplemented.cpp). I would propose to add IReferenceCounted as well

Source/processcontainers/ProcessContainer.h
```diff
@@ -130,7 +130,7 @@ namespace ProcessContainers {
         virtual void Logging(const string& globalLogPath, const string& loggingoptions) = 0;
 
         // Returns ids of all created containers
-        virtual IContainerIterator* Containers() = 0;
+        virtual Core::ProxyType<IContainerIterator> Containers() = 0;
```
MFransen69: See comment in other Commit, please return the raw interface

Source/processcontainers/ProcessContainer.h
```diff
@@ -102,12 +102,10 @@ namespace ProcessContainers {
         // Stops the running containerized process. Returns true when stopped.
         // Note: if timeout == 0, call is asynchronous
         virtual bool Stop(const uint32_t timeout /*ms*/) = 0; 
-
-        virtual void AddRef() const = 0;
-        virtual uint32_t Release() = 0;
     };
 
-    struct IContainerAdministrator {
+    struct IContainerAdministrator
```
MFransen69: I'm not a big fan on it also but we follow the WebKit style guide where it is class/struct name { 
So could you please use that style all over the file(s) (I think Bram created a code beautifier script for this if that is more convenient)

Source/processcontainers/ProcessContainer.h
```diff
@@ -102,12 +102,10 @@ namespace ProcessContainers {
...
-        virtual void AddRef() const = 0;
-        virtual uint32_t Release() = 0;
     };
 
-    struct IContainerAdministrator {
+    struct IContainerAdministrator 
+    {
         static IContainerAdministrator& Instance();
 
         IContainerAdministrator() = default;
```
MFransen69: Perhaps better to remove this altogether for constancy if not private. We normally also not in this file declare the c'tor for an interface

Source/processcontainers/common/BaseAdministrator.h
```diff
@@ -20,59 +20,58 @@
...
 
 #include "processcontainers/ProcessContainer.h"
+#include "BaseContainerIterator.h"
+#include "Lockable.h"
 
 namespace WPEFramework {
 namespace ProcessContainers {
 
-    template <typename TContainer, typename Mixin> // IContainerAdministrator Mixin
+    template <typename TContainer, typename Mixin> // IContainerAdministrator, Lockable Mixin
```
MFransen69, line 28: Let's for now leave it like it is. I had a look and I understand what you are trying to achieve and we can have a chat about when the times comes again we can be in front of a whiteboard again. It is implementation behind an interface and we can always change it if we want. One remark perhaps, be a bit reluctant using mixins, they do add to the complexity of the software and not everybody might understand what you are trying to achieve

MFransen69 (changes requested):

Source/processcontainers/ProcessContainer.h
```diff
@@ -27,71 +27,71 @@ namespace ProcessContainers {
...
         // Return numver of ip addresses assigned to the interface
         virtual uint32_t NumIPs() const = 0;
 
         // Get n-th ip assigned to the container. Empty string
         // is returned if n is larger than NumIPs()
-        virtual std::string IP(uint32_t n) const = 0;
-    protected:
-        uint32_t _current;
-        uint32_t _count;
+        virtual string IP(uint32_t n = 0) const = 0;
```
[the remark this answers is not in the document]
mipo57: Well - its for convenience. In most cases, you will only have either 0 or 1 IP assigned to an interface and probably on 99.9% cases you would only care about one of them (the most likely first one)

Source/processcontainers/common/BaseContainerIterator.h
```diff
@@ -0,0 +1,108 @@
...
+#pragma once
+
+#include "processcontainers/ProcessContainer.h"
+
+namespace WPEFramework {
+namespace ProcessContainers {
+
+    class BaseContainerIterator : public IContainerIterator {
+    public:
+        BaseContainerIterator(const std::vector<string>&& idList)
```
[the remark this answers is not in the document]
mipo57: I don't know why, but for some reason, GitHub displays outdated code to review

Source/processcontainers/common/BaseContainerIterator.h
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        virtual uint32_t Count() const
+        {
+            return _ids.size();
+        }
+        
+        const string& Id()
+        {
+            return _ids[_current];
```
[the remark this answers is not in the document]
mipo57: Isn't that whats IsValid function is for?
Eg. the normal pattern would be 
while (iterator.Next() == True) { // This checks if isValid()
[...]
temp = iterator.Id()
}
So checking IsValid inside would be redundant
But I guess assert could be useful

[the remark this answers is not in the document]
mipo57: Isn't that whats IsValid function is for?
Eg. the normal pattern would be
```c++
while (iterator.Next() == True) { // This checks if isValid()
    [...]
    temp = iterator.Id()
}
```
So checking IsValid inside would be redundant.
But I guess assert could be useful

Source/processcontainers/ProcessContainer.h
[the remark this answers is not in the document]
mipo57: What would be the reason of making the function parameter (non-pointer/reference) a const?

Source/processcontainers/ProcessContainer.h
```diff
@@ -23,75 +23,62 @@
...
 namespace WPEFramework {
 namespace ProcessContainers {
-    
     using IStringIterator = Core::IteratorType<std::vector<string>, const string>;
-    using IConstStringIterator = Core::IteratorType<const std::vector<string>, const string, std::vector<string>::const_iterator>;
-
-    class INetworkInterfaceIterator : public Core::IIterator 
-    {
-    public:
-        virtual ~INetworkInterfaceIterator() = default;
```
MFransen69: To keep it consistent with our codebase (I know it is not technically needed as d'tor already in base class) please keep the destructors (so also everywhere below) (and according to new rule here you would use override instead of virtual, but that is not yet everywhere in our code base)

Source/processcontainers/ProcessContainer.h
```diff
@@ -130,7 +114,7 @@ namespace ProcessContainers {
         virtual void Logging(const string& globalLogPath, const string& loggingoptions) = 0;
 
         // Returns ids of all created containers
-        virtual Core::ProxyType<IContainerIterator> Containers() = 0;
+        virtual IContainerIterator* Containers() = 0;
```
MFransen69: Sorry for not noticing this earlier but shouldn't this method be const?

MFransen69 (changes requested):

MFransen69 (approved):

pwielders (approved):

PR #65, opened 2020-04-14
Development/cg open ssl import keys

sebaszm:
OpenSSL cryptography implementation to load pre-shared keys off of disk on startup. Smoke-tested.


Source/cryptography/implementation/OpenSSL/Vault.cpp
```diff
@@ -24,64 +24,156 @@
...
 {
-    typedef uint8_t pkey[16];
+    /* Netflix vault specific initialisation */
+    if (id == CRYPTOGRAPHY_VAULT_NETFLIX) {
+        std::string path;
+        WPEFramework::Core::SystemInfo::GetEnvironment(_T("NETFLIX_VAULT"), path);
+        WPEFramework::Core::File file(path.c_str(), true);
+
+        if (file.Open(true) == true) {
+            struct NetflixData {
```
pwielders, line 59: Pragma pack(1) might be usefull on this as we do not know who or where it is compiled

sebaszm: It's there!

https://github.com/rdkcentral/Thunder/blob/e10166410a1a5d8ae09b337671d3e19684f261da/Source/cryptography/implementation/OpenSSL/Vault.cpp#L59-L65

pwielders (approved):

PR #66, opened 2020-04-15
Support for Floating point numbers in JSON

anand-ky:
1) Supports float (Core::JSON::Float) and double (Core::JSON::Double).
2) Floating-point values are converted to decimal or decimal exponent notation (%g conversion specifier) during serialization.
3) Infinity and NaN values are converted to "null" during serialization.


pwielders (approved):
Looks good. Let’s add this new functionality.

PR #73, opened 2020-04-29
[CRYPTO] Adds NetflixVaultGenerator

svdmaar:
Adds Netflix Vault Generator.


Tests/cryptography/NetflixVaultGenerator/main.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <openssl/bio.h>
+#include <openssl/evp.h>
+#include <openssl/aes.h>
+#include <iostream>
+#include <fstream>
+#include <openssl/modes.h>
+
+using namespace std;
+
+const uint32_t g_esnSize = 42;
```
sebaszm: Why is the ESN size fixed? I don't think it's always 42...? Did a quick search on the net and found different sizes of sample ESNs. The TA even has dedicated API to retrieve the ESN size, hence my doubts.

Tests/cryptography/NetflixVaultGenerator/main.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    NetflixData netflixData;
+
+    // 1. Create random IV vector and salt.
+    srand(time(nullptr));
+    uint8_t iv[16];
+    for (int i = 0; i < 16; i++) {
+        iv[i] = rand() % 256;
+        netflixData.m_salt[i] = rand() % 256;
+    }
```
sebaszm: OpenSSL has cryptography-strong random number generator functions RAND_***, why not use them instead?

Tests/cryptography/NetflixVaultGenerator/main.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    // 5. Write IV followed by encrypted NetflixData.
+    outFile.write(reinterpret_cast<const char *>(iv), sizeof(iv));
+    outFile.write(reinterpret_cast<const char *>(&encryptBuffer), sizeof(netflixData));
+
+    outFile.close();
+    inFile.close();
+
+    cout << "SUCCESS: Written netflix fault to " << argv[2] << endl;
```
sebaszm: Yeah, it's their fault! :)

sebaszm (changes requested):

svdmaar:
@sebaszm , thanks for you review feedback, I agree with all the points raised and amended a commit. Please have another look at this PR. Thanks

sebaszm (approved):
LGTM!

PR #74, opened 2020-04-30
[CORE] Store all locking callstacks for critical section

svdmaar:
Stores all callstacks used when locking CriticalSection.

If "CRITICAL_SECTION_LOCK_LOG" is enabled, all callstacks when locking a CriticalSection are logged, not just the most recent one. This allows one to investigate unmatched locks/unlocks.


pwielders:
I assume you will create a small Demo/PowerPoint presentation to explain how we can use this :-)

pwielders (approved):

PR #82, opened 2020-05-14
[Core][Netlink] Support for multimessage Netlink packets

mipo57:
This commit fixes problems with Netlink packages missing
and random interface reaction to the action on another interface. Also,
a watchdog on Netlink changes was added, so internal representation 
should be always in sync with interface MAC & name addresses

Summary of changes:
- Added full support for parsing of multi-message packages
- Interface information (only MAC for now) are now updated on change
- Removed few potential out-of-bound conditions


pwielders (approved):
The interface usage of the blocking NetLink calls need to be replaced by the NetlinkSocket (a-synhronous). But we will pick this up using a next Jira ticket.

PR #84, opened 2020-05-15
CompositorClient:Wayland:Zorder fraction calculation conversion issue updated

HaseenaSainul:
Currently the Zorder fraction calculation is not working, its always returning 1.0 since not casting the unit32_t order to double.

### DIFF
--- a/Source/compositorclient/Wayland/Westeros.cpp
+++ b/Source/compositorclient/Wayland/Westeros.cpp
@@ -564,6 +564,8 @@ namespace Wayland {
 
     void Display::SurfaceImplementation::ZOrder(const uint32_t order)
     {
+        // Max layers supported by Westeros have a limitation with 255, hence the ZOrder fraction
+        // difference calculation is limiting with std::numeric_limits<uint8_t>::max()
         ASSERT (order <= std::numeric_limits<uint8_t>::max());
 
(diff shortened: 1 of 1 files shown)


Source/compositorclient/Wayland/Westeros.cpp
```diff
@@ -566,7 +566,7 @@ namespace Wayland {
     {
         ASSERT (order <= std::numeric_limits<uint8_t>::max());
```
pwielders, line 569: We upgraded the zOrder we pass from the Compositor, to u uint16_t (used to be uint8_t indeed). So could you upgrade it to a uint16_t allows for 65535 layers, i.s.o. 256 :-)

pwielders (changes requested):
You could not know, but we upgraded the ZOrder from uint8_t -> uint16_t.

Source/compositorclient/Wayland/Westeros.cpp
HaseenaSainul: Sure will do

HaseenaSainul: The smaller difference the fraction is not working with ZOrder settings, it still behaves like 1.0 value. So it seems it supports only upto 255 layers

pwielders (approved):

PR #85, opened 2020-05-18
Development/security

pwielders:
**_Security change requirements:_**
The SecurityAgent needs to add the possibility to also evaluate the methods associated with the request (finer security granularity)
The plugin needs to be able to evaluate the token, used by the JSONRPC request to allow it to continue or to disallow it.

**_Changes:_**
The JSONRPC handling in the core now allows to be extended with a Lambda that can additionally test the method to Invoke on a plugin bases and passing it the active token for this method. This required in core changes.

There is also a deadlock fix, accidently triggered and caught (9f84032) @MFransen69  to code review this one.


Source/WPEFramework/PluginServer.cpp
```diff
@@ -125,10 +125,14 @@ namespace PluginHost
             return (result);
         }
         //! Allow a JSONRPC message to be checked before it is offered for processing.
-        virtual bool Allowed(const Core::JSONRPC::Message& message) const override
+        bool Allowed(const Core::JSONRPC::Message& message) const override
         {
             return ((_hasSecurity == false) || CheckMessage(message));
         }
+        string Token(void) const override
```
MFransen69, line 132: I would have expected string Token(), so without void

Source/WPEFramework/PluginServer.h
```diff
@@ -1394,1571 +1416,1592 @@ namespace PluginHost {
...
+                END_INTERFACE_MAP
 
+            private:
+                mutable uint32_t _refCount;
+                ServiceMap& _parent;
+                const CommunicatorServer& _comms;
+                const string _connector;
+            };
+            class Override : public Core::JSON::Container {
+            private:
```
MFransen69, line 1467: Should be in the public section :)

Source/WPEFramework/PluginServer.h
```diff
@@ -1394,1571 +1416,1592 @@ namespace PluginHost {
...
+                        , Configuration(copy.Configuration)
+                    {
+                        Add(_T("autostart"), &AutoStart);
+                        Add(_T("configuration"), &Configuration);
+                    }
 
-                        virtual ~Plugin()
-                        {
-                        }
+                    virtual ~Plugin()
```
MFransen69, line 1502: override?

MFransen69 (approved):

Source/WPEFramework/PluginServer.h
pwielders: This is code writtten in the time that the MIPS compiler was not supporting std11 (and thus the delete keyword).
We wgreed that we would not actively start updating all the code, but only if we tocuh it. I do not think this code was tocuched so hence it is till prior to the std11 standard (private copy and assignment ocnstructors if you do not support them)

Source/WPEFramework/PluginServer.h
pwielders: This is code writtten in the time that the MIPS compiler was not supporting std11 (and thus the delete keyword).
We wgreed that we would not actively start updating all the code, but only if we tocuh it. I do not think this code was tocuched so hence it is till prior to the std11 standard (override keyword was not supported on these compilers.... yet)

Source/WPEFramework/PluginServer.h
pwielders: I guess, it looks like I touched it because of the automatic styling tool I have installed. So I did not do it, my styling tool did it :-)

Source/WPEFramework/PluginServer.cpp
pwielders: Me too, seen that I am quite lazy, I can not imagen that I typed it so propbably I copy pasted it. However is does not hurt any functionality. Maybe it is this automatic style tool that did this. Interesting...

PR #92, opened 2020-05-26
[CyclicBuffer] Fixed the deadlock situation during the timed lock()- unlock() sequence, removed unwanted methods

anjalirajan:
Fixed the deadlock situation during the timed lock()- unlock() sequence, removed unwanted methods.


svdmaar:
Moved this work to the unit-test pull request, so this one can be closed without merging

PR #96, opened 2020-05-29
Add displayinfo comrpc convenience  library

bramoosterhuis:


Source/displayinfo/include/displayinfo.h
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * \brief Checks if a display is connected to the display output.
+ * 
+ * \param instance Instance of \ref displayinfo_type.
+ * 
+ * \return true a dispplay is connected, false otherwise.
+ * 
+ **/
+bool displayinfo_connected(struct displayinfo_type* instance);
```
svdmaar, line 126: If we want to make this header file also valid C, I don't think bool is allowed

pwielders: I was not sure myself as well so did a bit of googling:
Standard C (since C99) provides a boolean type, called _Bool . By including the header stdbool. h , one can use the more intuitive name bool and the constants true and false . ... Objective-C also has a separate Boolean data type BOOL , with possible values being YES or NO , equivalents of true and false respectively.
So I guess it os oke if we include <stdbool.h>

pwielders: And we do so I guess... GTG...

Source/displayinfo/DisplayInfo.cpp
```diff
@@ -0,0 +1,534 @@
...
+        {
+            ASSERT(_displayConnection != nullptr);
+            return _displayConnection->Height();
+        }
+
+        Exchange::IConnectionProperties::HDRType HDR() const
+        {
+            ASSERT(_displayConnection != nullptr);
+            return _displayConnection->Type();
+            ;
```
pwielders: ?????? for fun ????? an addional semicolon?

Source/displayinfo/DisplayInfo.cpp
```diff
@@ -0,0 +1,534 @@
...
+        }
+    }
+
+    DisplayInfoAdministrator(const DisplayInfoAdministrator&) = delete;
+    DisplayInfoAdministrator& operator=(const DisplayInfoAdministrator&) = delete;
+
+    void Register(DisplayInfo* client)
+    {
+        std::list<DisplayInfo*>::iterator index(std::find(_clients.begin(), _clients.end(), client));
+
```
pwielders, line 195: Be defensive... ASSERT if you find a client..

pwielders (changes requested):
DisplayInfo.cpp, the list needs to be protected using a CriticalSection. Currently, with the calback coming in from the multithreaded Thunder COMRPC, callback might be in progress while we are unregstering/Registerting. Also if multiple clients start to use these DislayIfo, requets might be coming in from multiple threads. I thik we need to make this library thread safe.

Source/displayinfo/DisplayInfo.cpp
```diff
@@ -0,0 +1,534 @@
...
+            }
+
+            client->Initialize(_comChannel);
+
+            if (client->IsValid() == true) {
+                Trace("Added client %p to registery[%d]", client, _clients.size());
+                _clients.push_back(client);
+            }
+
+            if (_clients.size() == 0) {
```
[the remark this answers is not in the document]
bramoosterhuis: If we add it to the previous if there is a issue when a user requests a non-existing implementation after requesting a valid one, it will also close the connection of the first one.

pwielders (approved):

PR #97, opened 2020-05-29
JSON Serialiser: Handle container end scenario using end-marker

HaseenaSainul:
JSON array container parsing logic is end-up with a infinite loop when the serialisation completed and there is no enough space to keep end marker in the stream (i.e, loaded reached to the max length). In such case this has to be marked and handle in the consequent iteration.

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -2807,6 +2807,8 @@ namespace Core {
                     _iterator.Reset();
                     stream[loaded++] = '[';
                     offset = (_iterator.Next() == false ? ~0 : PARSE);
+                } else if (offset == END_MARKER) {
+                    offset = ~0;
                 }
                 while ((loaded < maxLength) && (offset != static_cast<uint16_t>(~0))) {
                     if (offset >= PARSE) {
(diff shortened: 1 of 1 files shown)


Source/core/JSON.h
```diff
@@ -2809,9 +2811,13 @@ namespace Core {
...
                 }
-                if ((offset == static_cast<uint16_t>(~0)) && (loaded < maxLength)) {
-                    stream[loaded++] = ']';
-                    offset = 0;
+                if (offset == static_cast<uint16_t>(~0)) {
+                    if (loaded < maxLength) {
+                        stream[loaded++] = ']';
+                        offset = 0;
+                    } else {
+                        offset = END_MARKER;
```
pwielders, line 2829: This seems a bit scary. END_MARKER is just an enum value of 6. This means that you set the offset here to position 6. If for some reason we would continue to write characters at offset 6, we assume it is the end. Looks like the issue is more on the callee. DO you still know from where this was called ?

MFransen69, line 2829: It seems the unit tests also indicate this causes some regression. As Pierre mentions the offset of 6 now has a special meaning.

MFransen69 (changes requested):

Source/core/JSON.h
HaseenaSainul: Sorry, I missed your comments. I will check this

HaseenaSainul: Ok, @MFransen69

pwielders (approved):
Looks good. @MFransen69 did a deep dive into this as well. if you analyse the code in more detail, it seems that the PARSE (Max value) is always substracted before the actual Serialize/Deserialize is executed and after the Serialize/Deserialize it is added again. So the offset is more or less also used to maintain the state of the current status.
I do think we need to raie a JIRA to remove the QUIRK mode from the JSON tests. That is still an open issue but not related to this fix.

MFransen69 (approved):

PR #101, opened 2020-06-02
[Networking] Combined all netlink-related processing into SocketNetlink

mipo57:
- Moved every Netlink-related processing into a NetlinkSocket class.
- Added possibility to call exchange asynchronously
- Multi message packets are now parsed correctly
- Updated frame parsing to use NLMSG macros to make sure padding is correct


mipo57:
Still more work on making networking asynchronous needs to be done. Possibility to use Netlink asynchronously was added to SocketNetlink, however, NetworkControl plugin is still written in a synchronous way. At the moment, if something goes wrong  (especially during event-based flushing) we risk blocking ResourceMonitor thread.

Source/core/Netlink.h
```diff
@@ -396,7 +437,7 @@ namespace Core {
 
     public:
         SocketNetlink(const Core::NodeId& destination)
-            : SocketDatagram(false, destination, Core::NodeId(), 512, 1024)
+            : SocketDatagram(false, destination, Core::NodeId(), 4096, 8192)
```
pwielders, line 443: Did you get the sizes from a spec? Sending 4K and Receiving 8K? sounds big... Maybe good to put the reference here where you found this..

Source/core/NetworkInfo.cpp
```diff
@@ -1258,35 +1232,109 @@ namespace Core {
...
+
+        return (_channel->Exchange(modifier, modifier, waitTime));
     }
 
     uint32_t IPNetworks::Network::Gateway(const IPNode& network, const NodeId& gateway)
     {
         IPRouteModifyType<true> modifier(*this, network, gateway);
 
-        return (_channel->Exchange(modifier, modifier));
+#ifdef __DEBUG__
```
pwielders: Move this waitTime definition used (and now copied multiple times :-) ) to a single class static constexpr...

pwielders (changes requested):
Looks good just some minor remarks (create 1 setting for waitTime, that is different in debug time), and a general question. I assume you did not change any interfaces that are publicly used (I saw one change from Size() -> PayloadSize()) but I am hoping that is all.
The reason why I am asking:
1) Is windows still working :-)
2) The Netlink socket is used in projects you are not building and/or testing, should I rebuild/retest and fix those?

Source/core/Netlink.h
mipo57: I don't think they are limited by spec, however, in practice, it is limited to page size. Moreover, recommendation is to limit it to 8k if page size > 8k (https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/netlink.h?h=linux-4.9.y#n112). I limited send buffer to 4k as we control it and can always serve a more restricted 4k. I will add it in the comment :)

mipo57: In practice, I found multiple cases where the message was >4k and split into 2 ~3.5k  (as RPI page size is 4k). I think limiting it to 4k would be a little bit on the dangerous side here

Source/core/NetworkInfo.cpp
mipo57: 2 Changes were made. The Size() -> PayloadSize() you mentioned and Data() -> template<> Payload(), however, both changes were made in Frames class that is used to parse Netlink messages and wasn't used anywhere else (event in code you provided in conversation). 

These changes are obviously not absolutely required, however it was confusing at times whether Data() and Size() were referring to just payload or the whole frame.

mipo57: Actually after reviewing one more thing changed in Network (Event class), however, it came mostly from cutting changes into this smaller PR and I will undo it

pwielders (approved):

PR #114, opened 2020-06-12
Development/generate file out of source

bramoosterhuis:
During native development I saw that the generators where polluting the original sources even when I executed an out-of-tree build. With this change all the sources that are generated and those who where copied to a a different location configure time are now put in to the cmake binary directory. This leaves the original file structure untouched.


sebaszm (approved):
Lgtm

PR #121, opened 2020-06-22
[Core] Make the SynchronousChannel mark the outbound message as sent …

sebaszm:
…only once it's finished being processed completely.

The outbound message is marked as complete too early. The Write() at SocketPort still does one last iteration in its processing loop. If an inbound message is not expected and the sender reloads its outbound message quickly enough (as it is now free to do so), then the loop by accident picks up the new frame data to send to the socket instead of exiting.


pwielders (approved):
Good point. Challenging bug to find :-)

pwielders (approved):

PR #126, opened 2020-06-26
Development/json parser fix arraytypewithfloat

anjalirajan:
Fixed the serialization issue of ArraType with Float & Double values.

Please refer the test cases https://github.com/anjalirajan/Thunder/commit/522577c3b5995f4287d66599338c46c2abd5413b


pwielders (commented):
What was the issue? I think the static on line 1043 is very dangerous, multiple threads might be accessing this entry simutanously.

anjalirajan:
It fixes the serialization functionality of ArrayType with Float/Double values.  Means, without this fix, the else part of the condition is executed and it stores the content of _value to stream variable of size maxLength, where maxLength is 1, instead of the entire value.
https://github.com/anjalirajan/Thunder/blob/development/json-parser-fix-arraytypewithfloat/Source/core/JSON.h#L1105

The corresponding test cases are https://github.com/anjalirajan/Thunder/blob/anjali/unittest/Tests/core/test_json.cpp#L715 & https://github.com/anjalirajan/Thunder/blob/anjali/unittest/Tests/core/test_json.cpp#L734.

pwielders:
@gladish We do not use the whole float stuff, it was added by someone within comcast, maybe you can review this on validity?

PR #138, opened 2020-07-07
Anjali/unittest

svdmaar:
This pull request contains the work of @anjalirajan regarding the CyclicBuffer + a few fixes.

Long story short: there was a cond var used with unlocked mutex, a too strict comparison, and a few spots where we should update other clients waiting for an update. Besides that, it adds a bunch of unit tests (again credits to @anjalirajan for most of the work).


gladish:
Is this going to get merged? More unit-tests is good.

anjalirajan:
> Is this going to get merged? More unit-tests is good.

Yes. Test case implementation is ongoing.

pwielders (changes requested):
Agree we need to merge the Unit tets but we should not merge the CyclicBuffer changes. They seem to be too restrictive now. We need to make sure that this pull request only contains the unit tests and than add those. Than run the unit test that shows the issue in the cyclicbuffer and I would like to anaylse that test in more detail.
The Cyclic buffer is in the core of the tracing any error there will deadlock the system and make it usesless. I woud like to see these as two separate tasks:
1) Pull request 1: add UnitTests
2) Pull request 2: Make fixes for unit test cases that prove to fail (after analysing that test case)Agree we need to merge the Unit tets but we should not merge the CyclicBuffer changes. They seem to be too restrictive now. We need to make sure that this pull request only contains the unit tests and than add those. Than run the unit test that shows the issue in the cyclicbuffer and I would like to anaylse that test in more detail.
The Cyclic buffer is in the core of the tracing any error there will deadlock the system and make it usesless. I woud like to see these as two separate tasks:
1) Pull request 1: add UnitTests
2) Pull request 2: Make fixes for unit test cases that prove to fail (after analysing that test case)

PR #145, opened 2020-07-10
WebLink: Deserializer: Release inbound object if not yet released

HaseenaSainul:
If we set a download timeout less than actual required time, then download operation will be stopped and will destruct all the classes involved with that. In that case, observed a memory corruption from operator_delete for the Web::Response from ~DeserializerImpl().  It seems that it is trying to do operator_delete on the already destructed Web::Response class due to wrong reference count. So adjusting the reference count using Release() to ensure both ref count updated and memory get released

### DIFF
--- a/Source/websocket/WebLink.h
+++ b/Source/websocket/WebLink.h
@@ -123,6 +123,9 @@ namespace Web {
             }
             virtual ~DeserializerImpl()
             {
+                if (_current.IsValid()) {
+                    _current.Release();
+                }
             }
 
         public:


pwielders:
Thats interesting, since what you are doing now in the destructor is part of the _current destructor, so it could not resolve the issue you are describing, a leak.
See: https://github.com/rdkcentral/Thunder/blob/master/Source/core/Proxy.h#L319-L321 the destruction of any ProxyType<OUTBOUND> object should trigger those lines and clear (Release) the object. However it is a fraction later (if the destructor ~DeserializerImpl has run all the members (and thus also _current) will be destructed.
So would be interested in what you actually fix with this.
Rereading your comment, I think the order of destruction in the WebTransfer might be the issue. Lets discuss this in detail on Skype.

HaseenaSainul:
Sure @pwielders, will reach you on Monday :)

HaseenaSainul:
Closing this PR, since this part is handled with the PR: https://github.com/rdkcentral/Thunder/pull/152

PR #150, opened 2020-07-14
[NETLINK] Packages coming in, in one o, shoul be handled (read) as we…

pwielders:
…ll..


mipo57:
Yup, from the current networking side there was no use case for parsing DONE message as everything is consumed on-the-fly, so it was treated only as a multi-message ending token. However, I guess it's a good idea to allow Netlink parsers to use that information that all of the info has been fetched :)

mipo57:
Probably it was used to end collecting slaves in 1-wire

mipo57 (approved):

PR #152, opened 2020-07-15
WebTransfer: Changes to handle large file download

HaseenaSainul:
1. Changes to support large file download
2. Handled download exit during the download progressing time
3. Error handling of Downloaded file write to the filesystem due to device space
4. Partition size info class added for FileSystem and interface exposed through FirmwareControl jsonrpc


Source/core/FileSystem.h
```diff
@@ -311,10 +317,11 @@ namespace Core {
             return (Create((USER_READ|USER_WRITE|USER_EXECUTE|GROUP_READ|GROUP_EXECUTE), exclusive));
         }
 
-        bool Create(const uint32_t mode, const bool exclusive = false)
+        bool Create(const uint32_t mode, const bool exclusive = false, bool large = false)
```
pwielders: What happens if by default we add this O_LARGEFILE flag and it is only a small file ?

Source/core/FileSystem.h
```diff
@@ -323,32 +330,35 @@ namespace Core {
             return (IsOpen());
         }
 
-        bool Open() const
+        bool Open(bool large = false) const
         {
 #ifdef __POSIX__
-            _handle = open(_name.c_str(), O_RDONLY);
+            int fileType = large? O_LARGEFILE: 0;
```
pwielders: See pervious remark :-)

Source/core/FileSystem.h
```diff
@@ -323,32 +330,35 @@ namespace Core {
...
 #endif
             const_cast<File*>(this)->LoadFileInfo();
             return (IsOpen());
         }
-        bool Open(bool readOnly)
+        bool Open(bool readOnly, bool large = false)
         {
 #ifdef __POSIX__
-            _handle = open(_name.c_str(), (readOnly ? O_RDONLY : O_RDWR));
+            int fileType = large? O_LARGEFILE: 0;
```
pwielders: And  here?

Source/core/FileSystem.h
```diff
@@ -323,32 +330,35 @@ namespace Core {
...
 #endif
             LoadFileInfo();
             return (IsOpen());
         }
-        bool Append()
+        bool Append(bool large = false)
         {
 #ifdef __POSIX__
-            _handle = open(_name.c_str(), O_RDWR | O_APPEND);
+            int fileType = large? O_LARGEFILE: 0;
```
pwielders: and here :-)

Source/core/FileSystem.h
```diff
@@ -432,7 +442,7 @@ namespace Core {
             DWORD writtenBytes;
 
             ::WriteFile(_handle, buffer, size, &writtenBytes, nullptr);
-            return static_cast<uint32_t>(writtenBytes);
```
pwielders: I think this line was oke. You are only allowing to read blocks which ara a maximum of uint32, so what you return will always be <= than 32 bits. so the static_cast was oke.

Source/core/FileSystem.h
HaseenaSainul: In 64bit machine, it will be anyway supporting large file if we pass FILE_OFFSET_BITS=64. But in 32-bit case we want to specify this flag to open a file if want more than 2GB size, but it any way create overheads with size and mapping. So I kept this as default param with false and can set whenever it is really required.

Source/core/FileSystem.h
HaseenaSainul: Ok @pwielders

Source/core/FileSystem.h
HaseenaSainul: Reverted back to the original code

Source/core/FileSystem.h
HaseenaSainul: Reverted back to the original code

Source/core/FileSystem.h
HaseenaSainul: Reverted back to the original code

Source/core/FileSystem.h
```diff
@@ -456,26 +462,26 @@ namespace Core {
...
 
 #ifdef __POSIX__
-            return (lseek(_handle, offset, (relative ? SEEK_CUR : SEEK_SET)) != -1);
+            return (LSEEK(_handle, offset, (relative ? SEEK_CUR : SEEK_SET)) != -1);
 #endif
 #ifdef __WINDOWS__
             return (::SetFilePointer(_handle, offset, nullptr, (relative ? FILE_CURRENT : FILE_BEGIN)) != INVALID_SET_FILE_POINTER);
 #endif
         }
         int32_t Position() const
```
pwielders: Should we now not return an int64_t ?

pwielders (commented):
I think the Position method on the File shouldaso retrn an int64_t i.s.o. a int32_t

Source/core/FileSystem.h
HaseenaSainul: Already resolved

mhughesacn:
Hi @HaseenaSainul.  Blackduck is complaining about the Hash.cpp file file  but you can disregard all the information about code matches in the gist because it is confusing.  The real compliance problem is to do with the code from Brad Conte.  Do you have an original source from Mr. Conte where he licenses it as MIT?  The file header in your source is not a proper MIT license. I find a source: https://github.com/B-Con/crypto-algorithms where the code is stated to be public domain.  Policy requires that I seek approval from RDK Management for the use of this code.

- Martin Hughes, RDK CMF Compliance Team

HaseenaSainul:
Thanks @mhughesacn, I taken this form the link https://review.haiku-os.org/c/haiku/+/439/3/src/kits/crypto/hash/sha256/sha256.c. The licence of whole haiku is under MIT, hence I specified it as MIT licensed. But it seems these are same as https://github.com/B-Con/crypto-algorithms. So I will change the reference to this link with out specifying any license, since there is no license/copyright mentioned in this link. Please comment on this.

mhughesacn:
Hi @HaseenaSainul  I'm not convinced that we can assume that the sha256 code is MIT licensed, so yes, it should not say so.  I need to get permission from RDK Management before I ask you to make some changes to do with attribution.  Usually they approve permissive PD software but it may have to go to Comcast Legal.  I did check if the same code was used elsewhere already in RDK code (i.e. already approved) but I couldn't find it. - Martin

HaseenaSainul:
Thanks @mhughesacn, will wait for your updates.

HaseenaSainul:
@mhughesacn, any updates on the usage code from Brad Conte ?

pwielders (dismissed):

pwielders:
@mhughesacn  would like to merge this, but blackduck is not flying and not telling why it is not flying :-) Help me revive the duck :-)

mhughesacn:
Hi @pwielders The duck is fine.  The problem is that some software is matching public domain software and our policy requires approval.  I asked Glee who has referred it to Comcast legal.

HaseenaSainul:
@mhughesacn, as per the mail from Ajith this is approved by Legal team, but blackjack check is still failing. Could you please help to resolve ?

mhughesacn:
Hi Haseena,
Comcast legal has finally approved this, so please will you update the top level NOTICE file to append the following:

sha256 code from Brad Conte (brad AT bradconte.com)
This code is released into the public domain free of any restrictions.

I am sorry about the delay, and will discuss internally what we can do to speed up the wheels.  Thanks, Martin

HaseenaSainul:
Thanks @mhughesacn, I updated the NOTICE file with license information as you suggested. Please crosscheck

jamcoole:
Hi Haseena,

Thanks for updating the NOTICE file, I have now approved the request, Martin will do the ID work in blackduck protex later.

Thanks
James

jamcoole:
Hi Haseena, sorry I didn't mean to close the request

HaseenaSainul:
Thanks @jamcoole

pwielders (approved):

pwielders:
@mhughesacn I think the legal stuff is arranged, but the duck is fly in low speed mode ? Can we shift gears on the duck :-)

jamcoole:
hi @pwielders  there is an issue with the latest scan failing to send gist to github so blackduck status is not being set, I have notified the sysadmins on the team, will try to get it resolved asap

pwielders:
@jamcoole thanks for increasing the flight speed of the duck :-)

jamcoole:
hi @pwielders  still investigating the issue so in meantime we have overridden the blackduck status, apologies for the delay

pwielders:
Thanks @jamcoole . Merged it !

PR #164, opened 2020-07-23
Fix communication channels for DisplayInfo

404ag:
This fixes segmentation faults in situations when DisplayInfo is already
in use (i.e by Netflix) and other plugins are trying to acccess it (i.e Tracing).

### DIFF
--- a/Source/com/Administrator.h
+++ b/Source/com/Administrator.h
@@ -358,14 +358,14 @@ namespace RPC {
             }
 
             if (message->Label() == AnnounceMessage::Id()) {
-	            _handler->Procedure(source, message);
-	    } else {
-
+                ASSERT(_handler != nullptr);
+                _handler->Procedure(source, message);
+            } else {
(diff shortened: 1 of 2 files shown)


artur-gebicz-ml:
PLEASE DO NOT MERGE !!!!!!

pwielders (approved):
Looking at the code changes, I do not understand where the segmentation fault was caused, here. Do agree that by *not* linking the announce message, the ITraces where not returned, but I think that the rootcause (other side *not* receiving an interface for ITracing information) is the rootcause and should be fixed (The other side should cater for not receiving a ITracing interface)

Also think that, at the end, this module "should" include the announce message handling to be able to trigger runtime traces that might be in this code, but I think with this patch we are "hiding" the bug, the core not being cable of receiving a nullptr as a responce on getting the ITrace information.

This is an analyses by reading through the code, could we please add a callstack to a Jira ticket to further investigate the rootcasue ?

PR #166, opened 2020-07-25
[Interfaces] Make JMath.h also a public header

MFransen69:


pwielders (commented):
Think this JMath should be removed and it should be generated from the IMath. The JMath.h was/is a working example for the code generator developper to see what I was expecting from the code generator to generate. Since this is completed (/* json */ keyword) this JMath.h should now be generated from through the IMath.h as well.

MFransen69:
> Think this JMath should be removed and it should be generated from the IMath. The JMath.h was/is a working example for the code generator developper to see what I was expecting from the code generator to generate. Since this is completed (/* json */ keyword) this JMath.h should now be generated from through the IMath.h as well.

Okay, agree but it is not something I plan to do right now. I'll remove all the functionality in the JSON RPC plugin that uses it at the moment to make it work again.

PR #167, opened 2020-07-26
[JSON RPC] fix version handling

MFransen69:
- Repaired sending requested version in LinkType
- Repaired that adding duplicate method handler/callback will overwrite old one again (needed to make version handling work as the one being registered for this version needs to overwrite the one that can exist as it was copied from a previous version)

### DIFF
--- a/Source/websocket/JSONRPCLink.h
+++ b/Source/websocket/JSONRPCLink.h
@@ -583,12 +583,18 @@ namespace JSONRPC {
             , _localSpace()
             , _pendingQueue()
             , _scheduledTime(0)
+            , _versionstring()
         {
             if (localCallsign == nullptr) {
 	        static uint32_t sequence;
                 _localSpace = string("temporary") + Core::NumberType<uint32_t>(Core::InterlockedIncrement(sequence)).Text();
             } else {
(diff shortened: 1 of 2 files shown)


Source/websocket/JSONRPCLink.h
```diff
@@ -583,13 +583,19 @@ namespace JSONRPC {
...
+            , _versionstring()
         {
             if (localCallsign == nullptr) {
 	        static uint32_t sequence;
                 _localSpace = string("temporary") + Core::NumberType<uint32_t>(Core::InterlockedIncrement(sequence)).Text();
             } else {
                 _localSpace = localCallsign;
             }
+
+            uint8_t version = Core::JSONRPC::Message::Version(callsign + '.');
```
pwielders, line 595: Really needed the addition of the '.' to the callsign ?

pwielders (approved):
Just check the commentt, the added '.' seems strange.

Source/websocket/JSONRPCLink.h
MFransen69: Yes, otherwise the Version method does not work correctly (and I did not want to change that)

PR #170, opened 2020-07-28
Added option to disable compiler exception handling

HaseenaSainul:


cmake/modules/CompileSettings.cmake
```diff
@@ -20,6 +20,7 @@ This is a non-buildable target used to set the overall mendatory settings for th
 compiler for all sources of the Framework and its Plugins
 #]]
 option(POSITION_INDEPENDENT_CODE "Create position independent code on all targets including static libs" ON)
+option(EXCEPTIONS_DISABLE "Disable exception handling" ON)
```
MFransen69: To make it easier to understand perhaps better not to use inverse logic, for some people a bit harder to understand, and use "EXCEPTIONS_ENABLE" etc.

MFransen69 (changes requested):

cmake/modules/CompileSettings.cmake
HaseenaSainul: Ok, by default the exception flag is on, hence made it reverse

HaseenaSainul: Made the required changes

bramoosterhuis (approved):

MFransen69 (approved):

PR #171, opened 2020-07-28
[JSONRPC] Make SmartLinkType compile again

MFransen69:
Didn't know if tthe difference in wait times was deliberate. For now kept it functional equal to how it was but if can be replaced with one default wait time that would be easier, So if that is the case please leave a review comment and let me change it.
Also access has been kept in sync with how it was before.


pwielders (approved):
WaitTime was not done deliberately. Seems to be a bug, please change!

PR #207, opened 2020-07-29
[NetworkControl] Moved IPFrame and UDPFrame to core

mipo57:
Moved some IP & UDP frame logic to core.
@pwielders not sure if that is what you meat, that's why I do a PR on a branch rather than commit


pwielders (approved):
Hi @mipo57 .

This is indeed what I meant. It is still very limited, but now it can grow *(extend interfaces (TTL is one that is interested as it is used in the Probe, maybe time for a probe rewrite :-) but it can be extended in the near futeur.

Again, Michal, thanks for all the contributions and good luck with your studies and career!

PR #210, opened 2020-08-01
Move IPUDPPacket to core

MFransen69:


pwielders:
Already part of the pullrequest: https://github.com/rdkcentral/Thunder/pull/221 after this is applied, we can update the IPFrame here

PR #212, opened 2020-08-05
Development/explora/av caps/2 master

404ag:


pwielders (changes requested):
As discussed, move the dolby detection to the QueryInterface mechanism. SO only dolby supporting systems will be effected.

pwielders:
Implemeneted (on interface level) by #288

PR #214, opened 2020-08-07
XRE-14362: Adding properties to VAP plugins

vinod-jacob:
Reason for change: Adding properties required by Premium apps
for migration to rdkservices
Changes:
1. Added JSONRPC & COM support for additional
propoerties in DisplayInfo & PlayerInfo
Test Procedure: check ticket
Risks: Low
Signed-off-by: Vinod Jacob <vinod_jacob@comcast.com>


Source/interfaces/IPlayerInfo.h
```diff
@@ -78,5 +78,27 @@ namespace Exchange {
         virtual IAudioIterator* AudioCodec() const = 0;
         virtual IVideoIterator* VideoCodec() const = 0;
     };
+
+    struct EXTERNAL IPlayerAtmosProperties : virtual public Core::IUnknown {
```
pwielders: Atmos is part of the Dolby eco system. Sugegst to put this in the Dolby ecosystem:
https://github.com/rdkcentral/Thunder/blob/master/Source/interfaces/IDolby.h
I think there you will find that the setAudioAtmosOutputMode is already more specific defined.

Also our coding rules require methods to be PascalCase and not CamelCase. If you do not have the coding Rules, let me know, I can share them.

Source/interfaces/json/DisplayInfo.json
```diff
@@ -71,12 +71,203 @@
           "hdrtype"
         ]
       }
+    },
+    "videoPlaybackResolution": {
```
pwielders: Thunder is designed with consistency in mind. Take note that currently "all" JSON interfaces are designed lower case (tools will automatically make them lowercase) definigin it in the source as camelcase might confuse development. Please comply to the coding guidelines/rules.

Source/interfaces/IDisplayInfo.h
```diff
@@ -33,6 +33,34 @@ namespace Exchange {
...
+            virtual ~INotification() {}
+
+            virtual void ResolutionChanged(const int32_t&, const int32_t&, const string&, const string& ) = 0;
+            virtual void ResolutionPreChange() = 0;
+        };
+
+        virtual uint32_t Register(INotification* sink) = 0;
+        virtual uint32_t Unregister(INotification* sink) = 0;
+
+        virtual string getCurrentResolution() const = 0;
```
pwielders: Coding rules. Use PascalCase for methods.

pwielders (changes requested):
Coding rules: Methods are PascalCase.
Interface design: Using QI, interfaces are split into functional groups. Please adhere to the designs/architectural patterns and do not create GOD interfaces.

Source/interfaces/IDisplayInfo.h
```diff
@@ -33,6 +33,34 @@ namespace Exchange {
...
+        virtual string getConnectedVideoDisplay() const = 0;
+        virtual void getTvHdrCapabilities(string& hdrcaps /* @inout */, int& caps /* @inout */) const = 0;
```
pwielders: The design/architectural principles are to create small granular blocks of functionality that can be reused for multiple customers and guide developpers during development.
HDMI functionality is part of the IConnectionProperties, so please put functionality for the connection on the IConnectionProperties.
As you can see the Iconnection properties has already some HDR properties. So I think the added methods here are actually part of the IConnection Properties.
From a development perspective, it is good to be specific. Why not use the enums for HDR capabilities ?
This methods are quite ambigous. Why should you send hdrcaps to it and receive hdrcaps from the implementation if you are getting ? I would expect here only /* @out */ ? Also if it is a list (that you are relaying over the interface) suggest iterators (See https://github.com/rdkcentral/Thunder/blob/master/Source/com/IRPCIterator.h using enums as a template paramter) allowing for strong typed feedback.

Source/interfaces/IPlayerInfo.h
pwielders: Currently we are adding the IVideo Atmos functionality as a Dolby feature. Good to align maybe so these two would work together. See pull request: https://github.com/rdkcentral/Thunder/pull/212 but also there the remark to move to the dolby space :-)

Source/interfaces/IDisplayInfo.h
vinod-jacob: Done

Source/interfaces/json/DisplayInfo.json
vinod-jacob: Done

Source/interfaces/IPlayerInfo.h
pwielders: The concept of the IUnknown (comrpc) allows for interface navigation. Through interface navigation you can "check" if certain features are available or not (you get the interface or you do not get it). So there is no need to create additional or other plugin for the dolby. The PlayerInfo plugin and the DisplayInfo plugin will just by extended with IDolby*. I would even strongly suggest to use the same plugin to not scatter the logically the same information over different plugins.
As you can see these plugins have a sub folder where the implementation resides. As discusses, you can create an IARM implementation (since you delegate the retrieval of this information from yet another component). 
If for your platofrm you do not support the Dolby, it just means that the IDolby interface return null and there is no implementation for that platform (e.g. RPI will not have an IDolby implementation) 
So I think the interface navigation is already handled by the base system

Source/interfaces/IDisplayInfo.h
pwielders: > My intention behind adding all the new APIs under IDisplayProperties was in fact to keep the IConnectionProperties untouched. The reason being that if I add more functions to IConnectionProperties, it would instantly break the AmLogic, Rpi and Nexus implementations as these are derived from IConnectionProperties (unless stubs are written for them).

Left or right you need to add those properties to either the IDisplayProperties or to the IConnectionProperties for all the platforms. As we say in Holland: "Das lood om oud ijzer". So from an implementation perspective it is the same amount of work.

pwielders: > Surely, I could've done the same and added HLG to the enum list. But, that begged the question about how we keep this list exhaustive? For instance, tomorrow my device could return a new HDR format - say HDR20. In such a case I would need to add HDR20 as an enum inside Thunder before I can use it inside DisplayInfo plugin. Moreover, it would essentially result in constantly updating the HDR enum list within Thunder to keep it in sync with the enum list within DeviceSettings

In code you can only "act" on something if you know what it is. If a new standard is set forth, it means that new processing is required. The fact that the enum value for this future feature, and thus future processing is not available, is a good feature. If for example, somehwere you need to translate the HDR type/setting from the input to the enum and you can not convert it to that enum, I suggest rasing a serious log line there. If you do not prefer to log it there but in you plugin (I would agree to that) I would add a HDR catch all enum kuje HDR_UNKNOWN value. If you receive this value, you could raise an error n the plugin, like: Encountered an HDR mode for which I have no handling code. Than anyone reading the logs can already see why a HDR feature is not working. 
Also if it is a string, all code using this information   need to parse this string to determine what the should do with it.
As a rule of Thumb: Thunder development is DEFENSIVE. Which means use lots of ASSERTS where you know you are not catering for those scenarios and be TYPE SAFE. If possible use strong typed values (like enums) and not opaque objects like strings in which you could put in anything.
According to the V-Model of software development, the sooner you detect a bug (missing functionality) to sooner you raise it, the easier and cheaper it is to resolve it.

pwielders: And if I look at what we already have, it looks like the list is quite complete already:
https://github.com/rdkcentral/Thunder/blob/master/Source/interfaces/IDisplayInfo.h#L41-L47
Do like the addion of invalid, whihc I guess is the UNKNOWN.

Source/interfaces/IDisplayInfo.h
```diff
@@ -60,6 +81,26 @@ namespace Exchange {
...
+                HDR_10PLUS,
+                HDR_UNKNOWN
+            };
+
+            virtual bool IsValid() const = 0;
+            virtual bool Next() = 0;
+            virtual void Reset() = 0;
+            virtual HdrFormats HdrFormat() const = 0;
+        };
+
```
vinod-jacob, line 103: @pwielders, Based on your comments https://github.com/rdkcentral/Thunder/pull/214#discussion_r467864644 ,
I have introduced HDR iterator to relay the HDR formats over this interface.

Source/interfaces/IDisplayInfo.h
```diff
@@ -70,6 +111,11 @@ namespace Exchange {
         virtual uint32_t VerticalFreq() const = 0;
         virtual HDRType Type() const = 0;
         virtual HDCPProtectionType HDCPProtection() const = 0;
+        virtual int GetHdmiPreferences() const = 0;
+        virtual void SetHdmiPreferences(const int HdcpProtocol) = 0;
+        virtual IHdrIterator* GetTvHdrCapabilities() const = 0;
+        virtual IHdrIterator* GetStbHdrCapabilities() const = 0;
+        virtual bool IsOutputHDR() const = 0;
```
vinod-jacob, line 118: @pwielders, Based on your suggestion at https://github.com/rdkcentral/Thunder/pull/214#discussion_r467864644 ,
I have moved the HDR and HDMI related methods to the Connection Interface. Could you please confirm these changes are aligned with  your expectations

Source/interfaces/IDisplayInfo.h
```diff
@@ -60,6 +82,26 @@ namespace Exchange {
...
+                HDR_10PLUS,
+                HDR_UNKNOWN
+            };
+
+            virtual bool IsValid() const = 0;
+            virtual bool Next() = 0;
+            virtual void Reset() = 0;
+            virtual HdrFormats HdrFormat() const = 0;
+        };
+
```
vinod-jacob: @pwielders, based on your comment https://github.com/rdkcentral/Thunder/pull/214#discussion_r467864644, I have introduced iterator type to relay data over GetTvHdrCapabilities &  GetStbHdrCapabilities methods of this interface. Could you please confirm if these changes are in alignment with your expectations.

Source/interfaces/IDisplayInfo.h
```diff
@@ -70,6 +112,9 @@ namespace Exchange {
         virtual uint32_t VerticalFreq() const = 0;
         virtual HDRType Type() const = 0;
         virtual HDCPProtectionType HDCPProtection() const = 0;
+        virtual void HDCPProtection(const HDCPProtectionType) = 0;
+        virtual IHdrIterator* GetTvHdrCapabilities() const = 0;
+        virtual IHdrIterator* GetStbHdrCapabilities() const = 0;
```
vinod-jacob: @pwielders, Based on your comment https://github.com/rdkcentral/Thunder/pull/214#discussion_r467864644, I have moved the HDR & HDMI related methods to Connection Properties interface, Could you please confirm if these changes are in alignment with your expectations?

Source/interfaces/IDisplayInfo.h
pwielders: I think there is no need to create such an interface (also from a performance perspective) it is easier (simpler and less coder) to just define:

    typedef RPC::IIteratorType<HdrFormats, ID_STRINGITERATOR> IHDRIterator; // This is the interface definition
    typedef RPC::IteratorType<IHDRIterator> HDRIterator;  // This is the interface definition and declarion.

After these two definitions, you have the interface definition and an implemntation ffor it. Just instantiate it by supplying, on creation, the following syntax:

        virtual IHDRIterator* HDRSystemsSupported() const
        {
            return (Core::Service<HDRIterator>::Create<IHDRIterator>(_hdrSystems));
        }

Where _hdrSystems is the std::list<HdrFormats> _hdrSystems;

And done :-)

pwielders:
I guess, these remarks where from before we had the skype meeting, so I think the next step is to first determine the interface. For the iterators, please see the: https://github.com/rdkcentral/Thunder/pull/214#discussion_r470184523

Source/interfaces/IDisplayInfo.h
```diff
@@ -33,43 +33,108 @@ namespace Exchange {
...
+
+        // @property
+        // @brief HDR formats supported by TV
+        // @return HDRType: array of HDR formats
+        virtual uint32_t TVCapabilities(IHDRIterator& type /* out */) const;
+
+        // @property
+        // @brief HDR formats supported by STB
+        // @return HDRType: array of HDR formats
+        virtual uint32_t STBCapabilities(IHDRIterator& type /* out */) const;
```
vinod-jacob: @pwielders : I wasn't very sure about the IIterator datatype. There were no other usages of this dataype inside Thunder. Could you please have a careful look to see if I've used it correctly?

pwielders: Hi Vinod,
This really looks good. Indeed the returning of the reference is not a good idea since this can run over process boundaries. If the otherside crashes, you could still call this but than the interface to return should be a null ptr. So I think you really should return a pointer here i.s.o. references.

Source/interfaces/IDolby.h
```diff
@@ -38,9 +38,50 @@ namespace Exchange {
                 DIGITAL_AC3 = 4,
                 AUTO = 5
             };
-            
+
+            enum SoundModes : uint8_t {
```
pwielders, line 42: I have the impression (but you need to confirm it on ComCast side) That the enum type (just above here)  and SoundModes are the same, just different (or more values) If so, I think we can than combine

Source/interfaces/IPlayerInfo.h
```diff
@@ -23,60 +23,67 @@
...
+            RESOLUTION_1080I,
+            RESOLUTION_1080P,
+            RESOLUTION_2160P30,
+            RESOLUTION_2160P60
         };
 
-        virtual IAudioIterator* AudioCodec() const = 0;
-        virtual IVideoIterator* VideoCodec() const = 0;
+        // @property
+        // @brief Current Video playback resolution set from EPG
```
pwielders: The interface documentation is generated from the @brief yu add here. I think it is important to leave implenataion notes (how it works on ComCast STB where it is coming from, like EPG) out of the comments.
At the end the only think we do here is report the Resolution used by the player. 
How you set it is a different story and not part of this interface.

pwielders (changes requested):
Just a few minor remarks (text on @brief and changing reference to pointer) and a deeper check on overlapping atmust (SoundMode and type)

Source/interfaces/IDolby.h
vinod-jacob: RDK actually maintains them as separate enums.

The Type enum here is analogous to our "audio encoding" enum,

```
/**
 * This enumeration defines the type of audio encoding that can be used.
 */
typedef enum _dsAudioEncoding_t{
    dsAUDIO_ENC_NONE = 0,        /**< No digital audio output.                         */
    dsAUDIO_ENC_DISPLAY,     /**< Platform-selected digital audio encoding format. */
    dsAUDIO_ENC_PCM,         /**< PCM digital audio encoding format.               */
    dsAUDIO_ENC_AC3,         /**< AC-3 digital audio encoding format.               */
    dsAUDIO_ENC_EAC3,        /**< DD+/E-AC-3 digital audio encoding format.               */
    dsAUDIO_ENC_MAX          /**< Maximum index for audio encoding types.          */
} dsAudioEncoding_t;
```

Whereas SoundMode is maintained separately
```
/**
 * This enumeration defines the type of audio stereo mode that can be used.
 */
typedef enum StereoMode{
    dsAUDIO_STEREO_UNKNOWN,   /**< Unknown mode.                         */
    dsAUDIO_STEREO_MONO = 1,  /**< Mono mode.                            */
    dsAUDIO_STEREO_STEREO,    /**< Normal stereo mode (L+R).             */
    dsAUDIO_STEREO_SURROUND,  /**< Surround mode.                        */
    dsAUDIO_STEREO_PASSTHRU,  /**< Passthrough mode.                     */
    dsAUDIO_STEREO_MAX        /**< Maximum index for audio stereo types. */
} dsAudioStereoMode_t;

```
I believe it is best to retain Types and SoundModes as separate

pwielders: Oke

Source/interfaces/IDolby.h
```diff
@@ -38,9 +38,44 @@ namespace Exchange {
...
+
+            // @property
+            // @brief Sound Mode - Mono/Stereo/Surround
+            // @return mode: sound mode
+            virtual uint32_t SoundMode(Dolby::IOutput::SoundModes& mode /* @out */) const = 0;
+
+            // @property
+            // @brief Enable Atmos Audio Output
+            // @param enable: enable/disable
+            virtual uint32_t EnableAtmosOutput(const bool& enable /* @in */);
```
pwielders: I guess you just forgot but it should be a pure abstract method (= 0)

"Dolby Digital Plus is a digital audio compression scheme. On the other hand, Dolby Atmos is the name of a surround sound technology." I don't think it is correct to treat DD+ and Atmos as the same thing. It is possible to have DD+ compression (aka E-AC-3) without Atmos MetaData . So, in that sense, Mode(DIGITAL_PLUS); Does not imply turning on Atmos Dolby.

Under the given assumption as stated above, I think it is fine to add this method since Mode(DIGITAL_PLUS) is not assumed to be "enabling" the Atmos.

pwielders (dismissed):

pwielders:
@vinod-jacob started to implement these interfaces for the Metrological devices (NEXUS/RPI) and ran into some issues with the interfaces. I created a branch to develop on. Suggest we continue on this branch:
https://github.com/rdkcentral/Thunder/tree/development/InfoPlugins

pwielders:
Replaced by pull request #288

PR #223, opened 2020-08-18
Added Dobby backend for Thunder containers

Koczek9:
Added Dobby as source for containers in processcontainers.


Source/processcontainers/CMakeLists.txt
```diff
@@ -96,6 +100,49 @@ elseif (PROCESSCONTAINERS_CRUN)
                 PRIVATE
                 libcrun::libcrun
         )
+elseif (PROCESSCONTAINERS_DOBBY)
+        find_package(jsoncpp REQUIRED)
```
pwielders: Thunder already has a JSON Serializer/Deserializer, Why do you include yet another JSON library. This adds an unneeded dependency to the Thunder libraries. Please just use Core (which is already part of this setup)

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -0,0 +1,304 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "DobbyImplementation.h"
+#include "JSON.h"
```
pwielders: Since JSON handling is part of Core, you can remove this header file than :-)

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -0,0 +1,304 @@
...
+            
+            std::string containerInfoString = admin.mDobbyProxy->getContainerInfo(_descriptor);
+
+            if (containerInfoString.empty())
+            {
+                TRACE_L1("Failed to get info for container %s", _name.c_str());
+            }else{
+                // Dobby returns the container info as JSON, so parse it
+                JsonObject containerInfoJson;
+                WPEFramework::Core::OptionalType<WPEFramework::Core::JSON::Error> error;
```
pwielders, line 184: uhhh now you lost me :-) here you are using the Thunder JSON parsing stuff... So why include this jsoncpp ?

Source/processcontainers/CMakeLists.txt
```diff
@@ -96,6 +100,49 @@ elseif (PROCESSCONTAINERS_CRUN)
...
+        find_package_handle_standard_args(
+                LIBSYSTEMD
+                SYSTEMD_LIBRARIES SYSTEMD_INCLUDE_DIRS
+        )
+        target_link_libraries(${TARGET}
+                PRIVATE
+                # Dobby libraries
+                DobbyClientLib
+                IpcService
+                AppInfraCommon
```
pwielders, line 128: WHy is the dependency on AppInfrCommon Needed ? Is the Dobby package not self-containing ?

Source/websocket/WebTransfer.h
```diff
@@ -508,7 +508,7 @@ namespace Web {
 
             // Read all Data to calculate the HASH/HMAC
             while (size > 0) {
-                uint16_t chunk = std::min(sizeof(buffer), size);
+                uint16_t chunk = std::min(sizeof(buffer), (size_t)size);
```
pwielders: Coding rule violation, no C-Style casting, only C++ style casting:
This should be: uint16_t chunk = std::min(static_cast<uint16_t>(sizeof(buffer)), size);

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.h
```diff
@@ -0,0 +1,92 @@
+/*
```
pwielders, line 1: Do not think we need to include this. Can be taken out of the equation as the current JSON processing is only using the Thunder JSON processing.

cmake/modules/Findjsoncpp.cmake
```diff
@@ -0,0 +1,51 @@
+# If not stated otherwise in this file or this component's license file the
```
pwielders: Do not think we need to include this. Can be taken out of the equation as the current JSON processing is only using the Thunder JSON processing.

Koczek9:
I have fixed most of the issues you have pointed out. I even managed to get rid of Jsoncpp refrence at all (this requires pull request of https://github.com/rdkcentral/Dobby/pull/21 to be merged to Dobby, but this should be done today). I am still investigating why AppInfraCommon and AppInfraLogging aren't included even though DobbyClientLib is built statically.

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    bool DobbyContainer::Start(const string& command, IStringIterator& parameters)
+    {
+        bool result = false;
+        auto& admin = static_cast<DobbyContainerAdministrator&>(DobbyContainerAdministrator::Instance());
+
+        std::list<int> emptyList;
+
+        // construct the full command to run with all the arguments
+        std::string fullCommand = "/usr/bin/" + command;
```
goruklu, line 251: I think we can remove "/usr/bin" prefix and rely on PATH

Source/processcontainers/CMakeLists.txt
pwielders: This is not a blocking issue. It is bad practise :-)
Components metrological develops (and reviews) are self containing so the user of the library does not have to dig around to find potential other dependencies DobbyClientLib should include.
Dependencies could be managed through pkgconfig files. If in the future the dependency list changes for the  DobbyClientLib, you just change the pkconfig file that you generate with the DobbyClientLib artifact. Than this CMakeList.txt file does *not* require changes, as it just picks up the pkgconfig and the dependencies listed in there.
Now, with this setup, if you change the dependency list of the DobbyClientLib (I assume that is the library found through find_package(Dobby REQUIRED CONFIG)) you have to change this projects that is just "using" this library as well. I would expect that to be unnecessary and difficult if you start running different versions of the DobbyClientLib with different dependencies.

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
pwielders: If you are concerned about the usability outside of the Sky domain, please also not line 67 in this file which has hardcoded values.

goruklu:
@Koczek9 can you actually update the code to launch from an OCI bundle (config.json + rootfs) ? you may even drop support for launching from dobbyspec I think.

Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
goruklu: @pwielders I think we are OK with the line 67 hard coded values for now. 
Can we merge this PR soon ? thanks!

pwielders:
For now this will work in the Sky domain (hard coded paths in there, even stating the operator name Sky) , it requires changes in code to be used for other operators.
The code does not meet the code quality rules applied to Metrological code.
Violations on:
**Portability/Usability** => Hard coded paths and operator names (Sky)
**Maintainability** => Changes in the Dobby archive need to be reflected into this projects CMakeList.txt i.s.o. in a pkgconfig that is associated with the dobbby archive.

Since there are sufficient alternatives for RDK operators to run containers (LXC/RUNC and CRUN) it is fine to merge.

pwielders (approved):
Not compliant to the Metrological Code Quality rules but since it is limited to the Sky project, good to go.

goruklu:
@pwielders thanks for reviewing and approving this PR. Can you please merge it ?

PR #229, opened 2020-08-24
Made postfix name configurable for Persistent&Volatile Path of plugins

### LINKED ISSUE #154: Add option to configure plugin to use data/persistent/volatile paths of 'main' callsign
For example, WebKitBrowser plugin has multiple 'alias' callsigns that use Shared Data/Persistent/Volatile paths derived from their name.

The request is to have a way to define/configure what 'alias' callsigns should be using.

HaseenaSainul:


Source/plugins/Configuration.h
```diff
@@ -102,10 +112,12 @@ namespace Plugin {
             return (basePath + ClassName.Value() + '/');
         }
         string PersistentPath(const string& basePath) const {
-            return (basePath + Callsign.Value() + '/');
+            string postfixPath = (PersistentPathPostfix.IsSet() == true) ? PersistentPathPostfix.Value(): Callsign.Value();
```
pwielders: Can we also check for not empty.

Source/plugins/Configuration.h
```diff
@@ -102,10 +112,12 @@ namespace Plugin {
...
+            return (basePath + postfixPath + '/');
         }
         string VolatilePath(const string& basePath) const {
-            return (basePath + Callsign.Value() + '/');
+            string postfixPath = (VolatilePathPostfix.IsSet() == true) ? VolatilePathPostfix.Value(): Callsign.Value();
```
pwielders: Can we also check for not empty.

pwielders (changes requested):

Source/plugins/Configuration.h
HaseenaSainul: Changes added

Source/plugins/Configuration.h
HaseenaSainul: Changes added

pwielders (approved):

PR #236, opened 2020-08-27
DebugOptimzation with set Optimization flag to avoid overriding it from the build environment

HaseenaSainul:
In Yocto, RPI RDK build always used -Os optimisation level, this causing issue for Thunder debug build. 
So added option to set 0 optimisation level debug build to avoid this issues with different build system environment values

### DIFF
--- a/cmake/modules/CompileSettingsDebug.cmake
+++ b/cmake/modules/CompileSettingsDebug.cmake
@@ -30,7 +30,7 @@ include(CMakePackageConfigHelpers)
 #
 if("${BUILD_TYPE}" STREQUAL "DebugOptimized")
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_COMPILER_IS_GNUCXX} )
-        target_compile_options(CompileSettingsDebug INTERFACE -g)
+        target_compile_options(CompileSettingsDebug INTERFACE -O0 -g)
     endif()
 
 elseif("${BUILD_TYPE}" STREQUAL "ReleaseSymbols")
--- a/cmake/modules/CompileSettingsDebug.cmake
+++ b/cmake/modules/CompileSettingsDebug.cmake
@@ -30,12 +30,12 @@ include(CMakePackageConfigHelpers)
 #
 if("${BUILD_TYPE}" STREQUAL "Debug")
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_COMPILER_IS_GNUCXX} )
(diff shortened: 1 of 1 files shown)


404ag (dismissed):
LGTM

pwielders (commented):
I would expect that what you are now trying to achive is the BUILD_TYPE == "Debug". As this is called "Debug*Optimized*" I think it is correct that the O0 flag is not in there.
@bramoosterhuis  is tehre a "Debug" that does this ?

HaseenaSainul:
I found this issue while enabling debubOptimized option for RPI yocto build.

HaseenaSainul:
Same issue with Debug BuildType as well. So need to avoid the -Os level for debug* build types

cmake/modules/CompileSettingsDebug.cmake
```diff
@@ -28,9 +28,14 @@ include(CMakePackageConfigHelpers)
...
-if("${BUILD_TYPE}" STREQUAL "DebugOptimized")
+if("${BUILD_TYPE}" STREQUAL "Debug")
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_COMPILER_IS_GNUCXX} )
-        target_compile_options(CompileSettingsDebug INTERFACE -g)
+        target_compile_options(CompileSettingsDebug INTERFACE -O0)
+    endif()
+
+elseif("${BUILD_TYPE}" STREQUAL "DebugOptimized")
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_COMPILER_IS_GNUCXX} )
+        target_compile_options(CompileSettingsDebug INTERFACE -O1 -g)
```
pwielders: @bramoosterhuis / @HaseenaSainul just double checking: I would like to have the same optimization level in Release as in Debig so we test the same in both environments. I thought that the Release Optimisation level is 2 (-O2) if so, can we change it to -O2, if it is -O1 this pull request is good to go..

bramoosterhuis: I was hoping these flags would come from the build system. If we need to set them our selves I guess we want these flags:
DebugOptimized: "-O2 -g"
Debug: "-O0 -g"

HaseenaSainul: Yes,  but usage of -Os flag from build system is creating build issues. So think it is better to define the optimisation level from thunder itself. So update the flags as per your comments

pwielders (approved):

PR #237, opened 2020-08-27
Validate OpenCDMAccessor::Instance() before accessing it

varumugam123:
It is possible for OpenCDMAccessor::Instance() to return a NULL, when the below connection request fails / timesout.
` _remote = _client->Open<OCDM::IAccessorOCDM>(_T("OpenCDMImplementation"));`

It would be good to handle that gracefully in the libocdm.


pwielders (approved):

varumugam123:
Hi @pwielders, what is the next step? Will this PR be automatically merged or it should be merged by somebody?

pwielders:
I have to wait fro the branch to be updated and than I merge it :-)

PR #240, opened 2020-08-27
[PLUGINSERVER] Change the order of the Deactivation of the plugins

SKumarMetro:


MFransen69:
In the mean time we have changed out structure and are now handling the preconditions, termination conditions and subsystems handled by a plugin in code.

PR #243, opened 2020-08-28
Update accessor info only after network subsystem ready

HaseenaSainul:
Using current code DIALServer is not working during the device boot up time. During the analysis found during the boot up it always endup with the error no interface ready: [here](https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/Config.h#L551). Hence the Accessor URL for the  plugin services are empty and DIALServer is creating wrong commands.

With this changes, update the accessor info once the network event is ready, so that means at least one interface is ready.

### DIFF
--- a/Source/WPEFramework/SystemInfo.cpp
+++ b/Source/WPEFramework/SystemInfo.cpp
@@ -35,6 +35,7 @@ namespace PluginHost {
         , _time(nullptr)
         , _provisioning(nullptr)
         , _flags(0)
+        , _type(END_LIST)
     {
         ASSERT(callback != nullptr);
     }
--- a/Source/WPEFramework/SystemInfo.cpp
+++ b/Source/WPEFramework/SystemInfo.cpp
@@ -35,7 +35,6 @@ namespace PluginHost {
         , _time(nullptr)
         , _provisioning(nullptr)
         , _flags(0)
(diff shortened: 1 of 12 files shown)


Source/WPEFramework/SystemInfo.cpp
```diff
@@ -35,6 +35,7 @@ namespace PluginHost {
         , _time(nullptr)
         , _provisioning(nullptr)
         , _flags(0)
+        , _type(END_LIST)
```
pwielders: I think I am missing something? What goal does this _type serve ?

pwielders (commented):
I think too much code is checked in ?

Source/WPEFramework/SystemInfo.cpp
HaseenaSainul: Just to get the current event type fired. From the callback will use this type and updating the accessor details if the current event type is Network event

HaseenaSainul:
Yea these changes added to do the update accessor once network event fired

Source/WPEFramework/SystemInfo.cpp
HaseenaSainul: Moved this logic to PluginServer.h as per discussion

pwielders (approved):

PR #254, opened 2020-09-03
 DELIA-44794- [Thunder] string serialization issue with embedded ne…

raziththamim:
update the code changes in JSON.h  for DELIA-44794


pwielders:
@MFransen69 I think this does not work ? it is taken a char and you fill it with 2 characters??? I would expect a compile error? Maybe I misinterprete?

MFransen69:
> @MFransen69 I think this does not work ? it is taken a char and you fill it with 2 characters??? I would expect a compile error? Maybe I misinterprete?

Yes I guess the code will not result in what the intent is. Clang and gcc give warnings, not errors (indeed also would have expected an error): "warning: multi-character character constant" and "warning: implicit conversion from 'int' to 'char' changes value from 23662 to 110". MSVC apparently does not (at least in the online version I tried)

Source/core/JSON.h
```diff
@@ -1934,19 +1934,19 @@ namespace Core {
                 char replacement = current;
                 switch (current) {
                 case 'n':
-                    replacement = '\n';
+                    replacement = '\\n';
```
MFransen69: I guess it is not the intent of the code the new value is capped, please see comments in the Pull Request for details.

MFransen69 (changes requested):

pwielders (changes requested):
Currently it generates warnings. Also the intention is not clear. Can you fix the warnings and describe the issue and how you think you fixed it?

pwielders:
Code technically this might now fix an issue, but looking at the whole picture, my guess, not even testing this, is that this whole method should be removed. I do not think that the error is only causing issues with the /n, I would expect that it is also applicable than to /r /t /f /b. so I guess we should run this through our Testcases and see if indeed this whole method can be removed. I think the current patch is a partial fix (only for /n not the other values) or the parser is used incorrectly.
Can you share the code that is fixed by this partial fix ?

pwielders:
Sorry wrong button...

raziththamim:
> Code technically this might now fix an issue, but looking at the whole picture, my guess, not even testing this, is that this whole method should be removed. I do not think that the error is only causing issues with the /n, I would expect that it is also applicable than to /r /t /f /b. so I guess we should run this through our Testcases and see if indeed this whole method can be removed. I think the current patch is a partial fix (only for /n not the other values) or the parser is used incorrectly.
> Can you share the code that is fixed by this partial fix ?

 @pwielders  ,  as per test cases mentioned in the ticket. in last fix i have replaced the  '\n' with 'n' . it  leads to changed the JSON response in output. for example. if input  string was ma\n, the output JSON response is changed to "man". we once again discussed with DELIA-44794 reporter . confirmed that hard newlines should be replaced with \n. so that JSON deserialization can get back to the original newlines. we are working on that \n issue. and also check those \t \r \f \b escape sequence with test cases with small & long strings. we will let you know

raziththamim:
Test Case validated after fix:

For Newline (\n):
Input string: curl -H "Content-Type: application/json"  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.aHR0cDovL2xvY2FsaG9zdA.mLd2vW6_3hHMahhbQCaVpgkuB5JM_GuMaxe4hmxgDl8" -X POST   -d '{"jsonrpc":"2.0","id":"3","method": "org.rdk.Network.1.testStringserial", "params":{"teststring":"\n solution \n for \n string \n serialization\n","iWa":false}}' http://127.0.0.1:9998/jsonrpc

Output String:root@arrisxi6:~#
7.0.0.1:9998/jsonrpcerial", "params":{"teststring":"\n solution \n for \n string \n serialization\n","iWa":false}}' http://1
{"jsonrpc":"2.0","id":3,"result":{"teststring":"\n solution \n for \n string \n serialization\n","iWa":false,"success":true}

==============================================================================================

For Tab (\t):

Input String: curl -H "Content-Type: application/json"  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.aHR0cDovL2xvY2FsaG9zdA.mLd2vW6_3hHMahhbQCaVpgkuB5JM_GuMaxe4hmxgDl8" -X POST   -d '{"jsonrpc":"2.0","id":"3","method": "org.rdk.Network.1.testStringserial", "params":{"teststring":"\t solution \t for \t string \t serialization\t","iWa":false}}' http://127.0.0.1:9998/jsonrpc

Output String:root@arrisxi6:~#
7.0.0.1:9998/jsonrpcerial", "params":{"teststring":"\t solution \t for \t string \t serialization\t","iWa":false}}' http://1
{"jsonrpc":"2.0","id":3,"result":{"teststring":"\t solution \t for \t string \t serialization\t","iWa":false,"success":true}

====================================================================================================

For Carriage return(\r):

Input String:curl -H "Content-Type: application/json"  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.aHR0cDovL2xvY2FsaG9zdA.mLd2vW6_3hHMahhbQCaVpgkuB5JM_GuMaxe4hmxgDl8" -X POST   -d '{"jsonrpc":"2.0","id":"3","method": "org.rdk.Network.1.testStringserial", "params":{"teststring":"\r solution \r for \r string \r serialization\r","iWa":false}}' http://127.0.0.1:9998/jsonrpc

Output String:root@arrisxi6:~#
7.0.0.1:9998/jsonrpcerial", "params":{"teststring":"\r solution \r for \r string \r serialization\r","iWa":false}}' http://1
{"jsonrpc":"2.0","id":3,"result":{"teststring":"\r solution \r for \r string \r serialization\r","iWa":false,"success":true}

===========================================================================================================

Combination of Escape Sequence (\t \n \r):

Input String:curl -H "Content-Type: application/json"  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.aHR0cDovL2xvY2FsaG9zdA.mLd2vW6_3hHMahhbQCaVpgkuB5JM_GuMaxe4hmxgDl8" -X POST   -d '{"jsonrpc":"2.0","id":"3","method": "org.rdk.Network.1.testStringserial", "params":{"teststring":"\t solution \n for \t string \t serialization\n\t","iWa":false}}' http://127.0.0.1:9998/jsonrpc (edited) 

Output String:root@arrisxi6:~#
127.0.0.1:9998/jsonrpcial", "params":{"teststring":"\t solution \n for \t string \t serialization\n\t","iWa":false}}' http:/
{"jsonrpc":"2.0","id":3,"result":{"teststring":"\t solution \n for \t string \t serialization\n\t","iWa":false,"success":true}

pwielders (approved):

pwielders (approved):

MFransen69 (approved):

PR #257, opened 2020-09-04
Fix resuming of restarted apps (after crash)

emutavchi:
The ResidentApp (UX) callsign needs to be resumed after restarded by the Monitor

### DIFF
--- a/cmake/project.cmake.in
+++ b/cmake/project.cmake.in
@@ -119,6 +119,10 @@ macro(write_config)
                 map_append(${plugin_config} autostart ${autostart})
             endif()
 
+            if (NOT ${resumed} STREQUAL "")
+                map_append(${plugin_config} resumed ${resumed})
+            endif()
+
             if (NOT ${configuration} STREQUAL "")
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
This is an abuse of the functionality :-) It is not intended to continuously resume a plugin that was configured to start at startup. If you want to have that behavior, use the flag on the plugin called: 
https://github.com/rdkcentral/Thunder/blob/master/Source/plugins/Configuration.h#L53
This flag is intended to change the default start of the plugin to "always" start resumed. This functionality is intended to do it only once at initial startup. So please use the functionality indicated above .

emutavchi:
Hi @pwielders, thanks for review. I updated the pull request to make "resumed" configurable through the project config

pwielders:
I think that is fine but my cmake knowledge is limited, leave it up to bram to reviw.

pwielders (approved):

bramoosterhuis (approved):

PR #258, opened 2020-09-07
[RESCHEDULE] Add a possibility to reschedule a job if it is still pen…

pwielders:
…ding.

### DIFF
--- a/Source/core/WorkerPool.h
+++ b/Source/core/WorkerPool.h
@@ -67,6 +67,15 @@ namespace Core {
                 }
                 return (result);
             }
+            bool Reschedule(const Core::Time& time)
+            {
+                bool result = false;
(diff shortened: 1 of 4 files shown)


Source/core/WorkerPool.h
```diff
@@ -67,6 +67,15 @@ namespace Core {
                 }
                 return (result);
             }
+            bool Reschedule(const Core::Time& time)
+            {
+                bool result = false;
```
MFransen69: Not used

MFransen69 (dismissed):
Small remark but should not stop merging if you are in a hurry

Source/core/WorkerPool.h
pwielders: removed

gladish (approved):

PR #271, opened 2020-09-11
Modified to fetch & build brcm specific code from external repository

ghost:


MFransen69 (dismissed):

MFransen69 (changes requested):
Hi Neeraj, think I was not clear in my last request, sorry. It is not a special flag to override when to take the sources from Gerrit, the idea was that you could specify for a component to not take the sources from Gerrit (or RDK Central) but you could specify a folder to get the sources from (so you can have them local on a disk or something similar instead)

MFransen69 (approved):

woutermeek (approved):

Source/broadcast/CMakeLists.txt
```diff
@@ -55,7 +55,19 @@ target_link_libraries(${TARGET}
...
-     target_sources(${TARGET} PRIVATE Implementation/Nexus/Tuner.cpp)
+    if (BROADCAST_IMPLEMENTATION_PATH)
+        target_sources(${TARGET} PRIVATE ${BROADCAST_IMPLEMENTATION_PATH}/Tuner.cpp)
+    elseif (CODE_FROM_EXTERNAL_REPOSITORY)
+        include(GetExternalCode)
+        GetExternalCode(
+            GIT_REPOSITORY "https://code.rdkcentral.com/r/soc/broadcom/components/rdkcentral/thunder/broadcast"
+            GIT_TAG "master"
+            SOURCE_DIR "Implementation/NexusExt"
+        )
```
gladish, line 66: Shouldn't this reference a stable version of the source instead of just master?

gladish (changes requested):
All the external git repos reference master. I think they should reference a release branch or at least a tag so that the build is table.

gladish (approved):

PR #281, opened 2020-09-17
Minor updates to match fixes that were made on JsonGenerator

vinod-jacob:
Updates to JsonGenerator mandate that property names and its argument names need to different


vinod-jacob:
more changes incoming

PR #284, opened 2020-09-21
Avoid of earlier closelog to prevent related crashes

Gavriliuk:


pwielders (changes requested):
@Gavriliuk can you sign the CLA (some kind of form that you are aware this is open source) Only if you signed it I can merge your change, but also could you just move The closelog() lines till after the Core::Singleton::Dispose(). 
I do think it is important to clean up what you created (or close what you opened) agree, I think we are doing that here to early.

goruklu (dismissed):

Gavriliuk:
@pwielders
I've signed the CLA a few days ago: https://cla-assistant.io/rdkcentral/Thunder shows that I've done with this
But this page doesn't see the signature and asks me to sign the CLA again
Maybe this is because of using a few E-mail addresses in my work:
- gavriliuk@gmail.com - primary address on GitHub
- agavriliuc@productengine.com - company address
- Alexandr_Gavriliuc@comcast.com - Comcast address
I don't know how to solve this

Regarding **closelog()**

We observe a lot of crashes related to logging attempts from any threads after the process is being terminated
I'm not sure that moving the call of closelog() one line below fixes the issue

From the other hand, it seems we don't need to call closelog() because it's use is optional
https://linux.die.net/man/3/closelog

**"Syslog connections are automatically closed on exec or exit."**
https://www.gnu.org/software/libc/manual/html_node/closelog.html

Currently we call **openlog()** the following way:

> openlog(argv[0], LOG_PID, LOG_USER);

It seems we'd better use the following form:

> openlog(nullptr, LOG_PID, LOG_USER);

This means use argv[0] internally and don't reserve a pointer for **ident**

pwielders (changes requested):

MFransen69:
As there was no activity for some time I close the PR. Of course please reopen if it is still desired to respond to the requested changes.

PR #297, opened 2020-09-24
[JsonGen] Fix r/w properties

sebaszm:


HaseenaSainul (approved):
Tested and looks good

PR #302, opened 2020-09-25
DisplayInfo: header update to create jsonrpc file using JsonGenerator

HaseenaSainul:


pwielders (approved):
@HaseenaSainul good that you updated the IGraphicsProperties :+1:

HaseenaSainul:
Thanks @pwielders

PR #303, opened 2020-09-25
[JsonGen] Fix includes when no objects/enums have been generetad

sebaszm:


HaseenaSainul (approved):
Tested and looks good

PR #308, opened 2020-09-30
[Cryptography] Fix for non-nexus implementations

krplata:
After modifying the Cryptography CMakeLists.txt file, some code has been left out, which was responsible for configuring non-nexus implementations. This commit adds those lines back.


Source/cryptography/implementation/CMakeLists.txt
```diff
@@ -39,3 +39,9 @@ if(${CRYPTOGRAPHY_IMPLEMENTATION} STREQUAL "Nexus")
         add_subdirectory(Nexus)
     endif()
 endif()
+
+if(NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/${CRYPTOGRAPHY_IMPLEMENTATION})
+    message(FATAL_ERROR "Invalid value \'${CRYPTOGRAPHY_IMPLEMENTATION}\' for CRYPTOGRAPHY_IMPLEMENTATION.")
+endif()
+
+add_subdirectory(${CRYPTOGRAPHY_IMPLEMENTATION})
```
MFransen69: Shouldn't this be in an else for the if(${CRYPTOGRAPHY_IMPLEMENTATION} STREQUAL "Nexus")?

krplata: Yeah I haven't thought of that. Adjusted the switches with latest revision.

MFransen69 (approved):

pwielders (approved):

PR #311, opened 2020-09-30
[interfaces] Add Amazon Prime initial interface

bramoosterhuis:


pwielders (changes requested):
I think there is quite some overlap with the IStateControl interface. You need to take out the StateControl functionality and move it to the StateControl Functionality and only limit yourself here to the Specifics of Amazon.
The IStateControl (See WPEFramework documentation) has a fundamental meaning in the THunder ECO system and now we have it redundantly in here as well.
Changing state to Exit is probably the redundant method of the Deactivate functionality of the Thunder framework!

bramoosterhuis:
@pwielders As discussed I removed the redundancy

pwielders (approved):

PR #321, opened 2020-10-02
XIONE-1354: revert format C++ names

justin-ware-red:
Reason for change: commit 9eb937e changed the format of C++ names in JsonData_*.h header files. This is breaking existing code that uses these files.
Risks: Medium
Test Procedure: DRMActivationService plugin will build again.

Signed-off-by: Justin Ware <justin.ware@sky.uk>


hteece:
Would also point out that while this is not a bundled service within RDK, the DTV service also suffers from this problem.

sebaszm (commented):
This is breaking JSON-RPC glue generation from C++ headers.

justin-ware-red:
Hi @sebaszm , are you saying that this change will "... break[ing] JSON-RPC glue generation from C++ headers." ? Was the change in the format of the C++ class names in the "JsonData_*.h" intentional and the new format will be the correct one going forwards? For example used to be "GetDrmIdentificationResultData" but is now "GetdrmidentificationResultData" for one of our JSON-RPC interfaces.

sebaszm (changes requested):
Please change self.name to self.true_name. That should make both worlds work correctly.

sebaszm:
Hi Justin, no, that was not intentional, but the change has impact on other code.

justin-ware-red:
> Please change self.name to self.true_name. That should make both worlds work correctly.

Will do.

justin-ware-red:
Hi @sebaszm , updated as you've suggested and can confirm this resolves our broken builds. I don't understand the wideer implications of this change though.

sebaszm (approved):
Good to merge!

justin-ware-red:
Hi @sebaszm , thanks for the approval. I'm not authorized to do the merge will someone else be doing this?

justin-ware-red:
On advice of Vjiay Selvaraj (ComCast) and Thunder component owner this pull request has been cancelled. The new behaviour of all lowercase generated C++ type names is now to be used.

hteece:
I'm arguing this back, as we've currently got a mixed case here.

PR #327, opened 2020-10-07
WebSerializer: Calculate hashes of existing file portion while appending on it

HaseenaSainul:


Source/websocket/WebSerializer.h
```diff
@@ -289,6 +295,28 @@ namespace Web {
             return deserialized;
         }
 
+    private:
+        void LoadHash() {
```
pwielders, line 299: If you look at the FileBody (inheritance tree), you will see that that it inherits from DataElementFile -> DataElement. DataElement will give you access to the file as if it is a "normal" buffer, So I think you can really simplify this. In the Deserialize I suggest:
_hash.Reset();
if (FileBody::Size() > 0) {
    _hash.Input(FileBody::Size(), FileBody::Data());
}
return (FileBody::Deserialize());

and you are done, I hope :-)

HaseenaSainul: Thanks @pwielders for the comments.
FileBody class is inherited from Core::File & IBody classes which does not have any parents class. So not able to use Data()

pwielders: Ohhh good point, I thought we already changed that some time ago to inherit from the DataElementFile, but you are right it is not. 
Than we have two option, move the FileBody to inherit from DataElementFile, maybe a bit too much work for this, or use the File as the accessor object (as it is right now).
Agree prefer the latter (use the File object), but if we are using the file object I suggest to use the FileObject (position and Read) to add it to the _hash again (pseudoCode) since the Serializecalls currently used in the reconstruction of the Has might have overkill (or the file Might not be open at all and otherwise it gets created, through the Deserialis)
if (Core::File::IsOpen() == true) {
    while(dataToRead) {
        uint8_t buffer[FixedSize]
        realReadData = Core::File::Read(buffer, maxSize);
        _hash.Input(realReadData, buffer)
    }
}

HaseenaSainul: Thanks Pierre, will then use to Core::File calls directly

pwielders: :+1:

pwielders (approved):

PR #331, opened 2020-10-08
HelloWorld Example Plugin

sramani-metro:


MFransen69:
As disused with Santhosh we will not merge this PR

PR #333, opened 2020-10-08
Added capability for use to configure additional paths to load plugins.

gladish:


Source/plugins/IShell.h
```diff
@@ -306,6 +306,12 @@ namespace PluginHost {
 
     private:
         void* Root(uint32_t& pid, const uint32_t waitTime, const string className, const uint32_t interface, const uint32_t version = ~0);
+
+        /* @stubgen:omit */
+        virtual std::vector<string> GetAllLibrarySearchPaths(const string& locator) const
```
pwielders: Why not make this pure vritual? That makes it mandatory to implement and an empty set adds no value at all, I think..

pwielders:
Just as a general remark, do we always want to do the default search paths to be used, also if you fill in the new config field, or if you start filling in the config, it is all up to the config to do it, and drop the defaults? I have a slight preference for the latter since it is more in line with how the configs are treated. If there is no config you have a default and otherwise the set valuue is the one it takes.

gladish:
When making it pure-virtual, the compilation fails on 

/home/gladish/Work/Thunder/Thunder/build/Source/proxystubs/generated/proxystubs/ProxyStubs_Shell.cpp:622:11: note:   because the following virtual functions are pure within ‘WPEFramework::ProxyStubs::ShellProxy’:
  622 |     class ShellProxy final : public ProxyStub::UnknownProxyType<IShell> {
      |           ^~~~~~~~~~


That's generated code.  If  I remove the  /* @stubgen:omit */, the the code-generator fails.

ProxyStubGenerator: ERROR: /home/gladish/Work/Thunder/Thunder/build/Source/proxystubs/include/IShell.h(310): 'std::vector<string>': undefined type
CMake Error at /usr/share/cmake-3.16/Modules/FindProxyStubGenerator.cmake:95 (message):
  ProxyStubGenerator generator failed.

gladish:
For your other comment, do you mean something like this?

```
sandsmark# git diff
diff --git a/Source/WPEFramework/PluginServer.h b/Source/WPEFramework/PluginServer.h
index feb426d1..830b6f17 100644
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -857,16 +857,22 @@ namespace PluginHost {
         private:
             virtual std::vector<string> GetAllLibrarySearchPaths(const string& locator) const override
             {
-                // system configured paths
                 std::vector<string> all_paths;
-                all_paths.push_back(_administrator.Configuration().PersistentPath() + locator);
-                all_paths.push_back(_administrator.Configuration().SystemPath() + locator);
-                all_paths.push_back(_administrator.Configuration().AppPath() + _T("Plugins/") + locator);

-                // additionaly defined user paths
                 const std::vector<string> temp = _administrator.Configuration().LinkerPluginPaths();
-                for (const string& s : temp)
-                    all_paths.push_back(Core::Directory::Normalize(s) + locator);
+                if (!temp.empty())
+                {
+                    // additionaly defined user paths
+                    for (const string& s : temp)
+                        all_paths.push_back(Core::Directory::Normalize(s) + locator);
+                }
+                else
+                {
+                    // system configured paths
+                    all_paths.push_back(_administrator.Configuration().PersistentPath() + locator);
+                    all_paths.push_back(_administrator.Configuration().SystemPath() + locator);
+                    all_paths.push_back(_administrator.Configuration().AppPath() + _T("Plugins/") + locator);
+                }

                 return all_paths;
             }
```

pwielders:
> When making it pure-virtual, the compilation fails on
> 
> /home/gladish/Work/Thunder/Thunder/build/Source/proxystubs/generated/proxystubs/ProxyStubs_Shell.cpp:622:11: note: because the following virtual functions are pure within ‘WPEFramework::ProxyStubs::ShellProxy’:
> 622 | class ShellProxy final : public ProxyStub::UnknownProxyType {
> | ^~~~~~~~~~
> 
> That's generated code. If I remove the /* @stubgen:omit */, the the code-generator fails.
> 
> ProxyStubGenerator: ERROR: /home/gladish/Work/Thunder/Thunder/build/Source/proxystubs/include/IShell.h(310): 'std::vector': undefined type
> CMake Error at /usr/share/cmake-3.16/Modules/FindProxyStubGenerator.cmake:95 (message):
> ProxyStubGenerator generator failed.

I guess that is a code generator bug than, let me pick it up with Sebas.

pwielders (approved):

PR #339, opened 2020-10-15
Process Group settings: use setgid to update group and call the group settings before doing the user

HaseenaSainul:


pwielders (changes requested):
The first part is good in the Process.cpp, the second part breaks the interface semantics of the class. It is expected to operate onn the given process but now you move it to the current process..
I think the reason that you might have seen it fail is because the user whi is executing it is not allowed to change the process group of another process, which is something that should be arranged by changeing the linux user rights in the OS.

HaseenaSainul:
Thanks @pwielders for the inputs. I would like to discuss this offline. I will reach you on this.

pwielders (approved):

PR #341, opened 2020-10-16
interfaces: Add ICustomerCareOperations

bramoosterhuis:


pwielders (changes requested):
Sofar the new COM interface can perfectlt be matched to the JSON interface without specific (custom)_ specification, so suggest to combine the JSON specification (and document generation) into the ICustomerCareOperations.
Referring here to the presentation Sebastian did during the last global team-meeting.

bramoosterhuis:
Removed the useless json file.

pwielders (approved):

PR #355, opened 2020-10-26
Development/jsonrpc on windows

pwielders:


Source/websocket/JSONRPCLink.h
```diff
@@ -594,7 +594,7 @@ namespace JSONRPC {
                 _localSpace = localCallsign;
             }
 
-            uint8_t version = Core::JSONRPC::Message::Version(callsign + '.');
+            uint8_t version = Core::JSONRPC::Message::Version(_callsign);
```
MFransen69, line 597: Could be that I'm mistaken, too late now to start debugging ;), but I think the +'.' was essential.
Have you tested if this change still works correctly with the version test cases in the json rpc example application? If I remember correctly it was essential to add the '.' for now and we files an jira ticket to fix the version parsing code so it can also handle the case if the '.' when it is not added (it is also in a few lines up from where you removed this). If I remember correctly the designator here is in the form "callsign.version" and you can test here what happens with the current parsing code: cpp.sh/4xq6t 
Of course if I'm mistaken just ignore this comment :)

MFransen69 (changes requested):

Source/websocket/JSONRPCLink.h
pwielders: It is essential, but if youlook at the fullcode you will see that the _callsign already adds the dot but only in case the given callsign is not an empty string. During the test under windows I observed a crash in case the callsign is empty, indicating that it should use any version of the Controller plugin, that it would now try to evaluate the dot only, whihc crashes. Since I did not want to repeat what is hapening a few lines higher, I opted for just using the value set there.
Testcase now succeeds and the concattenation is only done once.
So your concern is right, the dot should stay, but it does stay in case the callsign is not empty.

MFransen69 (approved):

PR #359, opened 2020-10-30
DELIA-46560 : incorrect thunder response serialization

npoltorapavlo:
Reason for change: for consistency take all bits
from the copy on either copy constructor or assignment


pwielders:
@npoltorapavlo Do you have a testcase that shows a failure that we can add to the unit tests to see if this fixes an issue.

I do agree that it is more consistent if we would align it but as this is core, we do need to be very carefull..

npoltorapavlo:
@pwielders, I've added 2 testcases in a comment to DELIA-46560.

pwielders (approved):

PR #362, opened 2020-11-04
[DOORBELL] Prevent an early binding of the socket.

pwielders:
If an UDP socket is used to signal the ringing of the doorbell, the single socket (used before) will automatically bind to a address, not the node id configured, if the Sending is done before the receiving. This is applicable if in the Thudner application (WPEFramework) a TRACE is issued prior to the TraceCOntrol being started. 

This fix splits the Sending and the Receiving so this can no longer be an issue at the cost of an additional socket.

### DIFF
--- a/Source/core/DoorBell.cpp
+++ b/Source/core/DoorBell.cpp
@@ -22,6 +22,9 @@
 #ifdef __WINDOWS__
 #include <Winsock2.h>
 #include <ws2tcpip.h>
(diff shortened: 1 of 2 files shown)


MFransen69 (approved):

MFransen69 (changes requested):
As discussed

pwielders:
Tested on Linux as well with succes.

MFransen69 (approved):

gladish (approved):

PR #379, opened 2020-12-02
Allow for json string parsing > unit16_max

vijs:


Source/core/JSON.h
```diff
@@ -121,19 +122,29 @@ namespace Core {
...
+            }
+
+            template <typename INSTANCEOBJECT>
+            static bool FromString(const char* text, string::size_type length, INSTANCEOBJECT& realObject, Core::OptionalType<Error>& error)
             {
                 uint32_t offset  = 0;
                 uint32_t handled = 0;
-                uint32_t size    = text.length();
+
+                if (length > std::numeric_limits<size_type>::max())
```
pwielders: I think we need a discussion on how JSON parsing was intended. We should be able to parse JSON with this parser, even if it is coming in 1 byte at a time. This is a parser that parses while we go and only pickup what is required. That is why Thunder is not using an OpenSource parser as this feature is not in any of the open source parsers out there.
For example TCP connections might get data in chunks, we need the be capable in handling those separate chunks as well. Do like that by changing the type we can now increase the block size we can handle at once, but we should not not limit it to 32 bits, as the uint16_t parser could go limitless!

PR #387, opened 2020-12-11
[METROL-115] Cleanup JSON RPC DetermineVersion Code

Vinodsathyaseelan:
Reason for change: Improvement
 - Currently the JSON RPC DetermineVersion() code
   cannot handle if the designator string does not
   end with a '.' (dot)

Solution :
 - Update Core::JSONRPC::Message::Version() code to handle
   both cases, ending with or without '.'


MFransen69 (changes requested):
Hi Vinod,

Indeed rather complex what we want to achieve here. I looked at the changes and certainly there is no flaw, it achieves what was the goal. I also realised there are still some optimizations to be made. On one hand you can use the info that if a @ is present (you need to search for it anyway so it is no added overhead) the version can be only in one place and in some of the uses of the version method it is known that a full designator (meaning the method part must be present so we also know where the version is, there is no ambiguity). For example where it is called in the JSON Message that must be true and we can use that information to optimize (so I added a parameter so the caller can indicate that). A small thing, and subjective I guess, I think that just always parsing backwards makes the code easier to understand than first parsing one way and then the other way. So I made a counter proposal for the implementation taking the optimizations and parsing into account, hope you don't mind. This is the link where you can see all three implementations (old, fwd parsing and new) and also test them: https://www.onlinegdb.com/ryBE4s82D 
What is you opinion? 
I also did some benchmarking and the new one is significantly faster, see the benchmark screenshot I attached.
I think (did not verify) that the biggest performance change actually is not using atoi, but indeed that was already used in the old implementation so I can imagine you saw no need to replace that. Still as this Version method is called on every JSON RPC message that is executed by Thunder, speed is a factor here and besides the atoi as explained above it has two optimizations in other areas as well. I would propose to use the new version. What are your thoughts?

Screenshot from 2020-12-15 21-56-30

Vinodsathyaseelan (commented):
Marcel, 
I have incorporated your suggestion and merged the optimized code changes. Fixed an edge case issue and tested. Please review the commits.

MFransen69 (approved):

pwielders (approved):

PR #391, opened 2020-12-12
[WPEFRAMEWORK] Make sure HTTP requests can be made a-synchronous

pwielders:
The WebBridge plugin delegates incoming requests to the Javascript plugin running in a JavaScriptInterpreter. This can potentially take some time so the system decouples the request from the response to the Javascript plugin. In case an HTTP request with a JSONRPC message is received, it was expecting a response on the invoke. Due to the decoupled WebBridge feature a JSONRPC request now nolonger return a response by default.

This patch assumes that if there is no JSONRPC message response on the invoke that it is an a-synchronous call and response will follow later. As  part of the initial request, is the request to "close" the TCP connection on completeion of this request, this information is forwarded and stored now as part of the channelInformation. If finally the JSONRPC response is received (and wrapped into an always HTTP response 200 OK message), the channel is closed (if requested) after the JSONRPC response has been send.

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -539,6 +539,7 @@ ENUM_CONVERSION_BEGIN(Core::ProcessInfo::scheduler)
         , _parent(static_cast<ChannelMap&>(*parent).Parent())
         , _security(_parent.Officer())
         , _service()
+        , _requestClose(false)
     {
         TRACE(Activity, (_T("Construct a link with ID: [%d] to [%s]"), Id(), remoteId.QualifiedName().c_str()));
     }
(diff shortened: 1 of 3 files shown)


MFransen69 (approved):

gladish (approved):
Verified working with WebBridge plugin

[COSWL-C051540]: curl -H "Content-Type: application/json"   -X POST   -d '{"jsonrpc":"2.0","id":1,"method":"Controller.1.status@org.rdk.Calculator"}' http://127.0.0.1:10001/jsonrpc
{"jsonrpc":"2.0","id":1,"result":[{"callsign":"org.rdk.Calculator","locator":"libWPEFrameworkWebBridge.so","classname":"WebBridge","autostart":true,"state":"activated","observers":0,"module":"Plugin_WebBridge","hash":"engineering_build_for_debug_purpose_only"}]}

[COSWL-C051540]: curl -H "Content-Type: application/json"   -X POST   -d '{"jsonrpc":"2.0","id":1,"method":"org.rdk.Calculator.1.add", "par
ams":[1,2]}' http://127.0.0.1:10001/jsonrpc
{"jsonrpc":"2.0","id":1,"result":3}

PR #404, opened 2020-12-21
[TIME] Add JulianDate calculation and a possibility to get a ToLocal …

pwielders:
…time.

### DIFF
--- a/Source/core/Time.cpp
+++ b/Source/core/Time.cpp
@@ -549,13 +549,32 @@ namespace Core {
         : _isLocalTime(localTime)
     {
         _time.wYear = year;
-        _time.wMonth = month % 12;
-        _time.wDay = day % 31;
+        _time.wMonth = month;
+        _time.wDay = day;
         _time.wHour = hour % 24;
         _time.wMinute = minute % 60;
         _time.wSecond = second % 60;
         _time.wMilliseconds = millisecond;
         _time.wDayOfWeek = static_cast<WORD>(~0);
+    }
+
+    /**
+     * Get day count since Monday, January 1, 4713 BC
+     * https://en.wikipedia.org/wiki/Julian_day
+     */
+    float Time::JulianDate() const {
+        uint16_t year = _time.wYear;
+        WORD month = _time.wMonth;
+        WORD day = _time.wDay;
+        WORD hour = _time.wHour;
+        WORD minutes = _time.wMinute;
+        WORD seconds = _time.wSecond;
+
+        // julian day number algorithm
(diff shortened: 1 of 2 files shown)


Source/core/Time.cpp
```diff
@@ -779,6 +814,21 @@ namespace Core {
         _ticks = (static_cast<uint64_t>(time.tv_sec) * MicroSecondsPerSecond) + (time.tv_nsec / NanoSecondsPerMicroSecond) + OffsetTicksForEpoch;
     }
 
+    Time Time::ToLocal() const {
+        struct tm local = _time;
+        time_t flatTime;
+        flatTime = mktime(&local);
+        localtime_r(&flatTime, &local);
```
HaseenaSainul, line 817: Why we are doing broken-down time representation to local time, then again reverse ?

Source/core/Time.cpp
```diff
@@ -856,6 +906,25 @@ namespace Core {
...
+        uint8_t day = _time.tm_mday;
+        uint8_t hour = _time.tm_hour;
+        uint8_t minutes = _time.tm_min;
+        uint8_t seconds = _time.tm_sec;
+
+        // julian day number algorithm
+        uint32_t JDN = (1461 * (year + 4800 + (month - 14) / 12)) / 4 + (367 * (month - 2 - 12 * ((month - 14) / 12))) / 12 - (3 * ((year + 4900 + (month - 14) / 12) / 100)) / 4 + day - 32075;
+
+        // JDN to JD algorithm
+        return (JDN + static_cast<float>(((hour - 12) + (minutes * 60) + (seconds * 60 * 60)) / (24.0 * 60.0 * 60.0)));
```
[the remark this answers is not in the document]
pwielders: Why do you think so ? I agree, I did reformat it a bit to mov tot he factionals only at the last stage, but I think the return both the same value.

Source/core/Time.cpp
pwielders: To allow a now time (alwyas in UTC) t be returned as a local time (using the locale).
Internally we calculate with UTC times, So any timer that is being set will be done in UTC, If someone now would like to know what that time would be in the system set local time zone, it can be done by the calling sequence:
Core::Time::Now().ToLocal()
The returned time for us (in CET) will than be 1 hour earlier.

HaseenaSainul: Ok, thanks

Source/core/Time.cpp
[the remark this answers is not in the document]
HaseenaSainul: I checked it using test app and got different value. Will share the code offline

HaseenaSainul (approved):

PR #406, opened 2020-12-21
[LAT/LONG] Add the Latitue and the Longitude to the Subsystem Location.

pwielders:
In case there is no location determination (for language or localization) allow the Latitude and Longitude to
be configured and the TimeZone/Language could be set from these configuration parameters.


HaseenaSainul (approved):
Just verified using compile.

PR #408, opened 2020-12-24
DELIA-47742 : multiple subscribers

npoltorapavlo:
Reason for change: if LinkType Subscribe is called more than once
for the same event name, nothing is subscribed.
Test Procedure: testcases provided in ticket
Risks: Low

### DIFF
--- a/Source/websocket/JSONRPCLink.h
+++ b/Source/websocket/JSONRPCLink.h
@@ -676,6 +676,9 @@ namespace JSONRPC {
 
             if ((result != Core::ERROR_NONE) || (response.IsValid() == false) || (response->Error.IsSet() == true)) {
                 _handler.Unregister(eventName);
+                if ((result == Core::ERROR_NONE) && (response->Error.IsSet() == true)) {
+                    result = response->Error.Code.Value();
+                }
             }
 
             return (result);
@@ -691,6 +694,9 @@ namespace JSONRPC {
 
             if ((result != Core::ERROR_NONE) || (response.IsValid() == false) || (response->Error.IsSet() == true)) {
                 _handler.Unregister(eventName);
+                if ((result == Core::ERROR_NONE) && (response->Error.IsSet() == true)) {
+                    result = response->Error.Code.Value();
+                }
             }
 
             return (result);


npoltorapavlo:
@pwielders , please review or suggest who can review this PR

pwielders:
Why should the change of the cardinality (1 event, 1 callback -> 1 event, n callbacks) change at this layer? Why not solve this in the callback, if you want to realize this functionality you could also do this from the invoked method.
So my first question would be where is this change request coming from, and why pushing this functionality to lower layers?

Did review the change. The interface is now redundant, which should be resolved. The replace parameter should be dropped as the replaced=true functionality can be realized by unregistering the event before registering it again, which is the way to use symetrical interfaces :-) . Remainder of the commit is fine.

npoltorapavlo:
@pwielders, thanks for the comments, I've updated the PR. The change is required for DELIA-47742 to support multiple subscriptions for the same event.

pwielders:
I think you picked up the remark;
Why not solve this in the callback, if you want to realize this functionality you could also do this from the invoked method.
very nicely, and since you picked it up like this, I think the last step would be to move this new code outside of Thunder (last part of my remark, why pushing this functionality to lower layers). I think there is no need to push it here to keep the design (and cardinality) simple, 1-1. Complexity happens on the next layer (where you need this functionality).
Right now the interface behavior is a-symetrical. 
You can subscribe an event and a Callback, but you can not unsubscribe that handler. If you would unregister, "all" handlers will be unsubscribed. 
So as long as there is 1 handler subscribed it is symmetrical (you can unsubscribe) but if a second handler is subscribed, that get "automagically" unsubscribed, which might lead to unexpected behavior in code. 
I see two options:
1) continue moving this functionality to a higher layer (your implementation layer) and there is no change required here (KIS).
2) Create a symmetrical interface by adding the method to unsubscribe with the Unsubscribe(https://github.com/rdkcentral/Thunder/blob/19e6b11f92bd5306a1aac6c50016ebb62789682d/Source/websocket/JSONRPCLink.h#L755) 

As option 2 is causing probably a waterfall of issues since this interface is already deployed and thus not recommended and is causing a performance/resource penalty for all deployments out there already, for which the get no benefits (these deployments work with a cardinality of 1-1), I would like to go for option 1 :-) and most of the work you already did.

For option 1, I think it would be a good option to move the class you already created (EventHandler, no need to inherit) outside the LinkType template and offer it as a glue class to realize this next layer, so the next one who also want to subscribe multiple handlers to 1 event can re-use the work you do here.
 
I also do think it is good to return an ERROR in case the event is already subscribed to help users of this method to know that their subscribe did not succeed. Right now you return ERROR_NONE, which I think is misleading as the callee might assume it succeeded but it did not ...
In Thunder we have a coding rule that there is only 1 return allowed in the method this for Maintainability aspects so could you please move the return to the end of the method.

pwielders (approved):

PR #410, opened 2020-12-30
[THUNDER] Closing down in debug leads to an ASSERT.

pwielders:
In the Singleton list, the Traces are not being closed. Close the Trace resources before calling the Singleton::Dispose.


ghost (approved):
LGTM.
Observed that TraceUnit::Close() is called even without this change when WPEFramework process exits.
One more observed issue is: 
If any TraceControl JSON RPC service is used, plugin processes are not exiting when  WPEFramework is stopped OR when any plugin is deactivated.

PR #423, opened 2021-01-07
Development/metrol 173

krplata:
The ResourceMonitor plugin, while executing functions `Collect(Single/Multiple/WPEProcess)` marks pages for processes unrelated to the `WPE` stack. If one of these processes exits between the following function calls:
1. After https://github.com/rdkcentral/ThunderNanoServices/blob/master/ResourceMonitor/ResourceMonitorImplementation.cpp#L182 
2. Before https://github.com/rdkcentral/Thunder/blob/master/Source/core/ProcessInfo.cpp#L617
the pagemap file corresponding to it's `PID`, is no longer available.

In such a case after executing - `FILE * pagemapFile = fopen(pagemapPath, "rb");`, the pagemapFile contains a value of `nullptr`.


svdmaar (approved):
Agree, might indeed be the process no longer exists in the meantime. The rest of the PR seems to be reformatting. Maybe next time best to split this over two PRs.

PR #431, opened 2021-01-13
Calls Suspend after Send

HaseenaSainul:


HaseenaSainul:
Closing the changes are already taken care as part of Close() itself

PR #432, opened 2021-01-13
Added regex workaround so as to allow OCDM shared lib unloading from memory after deactivation.

ghost:


pwielders (approved):
@neeraj-deshpande can you also create a Jira ticket to further investigate this pll request so we can emove this hack at a later stage..

pwielders:
@neeraj-deshpande as this also resolves the unloading of the SecurityAgent (also uses regex), an we move this to the R2.1.It brings no risk on regression (last famous words :-) )

ghost:
@pwielders I have created JIRA issue METROL-225.

PR #450, opened 2021-01-27
[WPEFramework] Fix access permission for /tmp/keyhandler

404ag:
Some of the process (like Amazon plugin) can run with a different
user and group than the Framework main process. In this
case, the processes that are using /tmp/keyhandler can't access
that file, thus loosing possibility to obtain key presess.
This commit fix this, with adding R/W file permission for
the 'other' group.

### DIFF
--- a/Source/WPEFramework/Config.h
+++ b/Source/WPEFramework/Config.h
@@ -219,7 +219,7 @@ namespace PluginHost {
                     : Locator("127.0.0.1:9631")
                     , Type(InputHandler::VIRTUAL)
 #else
-                    : Locator("/tmp/keyhandler|0760")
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
Could we test if the Read access is sufficient (0764) ? lets avoid other users from writing back keys into this pipeline.

404ag:
Seems like file permissions 0764 are not sufficient. At least not in the tested case (Amazon in Webkit)

> Could we test if the Read access is sufficient (0764) ? lets avoid other users from writing back keys into this pipeline.

pwielders (approved):

PR #469, opened 2021-02-15
Modified to make Controller plugin's delete method work.

ghost:


pwielders (approved):
This is, I guess, already something available on the R2 release that you want to back port to an older release. If so fine for me!

Source/core/FileSystem.cpp
```diff
@@ -193,7 +193,11 @@ namespace Core {
             if (result[length - 1] != '/')
 #endif
             {
-                result += '/';
+                Core::File file(location);
+
+                if (file.IsDirectory() == true) {
+                    result += '/';
```
HaseenaSainul: This change is not required, since Pierre removed this with the commit : https://github.com/rdkcentral/Thunder/commit/172be48714ea385923218c0bd0f2add0e7a12a8e#diff-0525db6180b9100c5f428338d271c6abc1a9b13f145622a1aa162e3787d3e6ed. Could you please cross check the delete without this change ?

ghost: Done.

HaseenaSainul (approved):
LGTM

PR #472, opened 2021-02-15
[CORE] Added logic to fix the deadlock situation during the lock()- unlock() sequence.

anjalirajan:
Added the logic to fix the deadlock situation in cyclicbuffer. The testcases to test the same is committed as a separate PR.


MFransen69:
Will be handled in a future redesign

PR #482, opened 2021-02-23
[ResourceMonitor] Change way of measuring memory usage on Linux, refactoring

lukiwa:
The older version used reading from /proc/PID/pagemap and retrieving PFNs from here, to count occupied pages. The results were suspicious so I checked them - in my opinion, one step (described here: https://www.kernel.org/doc/Documentation/vm/pagemap.txt ) was missing. After fixing the older method it appeared that the same results can be obtained when parsing /proc/PID/smaps.

Also reading this file enabled me to get also PSS + accurate Shared memory and Resident (previously was read from /proc/PID/statm where, as documentation stated, the result are inaccurate).

Removed Jiffies function as it does not bring useful information, at least in the context of CPU usage measurements per process.


Source/core/ProcessInfo.cpp
```diff
@@ -32,6 +32,7 @@
 #include <unistd.h>
 #endif
 
+#include "tracing/TraceCategories.h"
```
pwielders: This is not possible. This give cyclyc dependencies. You can not use tracing in core (only TRACE__LX allowed here)

pwielders (approved):

PR #484, opened 2021-02-23
Make the resume flag also report if it was *not* set

pwielders:


Damian-Danylko (approved):
Looks good, tested, works as expected

PR #485, opened 2021-02-25
[SOCKETPORT] Extract from UDP/RAW messages on which interface they wh…

pwielders:
…ere received.

To make the locator specific to the IP address on the interface it was received, extract the interface id, with the remote address, on which it was received. This way the Locator can be made to use the IPv4/IPv6 address for that interface to reach it back on.

### DIFF
--- a/Source/core/SocketPort.h
+++ b/Source/core/SocketPort.h
@@ -173,6 +173,9 @@ namespace Core {
         {
             return (m_ReceivedNode);
         }
+        inline uint32_t ReceivedInterface() const {
+            return (m_Interface);
+        }
         inline uint16_t SendBufferSize() const
         {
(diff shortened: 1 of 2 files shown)


MFransen69 (dismissed):

Source/core/SocketPort.h
```diff
@@ -265,6 +268,7 @@ namespace Core {
         uint16_t m_ReadBytes;
         uint16_t m_SendBytes;
         uint16_t m_SendOffset;
+        uint32_t m_Interface;
```
sebaszm, line 271: m_Interface is missing initialization (0?)

sebaszm (changes requested):

sebaszm (approved):
Works great. Good to merge

PR #495, opened 2021-03-10
WPEFramework: Trigger cleanup connections after close

HaseenaSainul:


Source/WPEFramework/Config.h
```diff
@@ -532,6 +532,7 @@ namespace PluginHost {
                 _communicator = Core::NodeId(config.Communicator.Value().c_str());
                 _redirect = config.Redirect.Value();
                 _version = config.Version.Value();
+                _idleTime = config.IdleTime.IsSet() ? config.IdleTime.Value() : 0;
```
pwielders: The default of the IDleTime is already 0, so just get the value of the IDleTime, no need to check if it is set. See the constructor.

HaseenaSainul: Done

pwielders (approved):

PR #510, opened 2021-03-17
Set all file descriptors as close on exit to prevent them from being …

gladish:
…inherited by WPEProcess

### DIFF
--- a/Source/core/DoorBell.cpp
+++ b/Source/core/DoorBell.cpp
@@ -34,7 +34,11 @@ namespace Core {
     DoorBell::Connector::Connector(DoorBell& parent, const Core::NodeId& node)
         : _parent(parent)
         , _doorbell(node)
+        #ifdef __LINUX__
+        , _sendSocket(::socket(_doorbell.Type(), SOCK_DGRAM|SOCK_CLOEXEC, 0))
+        #else
         , _sendSocket(::socket(_doorbell.Type(), SOCK_DGRAM, 0))
+        #endif
         , _receiveSocket(INVALID_SOCKET)
         , _registered(0)
     {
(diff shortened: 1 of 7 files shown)


Source/core/DoorBell.cpp
```diff
@@ -34,7 +34,11 @@ namespace Core {
     DoorBell::Connector::Connector(DoorBell& parent, const Core::NodeId& node)
         : _parent(parent)
         , _doorbell(node)
+        #ifdef __LINUX__
```
pwielders: The SOCK_CLOEXEC is also available on windows so no need to use the preprocessor to diffrentiate.

Source/core/SocketPort.cpp
```diff
@@ -644,7 +644,12 @@ namespace Core {
         }
 #endif
 
-        if ((l_Result = ::socket(localNode.Type(), SocketMode(), localNode.Extension())) == INVALID_SOCKET) {
+        #ifdef __LINUX__
```
pwielders: Windows does support SOCK_CLOEXEC as well so lets not diffrentiate her

Source/core/SocketPort.cpp
```diff
@@ -1132,7 +1138,11 @@ namespace Core {
         socklen_t size = sizeof(address);
         SOCKET result;
 
+        #ifdef __LINUX__
+        if ((result = ::accept4(m_Socket, (struct sockaddr*)&address, &size, SOCK_CLOEXEC)) != SOCKET_ERROR) {
+        #else
```
pwielders: Could not find the accept4 in the windows documentation that quickly, but suggest we remove the preprocessor stuff anyway. I will test it on windows and if I run into an issue I will just create a local accept4 method in this compilation unit that will push the SOCK_CLOEXEC option onto the socket. I am afraid if I am not running into a compilation issue, we forget about this change.

pwielders (changes requested):
Looks good, just would like to drop the __LINUX_ guards you added. If it hurst in windows (and I think only the accept4 will) we will fix it so that also on windows, connecting sockets get the SOCK_CLOEXEC flag.

pwielders (approved):
Looks good to me. Waiting for @HaseenaSainul to make sure that we do not introduce any regression if she reports no regression good to merge!

HaseenaSainul (approved):
LGTM

PR #511, opened 2021-03-17
Add null-check in RealAllocate

Gavriliuk:


pwielders (approved):
Love defensive coding! Good fix!

PR #518, opened 2021-03-23
RDK-29971: Thunder Messenger Restrictions

npoltorapavlo:
Reason for change: Allow non default
constructors of JSONRPC.
Test Procedure: It builds
Risks: Low

### DIFF
--- a/Source/plugins/JSONRPC.h
+++ b/Source/plugins/JSONRPC.h
@@ -412,6 +412,9 @@ namespace PluginHost {
         JSONRPCSupportsEventStatus& operator=(const JSONRPCSupportsEventStatus&) = delete;
 
         JSONRPCSupportsEventStatus() = default;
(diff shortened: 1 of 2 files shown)


Source/plugins/JSONRPC.h
```diff
@@ -412,6 +412,9 @@ namespace PluginHost {
         JSONRPCSupportsEventStatus& operator=(const JSONRPCSupportsEventStatus&) = delete;
 
         JSONRPCSupportsEventStatus() = default;
+        JSONRPCSupportsEventStatus(const TokenCheckFunction& validation) : JSONRPC(validation) {}
+        JSONRPCSupportsEventStatus(const std::vector<uint8_t> versions) : JSONRPC(versions) {}
```
pwielders: const vector<> and not a reference. A bit expensive on copying I guess :-) I assume a mistake the missing reference ? also the next line...

pwielders (approved):

PR #534, opened 2021-03-29
Metrol261

wiswis15:
Not sure how to test for any regression issues. I have tested Thunder on the pi, but very high level.


wiswis15:
ill discuss with Marcel first.

PR #537, opened 2021-03-29
[LEAKS] Register is sometime called more often (if Events -> 0)

pwielders:
In unhappy day scenarios, the SocketState might be 0, but the Socket still exists.


HaseenaSainul (approved):
LGTM

PR #541, opened 2021-03-30
NetworkInfo: AdapterObserver: Keep separate interfaces for Link update and address update

HaseenaSainul:
Note: Not able to test, since my current build is in epyc. Will update the status once get tested

### DIFF
--- a/Source/core/NetworkInfo.h
+++ b/Source/core/NetworkInfo.h
@@ -136,7 +136,8 @@ namespace Core {
             virtual ~INotification() = default;
 
             virtual void Event(const string&) = 0;
-            virtual void AddressEvent(const string&) {}
+            virtual void Added(const string&, const Core::IPNode&) {}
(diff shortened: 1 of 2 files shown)


Source/core/NetworkInfo.h
```diff
@@ -136,6 +136,7 @@ namespace Core {
             virtual ~INotification() = default;
 
             virtual void Event(const string&) = 0;
+            virtual void AddressEvent(const string&) {}
```
pwielders: If we are going to specialize the interface lets do it explicitly. Why not :
virtual void Added(const string& interface, const Core::IPNode& address) = 0;
virtual void Removed(const string& interface, const Core::IPNode& address) = 0;
That prevents the receiver from finding out what address change happenend and you do have that information from the calling position.

pwielders (approved):
LGTM :-)

sebaszm (approved):

HaseenaSainul:
Merging the changes since it is validated

PR #550, opened 2021-04-05
[PluginServer] Extend Override Save/Load

wiswis15:
Tested locally  in the pi.


wiswis15:
Did not use the correct branch.

PR #551, opened 2021-04-06
JSON updates for escape characters

HaseenaSainul:


Source/core/JSON.h
```diff
@@ -1691,8 +1691,15 @@ namespace Core {
                             
                             if (result < maxLength) {
 
+                                char convertedValue = IsEscapeSequenceValue(*source);
+                                if (convertedValue != 0) {
+                                    stream[result++] = '\\';
```
pwielders: This might fail if you run the Small parser test (only 1 character at a time in the buffer). See this testcase:
https://github.com/rdkcentral/Thunder/blob/master/Tests/unit/core/test_json.cpp#L137
this is testing with a buffer size of1. COuld we test this feature with this tester if it still works ?

HaseenaSainul: yes, this test passed. I ran all unit test cases with this changes and all are passed

Source/core/JSON.h
```diff
@@ -1938,17 +1949,66 @@ namespace Core {
...
+                 default:
+                     break;
+                 }
+                 return value;
+            }
+
             EscapeSequenceAction GetEscapeSequenceAction(char current) const
             {
                 EscapeSequenceAction action = EscapeSequenceAction::COLLAPSE;
                 if (current == 'u' || current == 'n' || current == 't' || current == 'r' || current == 'f' || current == 'b') {
```
pwielders: Can we all move this to 1 array, cause now the data is a bit distributed..
So create a LUT like: { { 'b', 0x08 }, { 'f', 0x0c }, ...} and than use it in the GetEscapeSequence/IsEscapeSequence but also here, that you check if the character IsEscapeCharacter, since I think there is now a mismatch, all conversions have b,f,n,r,t but no u and here is a check for the u. 
I think the LUT coudl even contain { {'\', '\'} ... maybe. it will make the code more transparent than.

HaseenaSainul: Ok, will update

Source/core/JSON.h
HaseenaSainul: all test cases ran with the changes and passed

Source/core/JSON.h
HaseenaSainul: Added Lookup Map to handle this

pwielders (dismissed):

pwielders (approved):

PR #553, opened 2021-04-06
Development/server override

wiswis15:
Successfully tested locally and on the pi.


MFransen69:
Needs some rework and Wissem is not available for some time

PR #555, opened 2021-04-08
Rdkb 27781

mrcomcast123:
Fixed gcc 4.8.4 compile errors encountered porting to xb7: RDKB-27781.  Apparently xb7 is being upgraded to Dunfell in 2103_sprint, so we could optionally wait for this and see if the newer gcc works without patch.  If we want to build thunder using older gcc versions, this pull is needed. @markrollins


pwielders (changes requested):
This is probably because the branch was not rebased before this got pushed in, but I think that the only essential part of this PR is: 1 commit. Will list it in a moment. That commit looks fine and happy to include. But I guess I need to cherry pick it or you need to create a new pull request on the master with only this change. What is your preference ?
https://github.com/rdkcentral/Thunder/pull/555/commits/f3b513d9d1dfd7c7a7c0769ac491c4b85cbfe0d5

mrcomcast123:
Yea, I tried to base off R2 branch.  I will close this and redo this properly.

PR #565, opened 2021-04-20
[METRO-247] PluginHost::IWebSocket Inbound(const string& identifier) …

Vinodsathyaseelan:
…identifier always empty

- Reason for change :  Bug Fix
      The Inbound(const string& identifier) on the
      PluginHost::IWebSocket interface will always have an empty identifier

- Solution :
      It should be filled with the identifier string when it receives a package
      like ' "identifier" : { "other" : "content" }'


Source/plugins/Channel.h
```diff
@@ -153,13 +155,46 @@ namespace PluginHost {
...
+                    while (current < length) {
+                        if (stream[current] == ':') {
+                            bValid = true;
+                            break;
+                        }
+                        current++;
+                    }
+                }
+
+                auto result = (bValid != false)
```
MFransen69, line 174: Hi Vinod, I think we did discuss but perhaps if I'm mistaken: I guess it would be better do have this parsed with the core JSON parser (we have the parsing code in one place and it can for example also parse when characters are coming in in chunks, or even one by one)

MFransen69 (changes requested):

Source/plugins/Channel.h
Vinodsathyaseelan: The input is not a valid JSON formatted string. So cannot use JSON parser here. Had a discussion earlier with Pierre on this. There is a plan to re-design this functionality.  This PR is no longer valid and we can close this.

PR #588, opened 2021-05-18
[CALLSTACK] ALlow callstack retrievel through a RESTFul call (Next de…

pwielders:
…livery will be the JSONRPC stack :-))


Source/WPEFramework/Controller.cpp
```diff
@@ -254,6 +270,16 @@ namespace Plugin {
             WorkerPoolMetaData(response->Process);
 
             result->Body(Core::proxy_cast<Web::IBody>(response));
+        } else if (index.Current() == _T ("Callstack")) {
+            if (index.Next() == false) {
+                result->ErrorCode = Web::STATUS_BAD_REQUEST;
+                result->Message = _T("Probe functionality not enabled!");
```
MFransen69: Think you might want to change the copied error message :)

MFransen69 (changes requested):

MFransen69 (approved):

PR #589, opened 2021-05-19
Proxy: use memmove for address overlapping cases

HaseenaSainul:


mhughesacn:
Hi Haseena: This repo also now expects a Metrological header.  Pierre needs to merge pull/462 first, then you should pick up the changed headers and this should go away.
Thanks,
Martin Hughes RDK CMF Compliance Team

HaseenaSainul:
Ok sure @mhughesacn

pwielders:
@HaseenaSainul just merged the pull @mhughesacn was refering to so if you rebase, the pull should be oke again

HaseenaSainul:
@pwielders, @mhughesacn, these changes also I cherry-picked from master to R2. currently the pull/462 is only applicable for master. So what we have to follow for R2 (since all the files in R2 is not get updated, do we need to update this alone)?  Here rebase will not work, since the origin branches are different for my development branch.

pwielders (approved):

PR #593, opened 2021-05-20
Extend Web::Authorization class to support BASIC header

susanmary007:
Extend Web::Authorization class to support BASIC header.
If username and password is needed for a web request, base64 encoded value of "username:password"
should be generated and passed as BASIC header value for the WebRequest.

### DIFF
--- a/Source/websocket/WebRequest.h
+++ b/Source/websocket/WebRequest.h
@@ -157,7 +157,8 @@ namespace Web {
     class EXTERNAL Authorization {
     public:
         enum type {
-            BEARER
+            BEARER,
+            BASIC
         };
 
     public:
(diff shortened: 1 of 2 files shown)


pwielders (approved):

mhughesacn:
Hi SusanMary: The headers on Thunder have been changed to "Metrological" from "RDK Management" on the request of RDKM.  This was merged to "master".  The tools are now set up to check for Metrological, so to fix your PR, you need to change line 4 of your files to "Copyright 2020 Metrological" instead. Thanks,
Martin Hughes RDK CMF Compliance Team

rjpdasilva:
Hi @mhughesacn, these changes are not touching any copyright notice, so this should not be a problem, right? Once merged into master from here, it should only take the actual changes, not the "old" copyright notice?

susanmary007:
Created v2(https://github.com/rdkcentral/Thunder/pull/599)

PR #608, opened 2021-05-27
Development/server override

wiswis15:


wiswis15:
I will update the docs once is code is approved.

Source/WPEFramework/ControllerJsonRpc.cpp
```diff
@@ -547,6 +559,202 @@ namespace Plugin {
...
+    // Return codes:
+    //  - ERROR_NONE: Success
+    //  - ERROR_GENERAL: Failed to set the version
+    uint32_t Controller::set_version(const Core::JSON::String& params)
+    {
+         uint32_t result = Core::ERROR_NONE;
+        
+        ASSERT(_pluginServer != nullptr);
+        
+        if (_pluginServer ==nullptr)  {
```
MFransen69: In principle the ASSERT should suffice but the way the if is now I predict it will never return the error but crash Thunder :)

Source/WPEFramework/PluginServer.h
```diff
@@ -1048,6 +1048,214 @@ namespace PluginHost {
...
+                    result = true;
+
+                    // Clear all currently set values, they might be from the precious run.
+                    Clear();
+
+                    // Red the file and parse it into this object.
+                    IElement::FromFile(storage);
+
+ 
+                    printf("Version read = [%s]\n", Version.Value().c_str());
```
MFransen69: Guess this is debug code that needs to be removed?

MFransen69 (changes requested):

MFransen69:
Indeed do not forget to generate the documentation and let's also ask Divya to test this feature when it is merged

Source/WPEFramework/Config.h
```diff
@@ -709,6 +721,9 @@ namespace PluginHost {
         inline uint16_t IdleTime() const {
             return (_idleTime);
         }
+        inline void SetIdleTime(const uint16_t& newValue)  {
```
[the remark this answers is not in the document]
wiswis15: mmm, confusing because some discussions point to a better performance by using const.  
But I have mainly used them for code readiness to  indicate that parameter shall  not be changed by the functions.
I removed the new ones , but for info, many other files use const for simple types: example: GATTSocket.h, ControllerJsonRpc.cpp...maybe we can clean them up later.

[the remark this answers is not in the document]
MFransen69: Eh think we do not understand each other. It was not about the const. I'll give you a call

Source/WPEFramework/Config.h
```diff
@@ -718,9 +733,15 @@ namespace PluginHost {
         inline int32_t Latitude() const {
             return (_latitude);
         }
+        inline void SetLatitude(const int32_t& newValue){
+            _latitude = newValue;
+        }
         inline int32_t Longitude() const {
             return (_longitude);
         }
+        inline void SetLongitude(const int32_t& newValue){
```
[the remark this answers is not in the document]
wiswis15: as discussed , i added a critical session

Source/WPEFramework/ControllerJsonRpc.cpp
wiswis15: Assert is only in debug mode right ? indeed Thunder will crash in that case. I corrected the logic in all functions.

Source/WPEFramework/ControllerJsonRpc.cpp
```diff
@@ -547,6 +559,198 @@ namespace Plugin {
...
+    //  - ERROR_NONE: Success
+    //  - ERROR_GENERAL: Failed to get the version
+    uint32_t Controller::get_version(Core::JSON::String& response) const
+    {
+        uint32_t result = Core::ERROR_NONE;
+   
+        ASSERT(_pluginServer != nullptr);
+        
+        if (_pluginServer !=nullptr)  {
+            response = _pluginServer->_config.Version();
```
MFransen69: Sorry for not noticing earlier. It is not the idea to use private members from another class, in this case _config, it breaks the idea of encapsulation. The reason it even compiles here is that there is that controller is a fiend class, guess to allow it to access a private method that should not be accessible by something else then the Controller class (indeed I guess a more specific friend to only allow the method inside controller using it access would have been better bit that is sometimes harder to achieve). But there is a public Configuration method to access the configuration, better to use that instead of _config (e.g. when somebody would rename _config, implement it differently, e.g. add lazy loading, make a aggregate via a pointer then unexpectedly the code would break because of the above usage of the private _config data member). And of course this goes also for the usage in the other methods below.

MFransen69 (changes requested):

Source/WPEFramework/Config.h
```diff
@@ -758,6 +791,7 @@ namespace PluginHost {
             return (added);
         }
         void UpdateAccessor() {
+            Core::SafeSyncType<Core::CriticalSection> scopedLock(_configLock);
```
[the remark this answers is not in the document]
wiswis15: changed as discussed

Source/WPEFramework/PluginServer.h
```diff
@@ -2978,7 +2978,8 @@ namespace PluginHost {
         {
             _dispatcher.Revoke(job);
         }
-        inline const PluginHost::Config& Configuration() const
```
MFransen69, line 3052: I would not remove the const version. I understand you now need a non const version but perhaps somebody in the future might have a const pluginhist and wishes to read the config, then he would not be able to do so (without const cast, but I think we agree that would not be a good solution, So I would leave the const version as well.

MFransen69 (changes requested):

MFransen69 (approved):

vijs (approved):

PR #640, opened 2021-07-12
[WarningReporting] Configuration propagated to OOP plugins, passing warnings to plugin, Serialization done

lukiwa:


mhughesacn:
Hi Lukiwa - the Blackduck complaint is because you need to change the RDK header to Metrological please.
- Martin (RDK CMF Compliance Team)

Source/core/WarningReportingControl.h
```diff
@@ -146,6 +147,30 @@ namespace Core {
 
 namespace WarningReporting {
 
+    class EXTERNAL ExcludedWarnings {
```
MFransen69: Better make it consistent with our other code (we alwyas add constructor (not defaulkt), destructor, (deleted) copy and assignment operators)

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -113,112 +106,72 @@ namespace WarningReporting {
         ASSERT(doorBell.empty() == false);
 
         if (fileName.empty() == false) {
-       
-            fileName +=  '.' + Core::NumberType<uint32_t>(identifier).Text();
+
+            fileName += '.' + Core::NumberType<uint32_t>(identifier).Text();
```
MFransen69, line 106: Just to make sure diud we check if this now doesn't colide with tracing? (I perhaps just copied the tracing code)

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -113,112 +106,72 @@ namespace WarningReporting {
...
     uint32_t WarningReportingUnit::Open(const string& pathName)
     {
         string fileName(Core::Directory::Normalize(pathName) + CyclicBufferName);
-        #ifdef __WINDOWS__
-        string doorBell("127.0.0.1:62002"); 
-        #else
-        string doorBell(Core::Directory::Normalize(pathName) + CyclicBufferName + ".doorbell" );
-        #endif
+#ifdef __WINDOWS__
+        string doorBell("127.0.0.1:62002");
```
MFransen69, line 117: Also here is the number different than the one for tracing?

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -113,112 +106,72 @@ namespace WarningReporting {
...
-                    (*index)->Enabled(enable);
-                }
-            } else {
-                //Disable/Enable traces for all modules
-                modifications++;
-                (*index)->Enabled(enable);
-            }
-
-            index++;
+        std::list<string> result;
```
MFransen69, line 154: Didn't check, perhaps this is not called while mutations are possible but just to make sure, shouldn't this be protected by  _adminlock?

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -234,120 +187,87 @@ namespace WarningReporting {
...
-        return isDefaultCategory;
     }
 
     void WarningReportingUnit::ReportWarningEvent(const char identifier[], const char file[], const uint32_t lineNumber, const char className[], const IWarningEvent& information)
     {
 
         const char* fileName(Core::FileNameOnly(file));
 
-        m_Admin.Lock();
+        Core::SafeSyncType<Core::CriticalSection> guard(_adminLock);
```
MFransen69: As mentioned above we have a rule to always use Lock and Unlock to prevent the lock being taken too long. Call me if you want the details, nice for a debate over a bear. Only exception is onelines for which SafeSyncType can be used

MFransen69 (changes requested):

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -189,7 +189,7 @@ namespace WarningReporting {
 
     void WarningReportingUnit::UpdateEnabledCategories(const Core::JSON::ArrayType<Setting::JSON>& info)
     {
-        Core::SafeSyncType<Core::CriticalSection> guard(_adminLock);
+        _adminLock.Lock();
```
MFransen69: You don't have to change this but this is an example where we now could move the Lock after the next statement, so take the lock a little shorter, but as influence will be minimal no need to change it

Source/warningreporting/WarningReportingUnit.cpp
```diff
@@ -279,6 +282,8 @@ namespace WarningReporting {
             fflush(stdout);
         }
 
+        _adminLock.Unlock();
```
MFransen69: And I think here we could move the Unlock up. At least to just after the _directOuput == true (but does that even need checking? Can it change at runtime?

MFransen69 (changes requested):

MFransen69 (dismissed):

MFransen69 (approved):

bramoosterhuis:
@MFransen69  @lukiwa 
WarningReporting is an optional feature, by this change it will become mandatory.... It breaks my build since don't need/want this feature to be enabled

The option is here:
https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L66-L67

bramoosterhuis:
@lukiwa  @MFransen69, Please review this fix:
https://github.com/rdkcentral/Thunder/pull/700

PR #642, opened 2021-07-13
Development/unavailable

pwielders:
Adding the state Unavailable to a Plugin. 
If a plugin is marked as Unavailable, it can not be started. Unavailable plugins should first be moved to the Deactivated state. Once they are in the Deactivated state they can be started. The reason for adding this state is to send a notification if a plugin is available, keeping Thunder an Event driven system (and no polling is required to see the availability of plugins).

This feature is to be used for:
1) Downloadable plugins, a Plugin responsible for downloading can move the plugin from Unavailable to Deactivated if all the prerequisites (download/decrypt/installation) are completed successfully.
2) Up selling / provisioning / Timed. If certain features should only be made available after it is bought or if the feature needs to be provisioned before it can be used or features that can only be used in specific time-slots..

### DIFF
--- a/Source/WPEFramework/Controller.cpp
+++ b/Source/WPEFramework/Controller.cpp
@@ -373,6 +373,29 @@ namespace Plugin {
                         }
                     }
                 }
+            } else if (index.Current() == _T("Unavailable")) {
+                if (index.Next()) {
+                    const string callSign(index.Current().Text());
+                    if (callSign == _service->Callsign()) {
+                        result->ErrorCode = Web::STATUS_FORBIDDEN;
+                        result->Message = _T("The PluginHost Controller can not set Unavailable.");
+                    } else {
+                        Core::ProxyType<PluginHost::Server::Service> pluginInfo(FromIdentifier(callSign));
+
+                        if (pluginInfo.IsValid()) {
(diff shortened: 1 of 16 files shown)


Source/WPEFramework/Controller.cpp
```diff
@@ -373,6 +373,29 @@ namespace Plugin {
...
+                        Core::ProxyType<PluginHost::Server::Service> pluginInfo(FromIdentifier(callSign));
+
+                        if (pluginInfo.IsValid()) {
+                            if (pluginInfo->State() == PluginHost::IShell::DEACTIVATED) {
+                                // Mark the plugin as unavailable.
+                                if (pluginInfo->Unavailable(PluginHost::IShell::REQUESTED) != Core::ERROR_NONE) {
+                                    result->ErrorCode = Web::STATUS_NOT_MODIFIED;
+                                    result->Message = _T("Marking the plugin as Unavailble failed.");
+                                }
+                            }
```
MFransen69, line 392: Know it is also not done above but as this is a new case which might be used by users perhaps less aware of the allowed state transitions perhaps have an else here to make that more clear?

MFransen69 (dismissed):

MFransen69 (approved):

PR #643, opened 2021-07-13
Unavailable state for plugins

pwielders:
Adding the state Unavailable to a Plugin.
If a plugin is marked as Unavailable, it can not be started. Unavailable plugins should first be moved to the Deactivated state. Once they are in the Deactivated state they can be started. The reason for adding this state is to send a notification if a plugin is available, keeping Thunder an Event driven system (and no polling is required to see the availability of plugins).

This feature is to be used for:

Downloadable plugins, a Plugin responsible for downloading can move the plugin from Unavailable to Deactivated if all the prerequisites (download/decrypt/installation) are completed successfully.
Up selling / provisioning / Timed. If certain features should only be made available after it is bought or if the feature needs to be provisioned before it can be used or features that can only be used in specific time-slots..


goruklu (approved):

goruklu (approved):

pwielders:
Just let me know if I can merge it..

goruklu:
@pwielders yes, please go ahead and merge!

PR #645, opened 2021-07-14
Make sure to export cmake modules outside Thunder

404ag:
Previously, cmake/modules were not installed and not added to the
CMAKE_MODULES_PATH. This resulted that external components,
that are using find_package(WPEFrameowrk), couldn't acccess common modules.

This is now changed, so that Thunder will install the modules in the
sysroots /usr/lib/cmake/WPEFramework. It will export that path,
so the find_package(WPEFramework) will append CMAKE_MODULE_PATH
with module location alowing usage of the modules.


bramoosterhuis:
@artur-gebicz-ml the modules stay private to this repo. Installing this basically will introduce a  "hidden" dependency to other projects. If a project/repo needs one of these modules we should add them to that repo. I know this is code duplication but it keeps the look-up of dependencies independent of other projects.

bramoosterhuis (commented):
@artur-gebicz-ml the modules should stay private to this repo. Installing this basically will introduce a "hidden" dependency to other projects. If a project/repo needs one of these modules we should add them to that repo. I know this is code duplication but it keeps the look-up of dependencies independent of other projects.

404ag:
Closing this PR as per @bramoosterhuis  comments.
Solution for the problem solved on:
https://github.com/WebPlatformForEmbedded/OCDM-Playready-Nexus-SVP/pull/20
https://github.com/WebPlatformForEmbedded/OCDM-Widevine-Nexus-SVP/pull/15

PR #658, opened 2021-07-24
[FEATURE] Expose non-happyday scenario on comrpc boundaries.

pwielders:


pwielders:
This pull-request exposes COMRPC functionality to which a plugin can subscribe to get notifications on objects that are dropped due to a dropped connection (other process crashed). This way subscribed proxies (allocated memory) that are now no longer operable can be actively removed (and deleted).
Implementation (done by @lukiwa ) led to a bug that is fixed with this pull-request as well. In the non-happy days scenario, the object is reported as being inoperabe (revoked) but was not actually deleted. This would cause a potential memory leak for proxies that never get deleted for crashing processes.

MFransen69 (approved):

lukiwa (approved):

vijs (approved):

PR #665, opened 2021-07-29
DELIA-51490: Thunder error codes

npoltorapavlo:
Reason for change: json rpc errors should indicate
ERROR_TIMEDOUT (11) if timed out,
ERROR_PLUGIN_NOTACTIVATED (46) if not activated,
or ERROR_ASYNC_FAILED (3) otherwise.
Test Procedure: See ticket.
Risks: Low


anand-ky:
To add more context, currently we see ERROR_ASYNC_FAILED returned for a lot of thunder errors. It would help in debugging and triaging thunder related issues if we can distinguish between the different errors as listed above.

anand-ky:
@pwielders - Pls review when you get a chance.

pwielders:
Most of the changes are oke, however the addition of the ERROR_PLUGIN_NOTACTIVATED error state does not seem like a good move.
The only think the system knows at that time was that the request issued could not be completed. If you look at the error condition: "if ( (_jsonrpc == nullptr) || (IsActive() == false) ) {" there are multiple reasons why it could not be completed. If the plugin does not support a JSONRP interface (_jsonrpc == nullptr) the plugin could well be activated and the result => ERROR_PLUGIN_NOTACTIVATED would be completely bogus. I suggest, if ERROR_BAD_REQUEST does not cover the message, use ERROR_UNAVAILABLE (already present) as that does cover the load better.

Further more, I think we should not mix application layer and framework layer. Th fact that a plugin has states (ACTIVATED/DEACTIVATED/SUSPENDED/RESUMED) is a WPEFramework application layer domain knowledge. The error codes are defined on a CORE level (framework, *NO* application knowledge) so you should not use state information from higher layers in the lower layers.

With respect to:  "ERROR_ASYNC_FAILED returned for a lot of thunder errors" is hopefully not what you mean cause if this error is returned for indicated different issues to a callee that would be a bug and needs to be solved. I think, and hope that you mean "ERROR_ASYNC_FAILED returned for a lot of thunder *calls* " and agree that is as expected because a lot of calls issue a-synchronous actions that could fail and indeed they will all report this with this distinct error.

Source/WPEFramework/PluginServer.h
```diff
@@ -661,7 +661,7 @@ namespace PluginHost {
                     Unlock();
 
                     result = Core::proxy_cast<Core::JSONRPC::Message>(Factories::Instance().JSONRPC());
-                    result->Error.SetError(Core::ERROR_BAD_REQUEST);
+                    result->Error.SetError(Core::ERROR_PLUGIN_NOTACTIVATED);
```
pwielders: I think ERROR_PLUGIN_NOTACTIVATED here should be ERROR_UNAVAILABLE, as indicated, the plugin might be active but just has no JSONRPC interface.

Source/core/Portability.h
```diff
@@ -706,7 +706,8 @@ namespace Core {
         ERROR_CODE(ERROR_UNAUTHENTICATED, 42) \
         ERROR_CODE(ERROR_NOT_EXIST, 43) \
         ERROR_CODE(ERROR_NOT_SUPPORTED, 44) \
-        ERROR_CODE(ERROR_INVALID_RANGE, 45)
+        ERROR_CODE(ERROR_INVALID_RANGE, 45) \
+        ERROR_CODE(ERROR_PLUGIN_NOTACTIVATED, 46)
```
pwielders: Mixing domain logic from higher layers  (WPEFramework knowledge) in lower layers is not according to the Thunder architecture. So I think we can remove this error, and as inidcated use the UNAVAILABLE.

pwielders (changes requested):
As the process for us is to start with this on master, could you submit it to the master, and if required by customers (e.g. Comcast/NOS/MCA/VF/KPN etc) we can "cherry-pick" it to those customer release branches . This way we make sure that changes you introduce on "older/other" versions, do not get lost for future releases.

Source/WPEFramework/PluginServer.h
anand-ky: sounds good. We can have ERROR_UNAVAILABLE.

anand-ky:
yes, I meant ERROR_ASYNC_FAILED returned for a lot of thunder calls. 
@npoltorapavlo - Pls make the suggested changes and submit it to master.

npoltorapavlo:
PR #674 for branch master

PR #670, opened 2021-08-03
Development/metrol 333

SKumarMetro:


Source/core/ProcessInfo.cpp
```diff
@@ -61,60 +61,63 @@ namespace Core {
...
-                    if ((size = read(fd, buffer, maxLength - 1)) > 0) {
-                        if (buffer[size - 1] == '\n') {
-                            buffer[size - 1] = '\0';
-                        } else {
-                            buffer[size] = '\0';
-                        }
-                    } else {
-                        buffer[0] = '\0';
-                    }
+        std::stringstream ss;
```
lukiwa, line 87: I know I did something similar later in this file ("pathToSmaps"), but on second thought I think we could replace this with string creation, as it seems to be more performant. 

https://quick-bench.com/q/SI-XzZyfd7YimTzp6dbdEQG__xo

Let me know what do you think and if you agree, could you also change stringstream I created?

lukiwa (changes requested):

PR #678, opened 2021-08-07
Fixed 'substitution failure' related errors with _CalculateHash(arg1) f…

Vinodsathyaseelan:
…unction.

Added missed function for enable_if <false> condition.
A few other compilation erros fixed to include Echo test code in ThunderNanoServices.


Vinodsathyaseelan:
Found some issues during testing. So closing this PR.

PR #688, opened 2021-08-16
[METROL-345][TypeTraits.h] Use STL implementations where possible

lukiwa:
Maintaining backward compatibility via _using_ ...


MFransen69 (approved):

MFransen69:
Just to make sure (as these low level changes could easily break code at other places): you have done some testing with these changes?

lukiwa:
Yes, I tested if the old implementation gives the same result as the STL one.

PR #694, opened 2021-08-18
Fix for converting shortUuid into long uuid

Preejar:
This patch is for getting 128-bit UUID value.A UUID is 128-bit
 value and to reduce burden of storing and transferring 128-bit
 value, range of values preallocated as shortUUIDs in 16-bit.
 All the standarad UUIDs represented in shortUUID and non-standard
 UUID(eg:OAD, voice) represented in 128-bit full value.

 There is already an existing api named ToString() which can be used
 to get both long and short Uuids as string value. But this api is
 representing first 32-bit value of UUId in reverse order(for example:
 0002a26-0000-1000-8000-00805f9b34fb(128-bit value) representing as
 2a26000-0000-1000-8000-00805f9b34fb).They are representing value like
 this for make easy to calculate shortUUID from this as 0x2a26.
 To Fix this introduced new api to get original longUUID value

### DIFF
--- a/Source/bluetooth/Profile.h
+++ b/Source/bluetooth/Profile.h
@@ -368,6 +368,14 @@ namespace Bluetooth {
                         }
                         return (result);
                     }
+                    string Uuid(bool full) const {
+                        bool naturalOrder = false;
+                        
+                        if (full && _uuid.HasShort())
+                            naturalOrder = true;
+
+                        return _uuid.ToString(full, naturalOrder);
+                    }
                     uint16_t Handle() const {
                         return (_handle);
                     }
@@ -445,6 +453,15 @@ namespace Bluetooth {
 
                     return (index != _descriptors.end() ? &(*index) : nullptr);
                 }
+                string Uuid(bool full) const {
+                    bool naturalOrder = false;
+
+                    if (full && _type.HasShort())
+                        naturalOrder = true;
+
+                    return _type.ToString(full, naturalOrder);
+                }
+
                 uint16_t Handle() const {
                     return (_handle);
                 }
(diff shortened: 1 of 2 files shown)


sebaszm (commented):
Good catch. But I think we should fix the underlying problem and put the short UUID value in proper place (i.e. index 13/14 instead of 15/16) rather than modify the string printer to cover the problem up.

pwielders:
@sebaszm I am not sure how to fix the underlying issue. Could you eleborate a little more on what we need to do here or is this fix needed anyway ?

sebaszm:
The UUID code is assuming that the 16-bit UUID is placed here xxxx0000-0000-1000-8000-00805f9b34fb, but in fact it's placed here 0000xxxx-0000-1000-8000-00805f9b34fb (and the 32-bit UUID is here yyyyyyyy-0000-1000-8000-00805f9b34fb). So, code that is comparing and converting UUIDs from 16bit to 128bit and vice versa should be corrected. Then the stringification code would need no modification for correct output.

Preejar:
Could you please pointing me where this conversion happening. I mean from the rcu we are getting uuids in the format as  0000xxxx-0000-1000-8000-00805f9b34fb right? I couldnt able to understand where this bit conversion happening to calculate short uuid

sebaszm:
@Preejar I believe changes in https://github.com/rdkcentral/Thunder/pull/714 will solve your issue, please check.

pwielders:
@Preejar could you make sure your issue is resolved with #714 and than close this pull request ?

Preejar:
Yes Sure.I will update after doing testing

pwielders:
@Preejar any update on this? Can we close it ?

Preejar:
> @Preejar any update on this? Can we close it ?

Sorry for the delay. I will update you soon by today or tomorrow.

Preejar:
Hi @pwielders ,

I tested with  #714  and working for me

rjpdasilva:
@pwielders, @MFransen69, this one can be closed, we are using your proposed solution already.

MFransen69:
Closed as agreed

PR #695, opened 2021-08-18
to  add custom test cases to handle utf-8 strings

kkumar495:
to fix the build error :
```
[ 68%] Linking CXX shared library libWPEFrameworkCOM.so
/usr/bin/ld: cannot find -lWPEFrameworkWarningReporting::WPEFrameworkWarningReporting
collect2: error: ld returned 1 exit status
```
Add custom test cases in json_testparser.cpp to check if thunder is able to handle utf-8 strings


Source/com/Communicator.h
```diff
@@ -1152,11 +1154,14 @@ namespace RPC {
 
                     // Anounce the interface as completed
                     string jsonDefaultCategories(Trace::TraceUnit::Instance().Defaults());
+#ifdef WARNING_REPORTING
                     string jsonDefaultWarningCategories(WarningReporting::WarningReportingUnit::Instance().Defaults());
+#endif
                     void* result = _parent.Announce(proxyChannel, message->Parameters());
 
+#ifdef WARNING_REPORTING
                     message->Response().Set(instance_cast<void*>(result), proxyChannel->Extension().Id(), _parent.ProxyStubPath(), jsonDefaultCategories, jsonDefaultWarningCategories);
```
MFransen69: I just checked and the unintended forced WARNING_REPORTING dependencies here and above were already fixed in the mean time ( that is the reason the PullRequest has merge conflicts) (and just to make sure this line is not only needed when warning reporting is active so it has been corrected in a different way). 
I propose to just remove this and above changes from the pull requests, that will also solve the merge conflicts.

Tests/unit/IPTestAdministrator.h
```diff
@@ -24,6 +24,10 @@
 #include <string>
 #include <time.h>
 
+#ifndef MODULE_NAME
+#define MODULE_NAME Tracing
```
MFransen69: Perhaps for a test not too important but this module name might be confusing :)

MFransen69 (changes requested):

kkumar495:
changes are made  as per the review comments.

pwielders:
@kkumar495 Change set is oke now, but you must sign the CLA before we can merge it!

kkumar495:
> @kkumar495 Change set is oke now, but you must sign the CLA before we can merge it!

Done

Source/com/Communicator.h
kkumar495: removed this change

MFransen69 (approved):

PR #701, opened 2021-08-23
Development/proxy release fix

HaseenaSainul:


Source/com/Administrator.cpp
```diff
@@ -118,6 +120,9 @@ namespace RPC {
                 if (index->second.size() == 0) {
                     _channelProxyMap.erase(index);
                 }
+                if ((*entry)->RefCount() == 0) {
```
pwielders, line 123: If the RefCount() == 0, it maeasn the object is destructed deleted, so I think you might be poking here in already released (freed memory).

HaseenaSainul: In this sequence we are not deleting the UnkownProxy object, instead we are sending invoke calls to do the release from the other end:https://github.com/rdkcentral/Thunder/blob/582bf30a716dd1004d6e79a1fe210080e64ef8b2/Source/com/IUnknown.h#L219 and UnknownProxy is deleting from https://github.com/rdkcentral/Thunder/blob/582bf30a716dd1004d6e79a1fe210080e64ef8b2/Source/com/IUnknown.h#L431-L440.
So the refcount usage is just indicating the usage status. 0 means none of them is used

HaseenaSainul:
Closing this PR, since it already handle using the PR: https://github.com/rdkcentral/Thunder/pull/725.

PR #704, opened 2021-08-24
RDK-33274: Add ProcessContainers SearchPath config option

TeknoVenus:
Allow searching for containers in a configurable location when using processcontainers. Uses
the plugin classname or `PersistentPathPostfix` value (if set) to lookup configs instead of the callsign so cloned plugins can share the
same container config. 

Allows for different permissions to be set on container bundle directories.

In `/etc/WPEFramework/config.json`, configure searchPath as follows:
```
 "processcontainers":{
  "searchpath":"/Containers"
 },
```

When built without ProcessContainers support, disable this feature.


goruklu (approved):

pwielders (changes requested):
I am a bit lost. Is there some kind of description of this new feature? It looks like a plugin specific search path, but it is implemented as a generic search path.
The concept in Thunder is to keep functionality together. It looks like this is creating a backdoor feature where containers can start to live anywhere as there is a new "global" search path.
The design discussion we had when we designed the containers is that the containers would adhere to the generic Thunder concept. 
- VolatilePath, memory not persisting over a boot.
- PersistentPath, on writable flash/hdd, persisting over boots.
- DataPath, ROM, in image, part of the flash image.

All plugins, sofar, created adhere to this concept. We would not like to deviate from it for a process (or container process) which is just an implementation detail unless there is a very good use case for it.

What would be the use case for it? Why are the already defined paths not sufficient ? What is special on this flavor ?

pwielders:
If the use case is:
"so cloned plugins can share the same container config" I suggest using the solution we discussed with the multiple WebKit Instances where the WebKitPlugin wanted to share the same space between the different instances. For this feature we added the PersistentPathPostfix and VolatilePathPostfix, see: https://github.com/rdkcentral/Thunder/blob/master/Source/plugins/Configuration.h#L67-L68

TeknoVenus:
Hi Pierre,

Apologies for any confusion. Let me try to explain this a bit clearer.

As you say, right now every plugin has 3 storage locations:

* Volatile - e.g. `/tmp/<callsign>` (r/w)
* Data - e.g. `/usr/share/WPEFramework/<classname>` (ro)
* Persistent - e.g. `/opt/persistent/rdkservices/<callsign>` (rw)

When a plugin is configured to run in a container, ProcessContainers looks in these 3 locations for a directory called `Container` for the container bundle/configuration.

The aim of this PR is to create a 4th location that ProcessContainers will use (in addition to the previous 3 paths) to find container bundles

* ProcessContainers Path - e.g. `/opt/containers/<classname>`

The base location of this path is configured in the main WPE config file the same as the other types of path.

One aspect of this change is allowing containers to be stored in a r/w location based on classname instead of callsign to allow cloned plugins to re-use the same container config. A r/w location is necessary for Dobby containers. As you mentioned, this can be achieved with the `PersistentPathPostfix` option

However, in the case of WebKitBrowser, we need to mount the persistent directory inside the container, since the plugin uses this directory to store localstorage etc. If we create the container config inside this directory the container config will end up mounted inside the container. Since we need to give r/w access to the persistent directory (so that the browser can actually store stuff in there!), that would allow the container to manipulate it's own configuration file which would be a major security risk.

As a result, it is desirable to store container bundles in a separate area of flash where permissions can be set accordingly and no plugin can manipulate it's own configuration.

If you have a preferred way of achieving this, or if this still isn't clear please let me know.

pwielders:
Thanks for the elboration. I think the current solution would still be sufficient as (looking at webkitplugin) you could configure the container (with the PersistentPathPostfix) assuming it is set to shared, for cloned instances into:
<PersistentPath>/shared/container (for the container definition) and placing the LocalStorage in <PersitentPath>/shared/data for example. using rights managments (linux), you make sure that the rights of <persistentPath>/shared/container can only be accessed by root and the ../data path only by the WebKit user.
So I think I still do not get the usecase completely. SHould we maybe setup a conf-call to discuss this ?

TeknoVenus:
Hi @pwielders 

Looking at this again, I tried your suggestion above at that seems to work OK - the container bundle and persistent storage can be moved to sub-dirs of the existing persistentstorage directory.

Will close this PR. Thanks for the feedback.

PR #711, opened 2021-08-30
Modified String Deserialize() to handle unicode character.

ghost:


pwielders:
@neeraj-deshpande I guess this is now superseeded by: https://github.com/rdkcentral/Thunder/pull/770 since there we have Serialization *and* deserialization of UTF8.

PR #712, opened 2021-08-31
Development/psgen no register on null

sebaszm:
Don't register null interfaces in stubs. (Also adds some logging fixes.)


HaseenaSainul (approved):
LGTM

PR #714, opened 2021-08-31
[Bluetooth] Fix converting 16-bit UUID to 128-bit UUID

sebaszm:


pwielders:
Does this supersede #694 ?

pwielders (approved):

PR #721, opened 2021-09-03
[METROL-353][Websocket] Pick the first supported protocol and respond with it

lukiwa:


Source/websocket/WebSocketLink.h
```diff
@@ -651,6 +651,20 @@ namespace Web {
             {
                 return (_protocol);
             }
+            inline bool IsProtocolSupported(const string& protocol)
```
pwielders: This is "higher level knowledge" and should not be inserted here.
 Do this check in the PluginServer, the server that really implements the sub protocolas and thus knows that it is possible.
here: https://github.com/rdkcentral/Thunder/blob/a42f71492333aafde27f9c431408721e7727c4f4/Source/websocket/WebSocketLink.h#L1000 the derviced class (e.g. the plugin server) gets an opportunity to "abort" an in-progress transition to a WebSocket. Check (and set) the used protocol, than in the parent (plugin server) and if the protocol is not supported, abort the upgrade through the AbortUpgrade method with a proper error.

pwielders (changes requested):

Source/core/TokenizedStringList.h
```diff
@@ -0,0 +1,108 @@
...
+
+    string All() const
+    {
+        string result;
+
+        if (!_parts.empty()) {
+
+            for (const auto& part : _parts) {
+                result += part;
+                result += ',';
```
pwielders: I think we want the SEPARATOR here and not the ','. If you update it also hed the Copyright header notice to this file ;-)

Source/core/TokenizedStringList.h
```diff
@@ -0,0 +1,108 @@
...
+    ~TokenizedStringList() = default;
+    TokenizedStringList(const TokenizedStringList&) = default;
+    TokenizedStringList& operator=(const TokenizedStringList&) = default;
+
+    explicit TokenizedStringList(const string& buffer)
+    {
+        Tokenize(buffer);
+    }
+
+    void Tokenize(string buffer)
```
pwielders: should this not be a const string& for buffer?

pwielders (changes requested):

Source/core/TokenizedStringList.h
```diff
@@ -0,0 +1,111 @@
...
+            string result;
+
+            if (!_parts.empty()) {
+
+                for (const auto& part : _parts) {
+                    result += part;
+                    result += SEPARATOR;
+                }
+
+                result.erase(result.rfind(','));
```
pwielders: I am afraid rgat this should read result.erase(result.rfind(SEPARATOR));  :-)
And could you add the copyright header to this file (C) metroogical blah blah from another file ..

Source/core/TokenizedStringList.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
+#include "Module.h"
+
+namespace WPEFramework {
+namespace Core {
+
+    template <char SEPARATOR, bool REMOVE_WHITESPACES>
+    class TokenizedStringList {
```
pwielders, line 27: While looking at the code once more,  I saw a lot of "forwarding". I was wondering why don't you just inherit from std::vector<string> ? Than you can drop all forwarders and I do not think there is anything specific you would like to hide, right ?

Source/core/TokenizedStringList.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
+#include "Module.h"
+
+namespace WPEFramework {
+namespace Core {
+
+    template <char SEPARATOR, bool REMOVE_WHITESPACES>
```
pwielders: I know we do not actively test/use the widechar (wchar_t) but if for some reason we should, all code is prepared for it, by using preprocessor definitions, and it might work (or at least we have a lot less work :-) ). 
To be inline with that possibility, pleas make the char -> TCHAR which for us translates to char, but in the future might translate to wchar_t.

Source/core/TokenizedStringList.h
lukiwa: I tend not to use inheritance of std::vector, as the STL containers do not have virtual destructors, and that prevents polymorphic use of derived classes

pwielders: Good point !

pwielders (approved):

PR #729, opened 2021-09-13
Changes for automatic documentation using jsongenerator

nkkrishnan2512:
This PR contains the changes for automatic documentation for wpeframework plugins using jsongenerator.This adds a function
DocGenerator() which contains the logic to generate document automatically and installs it to image.
This PR also contains option for including the cpp interface headerfile path that will substitute the {cppinterfacedir} tag
in JsonGeneratortool command for document generation.


sebaszm (approved):

rjpdasilva:
@pwielders, @sebaszm, can we have this also merged into the `R2` branch?

PR #732, opened 2021-09-16
[Metrol305] Changes for notifying syslog logging.

Karthick-Somasundaresan:
When WPEFramework does a Syslog, it will notify the
message to the registered monitoring clients

### DIFF
--- a/Source/tracing/ISyslogMonitorClient.h
+++ b/Source/tracing/ISyslogMonitorClient.h
@@ -20,11 +20,10 @@
 #include "Module.h"
 namespace WPEFramework {
 namespace Logging {
-    class EXTERNAL ISyslogMonitorClient
+    struct EXTERNAL ISyslogMonitorClient
     {
-        public:
-            virtual void NotifyLog(const std::string& logmessage) = 0;
-            virtual ~ISyslogMonitorClient(){}
+        virtual void NotifyLog(const std::string& logmessage) = 0;
+        virtual ~ISyslogMonitorClient() = default;
     };
 }
 }
--- a/Source/tracing/ISyslogMonitorClient.h
+++ b/Source/tracing/ISyslogMonitorClient.h
@@ -1,0 +1,30 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2021 Metrological Management
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Module.h"
+namespace WPEFramework {
+namespace Logging {
+    class EXTERNAL ISyslogMonitorClient
+    {
+        public:
+            virtual void NotifyLog(const std::string& logmessage) = 0;
+            virtual ~ISyslogMonitorClient(){}
+    };
+}
+}
--- a/Source/tracing/SyslogMonitor.h
+++ b/Source/tracing/SyslogMonitor.h
@@ -28,37 +28,51 @@ namespace Logging {
             SyslogMonitor& operator= (const SyslogMonitor& ) = delete;
             void RegisterClient(ISyslogMonitorClient* client)
             {
-                _monitorClients.push_back(client);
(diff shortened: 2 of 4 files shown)


Source/tracing/SyslogMonitor.h
```diff
@@ -0,0 +1,68 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Module.h"
+namespace WPEFramework {
+namespace Logging {
+    class EXTERNAL SyslogMonitorClient
```
sramani-metro: If this is intended to be an interface, then better to follow the style of declaring it as a struct with a virtual destructor with naming conventions like ISyslogClient or so.

Source/tracing/SyslogMonitor.h
```diff
@@ -0,0 +1,68 @@
...
+            void UnregisterClient(SyslogMonitorClient* client)
+            {
+                auto iter = std::find(begin(_monitorClients), end(_monitorClients), client);
+                if( iter != _monitorClients.end())
+                {
+                    _monitorClients.erase(iter);
+                }
+                return;
+            }
+            static SyslogMonitor& Instance()
```
sramani-metro, line 51: If this is supposed to be a singleton in the Thunder process, why not use the SingletonType Template to make use of the lifecycle management the framework offers?

sramani-metro (changes requested):

sramani-metro (dismissed):

Source/tracing/ISyslogMonitorClient.h
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Module.h"
+namespace WPEFramework {
+namespace Logging {
+    class EXTERNAL ISyslogMonitorClient
```
MFransen69: `for interfaces we usually use struct so we do not have to specify public, improves the readability a bit

Source/tracing/ISyslogMonitorClient.h
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public:
+            virtual void NotifyLog(const std::string& logmessage) = 0;
+            virtual ~ISyslogMonitorClient(){}
```
MFransen69: In interfaces we usually prefer = default instead of an empty implementation. In some cases that might be more performant.

Source/tracing/SyslogMonitor.h
```diff
@@ -0,0 +1,64 @@
...
+namespace WPEFramework {
+namespace Logging {
+    class EXTERNAL SyslogMonitor
+    {
+        public:
+            SyslogMonitor(const SyslogMonitor& ) = delete;
+            SyslogMonitor& operator= (const SyslogMonitor& ) = delete;
+            void RegisterClient(ISyslogMonitorClient* client)
+            {
+                _monitorClients.push_back(client);
```
MFransen69: For safety I would at least add an assert to check if the client is already registered and if they are not trying to register a nullptr

MFransen69: I guess some locking would also not hurt. What would happen if a client would just be (un)registered while at the same time another thread would log something to syslog?

Source/tracing/SyslogMonitor.h
```diff
@@ -0,0 +1,64 @@
...
+                return;
```
MFransen69: Might not be explicitly in our coding guideline but we never use returns without return value. So to keep our code base consistent could you please remove them?

Source/tracing/SyslogMonitor.h
```diff
@@ -0,0 +1,64 @@
...
+            }
+            void UnregisterClient(ISyslogMonitorClient* client)
+            {
```
MFransen69, line 41: would also add an assert for client being nullptr

Source/tracing/ISyslogMonitorClient.h
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+#include "Module.h"
+namespace WPEFramework {
+namespace Logging {
+    class EXTERNAL ISyslogMonitorClient
+    {
+        public:
+            virtual void NotifyLog(const std::string& logmessage) = 0;
```
MFransen69: Why a std::string? We use string to be able to move to multi byte codewords in tghe future (e.g. UTF-16).

MFransen69 (changes requested):

MFransen69:
As discussed: we are redesigning the way reporting is done in general that will have a big influence on this PR or even  makes it not needed anymore. Let's see when the new reporting engine is in place

PR #735, opened 2021-09-20
Modified to remove workaround for OCDM plugin as it no longer works.

ghost:


pwielders:
@neeraj-deshpande I saw some changes in the OCDM server plugin where you are now nolonger using the regex. Is that the reason why we can remove this? How does it relate to the change for the compiler flag addition you added to that module gnu_no_unique ? Is that still needed if you dropped the regex from the OCDM server?

ghost:
@pwielders This work-around solution is not working with gcc 9 in buildroot main.

ghost:
If gnu_no_unique flag is used for OCDM plugin (or SecurityAgent plugin), regex need not be removed.

pwielders (approved):

PR #736, opened 2021-09-23
Build: Custom tracing level

jfifranca:
This adds the capability to set a custom tracing level onto WPEFramework, thus permitting the usage of TRACE_Lx in an easy way.

This is disabled by default. A configuration should be provided to override these settings.


jfifranca:
Wrong email used on push

PR #738, opened 2021-09-23
Build: Custom tracing level

jfifranca:
This adds the capability to set a custom tracing level onto WPEFramework, thus permitting the usage of existing TRACE_Lx in an easy way.

This is disabled by default. A configuration should be provided to override these settings.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -64,6 +64,7 @@ option(WCHAR_SUPPORT
         "Enable support for WCHAR." OFF)
 option(DISABLE_TRACING
         "Disable tracing in debug" OFF)
+set(ENABLED_TRACING_LEVEL 0 CACHE STRING "If tracing is enabled, this sets its default level.")
 option(HIDE_NON_EXTERNAL_SYMBOLS 
         "Hide all non EXTERNAL tagged symbols" OFF)
 
--- a/Source/core/CMakeLists.txt
+++ b/Source/core/CMakeLists.txt
@@ -148,6 +148,9 @@ set(PUBLIC_HEADERS
 if(DISABLE_TRACING)
     target_compile_definitions(${TARGET} PUBLIC _TRACE_LEVEL=0)
     message(STATUS "Force trace level to 0")
+else()
+    target_compile_definitions(${TARGET} PUBLIC _TRACE_LEVEL=${ENABLED_TRACING_LEVEL})
+    message(STATUS "Force trace level to ${ENABLED_TRACING_LEVEL}")
 endif()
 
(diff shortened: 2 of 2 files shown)


pwielders:
We do not want to "promote" the usage of the TRACE_Lx macros since these macros are only intended for the use prioir to having the trace library. In the plugins we only allow the use of the TRACE macro. So this pull request suggest otherwise. Please use the TRACE macro in Plugins, also see: http://thunder.integraal.info/index.php/Tracing_v.s._Logging for a detailed explanation on tracing.

rjpdasilva:
@pwielders, I believe in R2 the TRACE_Lx() macros are still valid, although I understand deprecated.  
As long as they're still defined, I believe this patch still makes sense.
Once they are finally removed, so will be this patch, so no harm done anyway.
Do you see any inconvenient of merging this to R2 even assuming that one day it will just be gone?
Currently, in R2, it is proven useful.

rjpdasilva:
@pwielders, given the comment above, is still your opinion this PR doesn't make sense?  
Just trying to understand, we can always keep the patch on our side and close this PR if you think that's the best thing to do at this point.

rjpdasilva:
Hi @pwielders, so what should be the outcome of this MR, regarding the previous comments?

jfifranca:
Hi @pwielders, any update on the outcome of this MR?

pwielders:
Looked at the pull-request again, looks like it is only a LEVEL determination of the TRACE_Lx macros. We can definitly add this pull request but the only thing we want to make sure is that these Macros (TRACE_Lx) will not (and should not) be used beyond the scope of the core/cryptography/tracing module/libraries. Seems that this allows to externally set the level of the TRACE_Lx macro and the default value is 0 but I guess it does not hurt the system if we add it. Just be aware that you do *not* use these TRACE_Lx macros in plugins/interfaces/client libraries there you should use the TRACE() macros.

pwielders:
@rjpdasilva also in R3, the TRACE_Lx macro is still available and not deprecated, I just want to make sure that this macro is not used in any othere modue than Core/Tracing and Cryptography. By definition they should *not* be used in plugins/interfaces and clientlibraries (in R3 you will get a warning if you use them there). We are trying to remove as much of them as possible from the other Thunder Core libraries.

pwielders (approved):

pwielders:
@jfifranca Could you also cherry pick this to master, so it is also available in the future.

jfifranca:
@pwielders fully agree with your concerns regarding the future usage of TRACE_Lx macros. The sole intention of this change is to just enable on the existing code base.

karuna2git (approved):

PR #740, opened 2021-09-27
Modified SystemInfo.cpp to skip first space while reading Chipset info.

ghost:


Source/core/SystemInfo.cpp
```diff
@@ -521,7 +521,7 @@ namespace Core {
                     if (line.find("Hardware") != std::string::npos) {
                         std::size_t position = line.find(':');
                         if (position != std::string::npos) {
-                            result.assign(line.substr(position + 1, string::npos));
+                            result.assign(line.substr(position + 2, string::npos));
```
pwielders: Can we make this a bit more portable. Now the assumption is that after the ':' 1 space follows, it might be two or it might be none, it is all up to how it is implemented in the proc. Suggest to still take the first characer after the ':' and than trim off any spaces, at the begining, Something like:

// trim from beginning of string (left)
result.erase(0, result.find_first_not_of(_T(" \t\n\r\f\v")));

This way and white space character is erased in front.

pwielders (changes requested):
Make the codemore robust against no or more than 1 space (white character)

pwielders (approved):

PR #747, opened 2021-10-01
Virtual Input: Loading key map from an JSON Array

jfifranca:
Keymaps are curently loaded from a single method (VirtualInput::Keymap::Load()) that requires the key map file name as input paramter. This commit adds a new method to load the keymap from analysing an array of 'KeyMapEntry' data types.
This serves for those use cases where a keymap is embedded on a file containing other types of data.
In order to re-use code, VirtualInput::Keymap::Load() method calls the new VirtualInput::KeyMap::LoadFromArray() method.

### DIFF
--- a/Source/plugins/VirtualInput.cpp
+++ b/Source/plugins/VirtualInput.cpp
@@ -135,6 +135,84 @@ namespace PluginHost
         return (result);
     }
 
+    uint32_t VirtualInput::KeyMap::LoadFromArray(const Core::JSON::ArrayType<KeyMapEntry>& mappingTable)
+    {
+        uint32_t result = Core::ERROR_ILLEGAL_STATE;
+
+        if(mappingTable.Length() > 0) {
+            std::map<uint16_t, uint16_t> previousKeys;
+            std::map<uint16_t, int16_t> updatedKeys;
+
+            result = Core::ERROR_NONE;
+
+            while (_keyMap.size() > 0) {
+                // Keep reference count of repeated keys
+                previousKeys[_keyMap.begin()->second.Code]++;
+
+                // Create an empty list
+                updatedKeys[_keyMap.begin()->second.Code] = 0;
+
+                _keyMap.erase(_keyMap.begin());
+            }
+
+            Core::JSON::ArrayType<KeyMapEntry>::ConstIterator index(mappingTable.Elements());
+            while (index.Next() == true) {
+                if ((index.Current().Code.IsSet()) && (index.Current().Key.IsSet())) {
+                    uint32_t code(index.Current().Code.Value());
+                    ConversionInfo conversionInfo;
+
+                    conversionInfo.Code = index.Current().Key.Value();
+                    conversionInfo.Modifiers = 0;
+
+                    // Build the device info array..
+                    Core::JSON::ArrayType<JSONModifier>::ConstIterator flags(index.Current().Modifiers.Elements());
+
+                    while (flags.Next() == true) {
+                        switch (flags.Current().Value()) {
+                            case KeyMap::modifier::LEFTSHIFT:
+                            case KeyMap::modifier::RIGHTSHIFT:
+                            case KeyMap::modifier::LEFTALT:
+                            case KeyMap::modifier::RIGHTALT:
+                            case KeyMap::modifier::LEFTCTRL:
+                            case KeyMap::modifier::RIGHTCTRL:
+                                conversionInfo.Modifiers |= flags.Current().Value();
+                                break;
+                            default:
+                                ASSERT(false);
+                                break;
+                        }
+                    }
+
+                    // Do not allow same device code to point multiple input keys, it is not a real use-case
+                    if (_keyMap.find(code) == _keyMap.end()) {
+                        _keyMap.insert(std::pair<const uint32_t, const ConversionInfo>(code, conversionInfo));
+
+                        std::map<uint16_t, uint16_t>::iterator index = previousKeys.find(conversionInfo.Code);
+
+                        if (index != previousKeys.end()) {
+                            updatedKeys[index->first]++;
+                        } else {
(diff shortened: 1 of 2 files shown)


pwielders:
I do not fully get the issue. What is the usecase here ? If this is needed, I suggest that the Load, taking a file, also calls this method as the contents of the file is this array and now, I think we have duplicate information. 
Than the next step is probably to probably "hide" the JSON table, to avoid that people need to convert std::lists <> to JSON, and here back, so than I suggest a move operator of s std::list, with the data as it is used internally.

rjpdasilva:
@pwielders, we have a use case where we want to load a keymap directly from a `Core::JSON::ArrayType<KeyMapEntry>&` object., mainly because we have such keymaps embedded with other settings on JSON configuration files, so we can't just provide the file to `Load()`.

Regarding this:
> If this is needed, I suggest that the Load, taking a file, also calls this method as the contents of the file is this array and now, I think we have duplicate information.

Fully agree and our original patch was done that way, i.e., we've changed `Load()` to use the new `LoadFromArray()` to avoid logic duplication.
@jfifranca, I believe this PR is missing this change, please check.

rjpdasilva:
@pwielders,

Regarding this:
> Than the next step is probably to probably "hide" the JSON table, to avoid that people need to convert std::lists <> to JSON, and here back, so than I suggest a move operator of s std::list, with the data as it is used internally.

Can you please clarify?

pwielders:
> @pwielders,
> 
> Regarding this:
> 
> > Than the next step is probably to probably "hide" the JSON table, to avoid that people need to convert std::lists <> to JSON, and here back, so than I suggest a move operator of s std::list, with the data as it is used internally.
> 
> Can you please clarify?

I looked a bit more in the code and saw that the JSON specification (and conversion) is part of this class, just ignore my remark.

pwielders:
Did not thoroughly look at the code yet, but I can also image that we drop the reading from file completely (deprecate it) and only use the JSONArray<>. Than it is up to the RemoteControl/BluetoothRemoteControl to convert the file to this array. Keeping the loading/saving simple and with only one type.

jfifranca:
@rjpdasilva You are correct. This is not the full PR, since the changes made onto Load() method were not merged. I will fix this.

pwielders:
@jfifranca can we make this pull request than in such a way that the load and the save take a JSON::ArrayType<>, the original load and save take a file name, but lets mark them with the DEPRECATED keyword, than we will change our implementations to start using the load/save with the load and save with JSON::ArrayType<>, and if the build is error free agaion, we could remove the original load and free.
I think it is safe to keep the method names load/save for both as the parameter will discriminate between the two versions.
So my expectation would be to have the following signatures:
DEPRECATED uint32_t VirtualInput::KeyMap::Load(const string& fileName);
DEPRECATED uint32_t VirtualInput::KeyMap::Save(const string& fileName);

uint32_t VirtualInput::KeyMap::Load(const Core::JSON::ArrayType<KeyMapEntry>& mappingTable);
uint32_t VirtualInput::KeyMap::Save(Core::JSON::ArrayType<KeyMapEntry>& mappingTable);

Where the first two deprecated methods are wrappers. The Load is just getting the JSON::ArrayType from the file and calls the new load method that takes the JSON::ArrayType, it created from the file. The Save is first calling the new method to get the JSON::ArrayType and than stores it into the give file name.

pwielders (changes requested):
Could you please enhance it so it is symmetrical again (load and save) and deprecate the current methods so in due time the file functionality is no longer used. So suggest: (DEPRECATED is already a preprocessor statement defined in Core/Portability.h)

DEPRECATED uint32_t VirtualInput::KeyMap::Load(const string& fileName);
DEPRECATED uint32_t VirtualInput::KeyMap::Save(const string& fileName);

uint32_t VirtualInput::KeyMap::Load(const Core::JSON::ArrayType<KeyMapEntry>& mappingTable);
uint32_t VirtualInput::KeyMap::Save(Core::JSON::ArrayType<KeyMapEntry>& mappingTable);

jfifranca:
@pwielders can you explain why we should deprecate the versions with file name? I think that having both APIs is a benefit. 
Agree with having symmetrical APIs, but having 'Save' with 'Core::JSON::ArrayType' parameter seems "strange" since it behaves like an "export" kind of method. With that said, my suggestion is to rename to 'Import' (from 'LoadFromArray') and 'Export'.
What is your opinion?

pwielders:
@jfifranca : Agree on the naming, it is subjective and for me Import/Export would be fine as well and actually I like them even better. I leave it up to your imagination (Just think that putting a parameter type in a method name, LoadArray/ImportArray is not usefull).
On deprecation. By deprecating the file, we force the developpers to take the loading of the file in there own hands. In the end it is just adding two lines, Open a file -> Call JSONFromFile(). As this used to be the only way, it was good to have it here. Now since we are also offering another method, I would like to get rid of that code for maintenance reasons and for sending a simpler message to the users. Users will ask me what is the difference between the load and the import. If one method is deprecated they know they do not have to ask it :-) 
Furthermore, I like thin interfaces (thin interfaces are easy to grasp, avoid GOD interfaces). These two methods (load/save) are no longer needed, so can be removed to make the interface thinner.
For backwards compatibility, we can not remove the methods but at least we can trigger during the build, with the depreacted keyword that there is a better alternative that serves more purposes. Suppose you have a config file that has more JSON information in there, not only the keytabel but also a section that describes the remote. For this you can not use the file load method anymore as well,  since it assumes that it contains only the Array. Whihc is quire some limitation for this interface.
So for all these reasons I suggest to mark them deprecated (not remove them yet). We will run a build and fix the two places and if it is warning free, or in a next major release, we will really remove the methods. Keeping the code slim and simple (KISS!)

jfifranca:
Closed and created a new pull request with requested changes on #804

PR #748, opened 2021-10-01
Bluetooth Profile: moving code from BluetoothRemoteControl plugin

jfifranca:
After analyzing the code used in BluetoothRemoteControl plugin, it was concluded that:

  * Find
  * FindHandle
  * FindChraracteristic

methods are better suited to be part of the core bluetooth implementaion, since the implementation is generic for any Bluetooth related plugin.


sebaszm (approved):
These methods are trivial helpers, but agree that they are generic.

PR #753, opened 2021-10-05
Modified to create singletons in JSONRPC::LinkType class in Thunder (instead of in plugin)

ghost:


pwielders (changes requested):
This is a very bad hack :-) I think the plugin or the otherside is violating some reference count/cleanup rules. So lets please fix this in the proepr location.

PR #754, opened 2021-10-06
Get free memory from /proc/meminfo to get the correct values.

sramani-metro:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

### DIFF
--- a/Source/core/SystemInfo.cpp
+++ b/Source/core/SystemInfo.cpp
@@ -195,14 +195,17 @@ namespace Core {
         sysinfo(&info);
 
         m_uptime = info.uptime;
-        m_totalram = info.totalram;
         m_pageSize = static_cast<uint32_t>(sysconf(_SC_PAGESIZE));
-        m_freeram = info.freeram;
-        m_totalswap = info.totalswap;
-        m_freeswap = info.freeswap;
         m_cpuloadavg[0]=info.loads[0];
         m_cpuloadavg[1]=info.loads[1];
         m_cpuloadavg[2]=info.loads[2];
+
+        MemorySnapshot snapshot = TakeMemorySnapshot();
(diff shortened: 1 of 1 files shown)


Source/core/SystemInfo.cpp
```diff
@@ -195,14 +195,17 @@ namespace Core {
...
-        m_totalram = info.totalram;
         m_pageSize = static_cast<uint32_t>(sysconf(_SC_PAGESIZE));
-        m_freeram = info.freeram;
-        m_totalswap = info.totalswap;
-        m_freeswap = info.freeswap;
         m_cpuloadavg[0]=info.loads[0];
         m_cpuloadavg[1]=info.loads[1];
         m_cpuloadavg[2]=info.loads[2];
+
+        MemorySnapshot snapshot = TakeMemorySnapshot();
```
MFransen69, line 203: Hi Santhosh, the idea of the memory measurement here is that it is fast but less precise. The TakeMemorySnapshot was indeed added in our code base to be more precise but comes at a cost as it is more expensive (meaning requires more processing power).
Therefore the TakeMemorySnapshot is only used in our code to when the precision is really needed (e.g. ResourceMonitor)
So I propose to leave this like it is as here we prefer speed over precision. Do you agree?

MFransen69 (changes requested):

Source/core/SystemInfo.cpp
sramani-metro: Yes, as long as we dont use the free mem from sysinfo for any critical memory management. So, should i close this pull then?

MFransen69: Yes let's do that

PR #756, opened 2021-10-06
Onem 22659 check launch result before waiting

mikolaj-staworzynski-red:


Source/com/Communicator.h
```diff
@@ -840,10 +840,10 @@ namespace RPC {
                     _adminLock.Unlock();
 
                     // Start the process, and....
-                    result->Launch();
+                    uint32_t launchResult = result->Launch();
 
                     // wait for the announce message to be exchanged
-                    if (trigger.Lock(waitTime) == Core::ERROR_NONE) {
+                    if (launchResult == Core::ERROR_NONE && trigger.Lock(waitTime) == Core::ERROR_NONE) {
```
pwielders: Minor remark, maybe for the next delivery, will approve it for know, our coding rules state that you should not rely on everyone being familiar with operator precedence so please use brackets to ease reading. This line would be:
if ((launchResult == Core::ERROR_NONE) && (trigger.Lock(waitTime) == Core::ERROR_NONE)) {

pwielders (dismissed):

Source/com/Communicator.h
mikolaj-staworzynski-red: Updated with brackets

pwielders (approved):

PR #757, opened 2021-10-06
ONEM-22659: Synchronization in Release changed

mikolaj-staworzynski-red:
UnregisterProxy in Release is invoked after taking Administrator lock and
self synchronization object.

Without that we have race between UnknownProxyType::Relese
(that invokes destructor) and UnknownProxy::Relese
(invoked on connection closed for deadProxies objects in Communicator::Closed ).

It causes invalid lifetime of the UnknownProxy objects that are at first stage
manifested by the following logs:
* "Probably trying to delete a used CriticalSection" (with EBUSY error code
  on pthread_mutex_destroy) - operation on locked mutex
* "Probably creating a deadlock situation" (with EINVAL error code on
  pthread_mutex_lock) - operation on already destroyed mutex
* Probably does the calling thread not own this CriticalSection (with EINVAL
  error code on pthread_mutex_unlock) - operation on already destroyed mutex

Later in plugin activation/deactivation stress test it cause memory corruption
and random crashes.

### DIFF
--- a/Source/com/Administrator.cpp
+++ b/Source/com/Administrator.cpp
@@ -102,10 +102,8 @@ namespace RPC {
         proxy->Complete(response);
     }
 
-    void Administrator::UnregisterProxy(const ProxyStub::UnknownProxy& proxy)
+    void Administrator::UnregisterProxyLocked(const ProxyStub::UnknownProxy& proxy)
     {
-        _adminLock.Lock();
-
         ChannelMap::iterator index(_channelProxyMap.find(proxy.Channel().operator->()));
 
         if (index != _channelProxyMap.end()) {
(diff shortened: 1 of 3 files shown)


pwielders (commented):
I do recognize that on R3 we had a similar issue. It required some changes to the code, not present in R2. So I would love to run the test first on R3 to see if the isseu really still exists. The described scenario is now resolved by making sure that the last reference to the proxy is only "released" from 1 point (or at least synchronized from 1 point.
To do so the Invalidate call on the Proxy has been added. See: https://github.com/rdkcentral/Thunder/blob/77ebd7bdf82ffc5eeda301d51b1b7c51b1297296/Source/com/Administrator.cpp#L309-L313 and https://github.com/rdkcentral/Thunder/blob/master/Source/com/IUnknown.h#L158-L168.
No the proxy gets only removed once from the administration.

So for R3, we consider this PR not to be applicable.

pwielders:
Suggest to close this pull-request for R3 and first test if the issue is still observed in R3 as I think it is resolved (differently, using the Invalidate method)

PR #771, opened 2021-10-12
[Library] Fix memory leak - delete library handler when not referenced anymore

lukiwa:


lukiwa:
This was discovered during debugging some stuff on R2 - the leak is still present here. On the master, it is already resolved in another way.

PR #773, opened 2021-10-13
Adding documentation for tags used in tools.

Karthick-Somasundaresan:


sebaszm (changes requested):
Some remarks...
- Please add another column denoting the scope of each tag (e.g. file, class, method, method parameter, pod member, etc.)
- Please remove the redundant information (e.g. json only, stubgen only) as it's already present in the other columns
- Please be consistent with acronyms and keyword capitalisation (e.g. use JSON-RPC instead of jsonRPC or JSONRPC, enum instead of ENUM or Enum... etc.)
- Please remove @listener description, this is not implemented and will most likely never be
- Please remove the "???" comments, with someting like TBD / "To be documented" so it's always publishable
- Please refrain from using personal pronouns in technical documentation :)

MFransen69 (approved):

sebaszm (approved):
Thanks, let's take this in. We'll fill the missing pieces at a later point.

PR #779, opened 2021-10-14
[METROL-373]  Fix crash during library unload

lukiwa:
https://jira.rdkcentral.com/jira/browse/METROL-373
This is a cherry-pick of https://github.com/rdkcentral/Thunder/commit/5adfed41fe49959d94abb0c6b43f97b03da35c0d

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -159,17 +159,14 @@ namespace PluginHost
     };
 
     void Server::WorkerPoolImplementation::Dispatcher::Dispatch(Core::IDispatch* job) /* override */ {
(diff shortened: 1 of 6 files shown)


pwielders (approved):

pwielders (approved):
@gladish / @vijs the issues encountered @Sky where no issues on R3. @lukiwa is trying to back-port the fix that actually resolves it but it is getting harder and harder to back-port. Also here a lot has been changed. I can no longer oversee if this will break anything else. @lukiwa tested it as good as possible but we can not guarantee anything and we strongly advice you to move to R3.

This is a risky commit. Please properly QA it in your environment !

karuna2git:
As per the discussion with Gurdal, This change is submitted to R2 as patch and sanity tested by Steven
This can be merged and given it for full regression tested.

PR #793, opened 2021-10-20
JSONRPC Message element leak fix

sramani-metro:
The result of the Job::Process() is assigned to _element without releasing the reference it is holding.

### DIFF
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -2376,6 +2376,7 @@ namespace PluginHost {
                 {
                     ASSERT(Job::HasService() == true);
                     ASSERT(_element.IsValid() == true);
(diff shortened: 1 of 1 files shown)


pwielders (commented):
@sramani-metro does this resolve the leak in the MemoryGraph testcase? If so, I think there is a deeper problem, since the additiona you do here should make no difference, definitly not if the "last" statement was (and still is) _element.Release().

pwielders:
@sramani-metro I think this became a duplicate as here: https://github.com/rdkcentral/Thunder/pull/798 the rootcause seems to be resolved. Had to merge it since I needed it :-)

sramani-metro:
This turned out to be a workaround rather than a fix which simply avoids the buggy usage of ProxyType assignment. Will close this.

PR #798, opened 2021-10-23
[REFERENCELEAK] The new ProxyTypes did not release an existing refere…

pwielders:
…nce if the asignment operator overload was triggered.

This causes a small memory leak that could only be detected in case the memory graph got plottted. To prove the stability of the
proxypool, the WPEFramework application can now print the elements created in the factories as used by the WPEFramework plugin
server (press E) under continous load these should be stable.

### DIFF
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -601,6 +601,18 @@ namespace PluginHost {
                         }
                         break;
                     }
+                    case 'E': {
+                        uint32_t requests, responses, filebodies, jsonrequests;
+                        _dispatcher->Statistics(requests, responses, filebodies, jsonrequests);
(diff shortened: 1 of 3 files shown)


pwielders:
I think this pull request, makes the pull request: https://github.com/rdkcentral/Thunder/pull/793 unneeded since that pull requests avoids the use of the assignment operator where the objects are moved. This pull request also resolves the deeper issue. A bug in the ProxyType. @sramani-metro agree? If so can you close pull-request: https://github.com/rdkcentral/Thunder/pull/793

MFransen69 (approved):

sramani-metro:
Yes, this definitely looks like the root cause. Since the issue was only seen in the JSONRPC message flow when the memory graphs are used, I didnt suspect there could be a deeper issue in the ProxyType itself. I will close the other review. Thanks

PR #808, opened 2021-10-27
ProcessContainer-CleanUp

Vinodsathyaseelan:
- Reason for change :  Enhancement
	Current implementation of wpeframework code for managing container runtime is using a lot of unnecessary mixins.
	Initial plan was to re-design the container runtime.
        Since the current implementation is already in use by customers, it has been decided to do
	a code clean up and confirm the container runtime infrastructure works as expected.

- Solution :
	Clean up of container run time infrastructure.


Source/processcontainers/common/BaseAdministrator.h
```diff
@@ -25,15 +25,16 @@
 namespace WPEFramework {
 namespace ProcessContainers {
 
-    template <typename TContainer, typename Mixin> // IContainerAdministrator, Lockable Mixin
-    class BaseAdministrator : public Mixin {
+    template <typename CONTAINER>
```
MFransen69, line 28: Do we really still need the CONTAINER as a template? Did a quick scan but I think we only use methods from IContainer so could we just use the IContainer interface to make it easier? But perhaps I missed something :)

Source/processcontainers/common/NetworkInfoUnimplemented.h
```diff
@@ -79,24 +79,15 @@ namespace ProcessContainers {
             return 0;
```
MFransen69: Let's throw away this file completely and just return nullptr when NetworkInfo is not available for a Container (more like we do with com). (and adjust the plugin if needed so it deals with this case)

Source/processcontainers/implementations/AWCImplementation/AWCImplementation.h
```diff
@@ -83,7 +82,7 @@ namespace ProcessContainers {
         uint32_t _pid;
         int _runId;
         awc::awc_app_state_t _appState;
-        mutable Core::CriticalSection _adminLock;
+        //mutable Core::CriticalSection _adminLock;  /*Unused member*/
```
MFransen69: why not just delete it?

MFransen69 (changes requested):
Let's also remove the c interface (and the proxy implementation of that). There seems no real reason to have it, we now have all the major implementations without it :) ) 
It is only confusing to see two interfaces and if we ever needed we can raise it from the dead using the history of this repo

Source/processcontainers/common/BaseAdministrator.h
Vinodsathyaseelan: This is to ensure strict type checking even though the base class is IContainer.  Eg: Prevent adding LXC Container to CRunContainterAdmistrator.

Source/processcontainers/common/NetworkInfoUnimplemented.h
Vinodsathyaseelan: Done.

Source/processcontainers/implementations/AWCImplementation/AWCImplementation.h
Vinodsathyaseelan: Removed this commented statement.

MFransen69 (approved):

PR #811, opened 2021-10-29
Time: UT updates and code changes based on testing

HaseenaSainul:


Source/core/Time.cpp
```diff
@@ -44,40 +44,136 @@ namespace Core {
...
+            day = (6);
+            break;
         default:
-            return (static_cast<uint8_t>(~0));
+            break;
+        }
+        return (day);
+    }
+
+    static uint32_t WeekFromString(string buffer, uint8_t delimeter, bool fromName, uint8_t& wday) {
```
MFransen69: Why not const string& buffer? (that would prevent copying the string)

MFransen69: And more in line with our coding guidelines I think const uint8_t delimeter, cont bool fromName would be better

Tests/unit/core/test_time.cpp
```diff
@@ -21,13 +21,55 @@
...
+    if (localTime) {
+        localtime_r(&rawTime, &convertedTime);
+    } else {
+        gmtime_r(&rawTime, &convertedTime);
+    }
+
+    uint32_t timeRead = convertedTime.tm_hour * 3600 + convertedTime.tm_min * 60 + convertedTime.tm_sec;
+    // NOTE: Currently new Time is calcualting only gmtoff with higher values, since for negative gmtoff case
+    // thunder giving same value, with out subtracting the time difference
+    uint32_t newTime = timeRead + ((convertedTime.tm_gmtoff > 0) ? convertedTime.tm_gmtoff: 0);
```
MFransen69: As far as I understand the documentation you should not use this field to actual compensate the struct tm  that was retreived, it is for informational purposes only (it is even a nonstandard field of struct tm) unless you are trying to return UTC time again but I do not think that is the case. Let's have a call tomorrow to discuss because I assume if you do not add to the result your return value differs from that from time.h, correct? It would be good see what the correct value should be in the testcases (e.g. take you local timezone and see of the result of this and thunder is correct or not. But let's call to discuss this.

Tests/unit/core/test_time.cpp
```diff
@@ -21,177 +21,1637 @@
...
 
 using namespace WPEFramework;
 using namespace WPEFramework::Core;
+static constexpr uint32_t MilliSecondsPerSecond = 1000;
+static constexpr uint32_t MicroSecondsPerMilliSecond = 1000;
+static constexpr uint32_t MicroSecondsPerSecond = MilliSecondsPerSecond * MicroSecondsPerMilliSecond;
 
-TEST(Core_Time, timedetails)
+
+int32_t GetMachineTimeDifference(string zone)
```
MFransen69: I understand it is only testcode so just as a note const string& is more performant when passing string (or other complex parameters in general) as then no copying of the content needs to be done

MFransen69 (changes requested):

Source/core/Time.cpp
HaseenaSainul: done

Tests/unit/core/test_time.cpp
HaseenaSainul: done

Source/core/Time.cpp
HaseenaSainul: done

Tests/unit/core/test_time.cpp
HaseenaSainul: yes updated TCs. Found it is working properly if I use Time(with isLocaltime as false) case getting proper values. So updated TCs

Source/core/Time.cpp
```diff
@@ -811,20 +892,20 @@ namespace Core {
         _ticks = (static_cast<uint64_t>(time.tv_sec) * MicroSecondsPerSecond) + (time.tv_nsec / NanoSecondsPerMicroSecond) + OffsetTicksForEpoch;
     }
 
-    Time Time::ToLocal() const {
+    Time::TimeAsLocal Time::ToLocal() const {
```
MFransen69: Could you please also do this for the Windows version. You do not have to test or even build it, we'll fix it in case it might now work :)

Source/core/Time.cpp
```diff
@@ -811,20 +892,20 @@ namespace Core {
...
-        return (Time(
+        return (TimeAsLocal(Time(
             static_cast<uint16_t>(local.tm_year + 1900),
             static_cast<uint8_t>(local.tm_mon + 1),
             static_cast<uint8_t>(local.tm_mday),
             static_cast<uint8_t>(local.tm_hour),
             static_cast<uint8_t>(local.tm_min),
             static_cast<uint8_t>(local.tm_sec),
-            0, false));
+            0, true)));
```
MFransen69: This change should be reverted, it should be passed as UTC to make sure the local time will internally stored in the returning type without being changed

Source/core/Time.cpp
```diff
@@ -835,7 +916,7 @@ namespace Core {
             static_cast<uint8_t>(_time.tm_hour),
             static_cast<uint8_t>(_time.tm_min),
             static_cast<uint8_t>(_time.tm_sec),
-            0, true));
+            0, false));
```
MFransen69: This should also be reverted to be the opposite of the above

Source/core/Time.h
```diff
@@ -26,7 +26,44 @@
 namespace WPEFramework {
 namespace Core {
     class EXTERNAL Time {
-    private:
+
+    public:
+        class TimeAsLocal {
```
MFransen69: If we make it a nested class perhaps better to remove the Time part of the name as that is already clear by the nesting, so I propose to name it AsLocal

Source/core/Time.h
```diff
@@ -26,7 +26,44 @@
...
+        private:
+            friend class Time;
+
+            TimeAsLocal(Time time)
```
MFransen69: Discussed with Pierre, because of his use case he needs an default constructor as well so having this one private with a friend does not add much value anymore. So I propose to make this one public and also add the default constructor
Also please pass it as a const& or if you want to make it even better you could also add a move constructor :)

Source/core/Time.h
```diff
@@ -26,7 +26,44 @@
...
+            {
+                return _time;
+            }
+            Time& LocalTime()
+            {
+                return _time;
+            }
+
+        private:
+            Time& _time;
```
MFransen69: Don't think this will work correctly. It is now a reference to a temporary. Better to just make it a non reference member

MFransen69 (changes requested):

Source/core/Time.cpp
HaseenaSainul: Moved this class out of class Time

Source/core/Time.cpp
HaseenaSainul: done

Source/core/Time.h
HaseenaSainul: Now it is not nested class, hence name not changed

Source/core/Time.h
HaseenaSainul: Added default constructor and moved TimeAsLocal(Time time) as per comment

Source/core/Time.h
HaseenaSainul: changed to non reference

Source/core/Time.cpp
HaseenaSainul: done

Source/core/Time.h
```diff
@@ -311,6 +311,38 @@ namespace Core {
         uint64_t _ticks;
 #endif
     };
+
+    class TimeAsLocal {
```
MFransen69: Think this needs to be external as well

MFransen69 (changes requested):

Source/core/Time.h
HaseenaSainul: done

MFransen69 (approved):

PR #812, opened 2021-10-29
Bluetooth: add SDP and AVDTP profiles

sebaszm:
- Introduce SDP Profile (client+server) and AVDTP profile (client), RTP packetizing
- Small fixes around BR/EDR connecting and PIN handling
- Rename Profile to GATTProfile


pwielders (commented):
Suggest building with Hidden symbols :-) Missing the EXTERNAL keyword on classes. This is what is applicable in the Yocto build maybe we should enable it by default as wel.. (https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L90 )

pwielders (approved):

Source/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ else()
     message(FATAL_ERROR "Invalid BUILD_TYPE: '${BUILD_TYPE}'")
 endif()
 
+set (HIDE_NON_EXTERNAL_SYMBOLS true)
```
pwielders: Think we need to remove this setting, but maybe good to, by default, already enable this in our builds so it aligns better with the latest Yocto builds and keeps the binaries smaller.

pwielders (changes requested):
Minor, you left the hidden flag on but might be good to change the default to "on" and only with an override you can make everything visible again.

sebaszm:
Drat, I didn't mean to commit that change, that was only for test... But the test proved useful, there was more that needed EXTERNAL than I originally anticipated. Let me remove it, it shouldn't be done this way anyway.

pwielders (approved):
Oke lets bring it in :-)

PR #813, opened 2021-11-01
Thunder: initscript install support selection flag added

HaseenaSainul:


404ag (approved):
LGTM

Source/WPEFramework/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@
 get_filename_component(TARGET ${CMAKE_CURRENT_SOURCE_DIR} NAME)
 
 set(THREADPOOL_COUNT "4" CACHE STRING "The number of threads in the thread pool")
+set(ENABLE_INITSCRIPT_SUPPORT "true" CACHE STRING "Enable init script(systemd/initd) install support from the thunder source tree")
```
404ag: Maybe it's better to use BOOL here with ON/OFF instead of true/false string

404ag (changes requested):

404ag (approved):
LGTM

PR #815, opened 2021-11-02
Build: Export CRITICAL_SECTION_LOCK_LOG compile setting

jfifranca:
While having the `DEADLOCK_DETECTION` CMake option turned on, an 'undefined symbol' error occurred when loading a shared library on a plugin.
Comparing the generated libraries, there was an obvious difference in the method signature. This is caused by the missing declaration of the `CRITICAL_SECTION_LOCK_LOG` compilation specifier, on the plugin side.

`CRITICAL_SECTION_LOCK_LOG` compilation specifier changes the implementation on CriticalSection, thus creating type misalignment when is not declared on both sides. Currently, it is not being exported.

### DIFF
--- a/cmake/modules/CompileSettings.cmake
+++ b/cmake/modules/CompileSettings.cmake
@@ -54,6 +54,10 @@ if(PERFORMANCE_MONITOR)
     target_compile_definitions(CompileSettings INTERFACE -DTHUNDER_PERFORMANCE=1)
 endif()
 
+if(DEADLOCK_DETECTION)
+    target_compile_definitions(CompileSettings INTERFACE CRITICAL_SECTION_LOCK_LOG)
+endif()
+
 if(NOT BUILD_TYPE)
     set(BUILD_TYPE Production)
     message(AUTHOR_WARNING "BUILD_TYPE not set, assuming '${BUILD_TYPE}'")


bramoosterhuis:
@jfifranca we already expose [```CRITICAL_SECTION_LOCK_LOG```](https://github.com/rdkcentral/Thunder/blob/R2/Source/core/CMakeLists.txt#L165). So as soon as your plugin is using the CMake modules supplied by Thunder (e.g. ```find_package(${NAMESPACE}Core MODULE REQUIRED)``` or ```find_package(${NAMESPACE}Plugins MODULE REQUIRED)```) the define should be set automagically because it's exposed PUBLIC on the build interface of Thunder's Core library.

In the R3 release [```CRITICAL_SECTION_LOCK_LOG```](https://github.com/rdkcentral/Thunder/blob/R2/Source/core/CMakeLists.txt#L165) got renamed to [```__CORE_CRITICAL_SECTION_LOG__```](https://github.com/rdkcentral/Thunder/blob/R3/Source/core/CMakeLists.txt#L173)

jfifranca:
Hi @bramoosterhuis! Thanks for your review comment. I've re-tested and it is ok. The problem was only on the plugin side, but tested with both changes (plugin and framework). 

Also, thanks for the heads-up on R3!  👍 

Will close the PR in accordance.

PR #819, opened 2021-11-11
WebTransfer: call HashLoading from Download start sequence instead de…

HaseenaSainul:
…serialization, since for larger file it endup read timedout or long hang on the worker thread

### DIFF
--- a/Source/websocket/WebSerializer.h
+++ b/Source/websocket/WebSerializer.h
@@ -123,7 +123,6 @@ namespace Web {
     protected:
         uint32_t Deserialize() override
         {
-            _hash.Reset();
             return (TextBody::Deserialize());
         }
         uint16_t Deserialize(const uint8_t stream[], const uint16_t maxLength) override
@@ -271,37 +270,12 @@ namespace Web {
         {
             return (_hash);
         }
-        uint16_t Serialize(uint8_t stream[], const uint16_t maxLength) const override
-        {
-            return Core::File::Read(stream, maxLength);
-        }
-
-    protected:
-        uint32_t Deserialize() override
+        inline void ResetHash()
         {
             _hash.Reset();
-            uint32_t result = FileBody::Deserialize();
-            if (result) {
-                LoadHash();
-            }
-
-            return (result);
         }
-        uint16_t Deserialize(const uint8_t stream[], const uint16_t maxLength) override
+        void LoadHash()
         {
-            uint16_t deserialized = FileBody::Deserialize(stream, maxLength);
-
-            // Also pass it through our hashing algorithm.
-            if (deserialized == maxLength) {
-                _hash.Input(stream, maxLength);
-            }
-
-            return deserialized;
-        }
-
-    private:
-        void LoadHash() {
-
             if ((static_cast<int64_t>(Core::File::Size() > 0)) && (Core::File::IsOpen() == true)) {
                 // Set file position to beginning
                 Core::File::Position(false, 0);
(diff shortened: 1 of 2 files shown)


Source/websocket/WebSerializer.h
```diff
@@ -322,6 +298,25 @@ namespace Web {
...
+            _hash.Reset();
+
+            return (FileBody::Deserialize());
+        }
+        uint16_t Deserialize(const uint8_t stream[], const uint16_t maxLength) override
+        {
+            uint16_t deserialized = FileBody::Deserialize(stream, maxLength);
+
+            // Also pass it through our hashing algorithm.
+            if (deserialized == maxLength) {
```
pwielders: I think you should always do this if (maxLength > 0). This is used to calculate the hash over the data that is comingin.

Source/websocket/WebTransfer.h
```diff
@@ -300,17 +306,26 @@ namespace Web {
 
             return (result);
         }
-        inline uint64_t FileSize() const {
+
+        inline void LoadFileInfo()
```
pwielders: Where are you referring to this new function ? What is the purpose of this ?

Source/websocket/WebSerializer.h
HaseenaSainul: Let me check, any issue with that case

Source/websocket/WebTransfer.h
```diff
@@ -377,8 +392,23 @@ namespace Web {
             }
         }
 
-        IS_MEMBER_AVAILABLE(Hash, hasHash);
+        IS_MEMBER_AVAILABLE(LoadHash, hasLoadHash);
```
pwielders, line 385: Why do we need these notifications? I assume it is debug code ?

pwielders (changes requested):
Why do we need these changes in Master? If I understand correctly the issue you are investigating is only on R2 and can not be reproduces on R3/master?

Source/websocket/WebTransfer.h
HaseenaSainul: Since Loading is taking a long time, so just wanted to inform the caller a progress or indication about it

Source/websocket/WebTransfer.h
HaseenaSainul: Since hash Loading for large file is taking a long time, so just wanted to inform the caller a progress or indication about it

Source/websocket/WebTransfer.h
HaseenaSainul: Sorry, this was added testing purpose, will remove

Source/websocket/WebSerializer.h
HaseenaSainul: used if(deserialized) for the check

Source/websocket/WebTransfer.h
HaseenaSainul: removed

Source/websocket/WebTransfer.h
HaseenaSainul: removed

pwielders:
@HaseenaSainul is this still applicable? Can we now reproduce this issue ? What was the rootcause ?

HaseenaSainul:
> @HaseenaSainul is this still applicable? Can we now reproduce this issue ? What was the rootcause ?

yes, it is applicable. Here if we do the resume, it start to calculate the hash value of already loaded file part and it cause a timeout for the subsequent fetches. Hence I moved it from Deserialisation to Download, to avoid delay in between deserialisation

HaseenaSainul:
Closing this, since working on another branch to handle the hash reload from the saved context. So will create new PR, once it is ready

PR #820, opened 2021-11-12
WebTransfer: move hash calculation to the caller thread context, to avoid deadlock with jsonrpc socket

HaseenaSainul:


HaseenaSainul:
Closing this now, will create later with required changes

PR #826, opened 2021-11-18
Fix desktop compilation issues for Thunder R2 branch

zrash:


pwielders:
@bramoosterhuis can you check if this oke and if it is oke, if we also need this on R3? I guess not as we did a lot of work in R3 for building this ion the PC.

PR #827, opened 2021-11-19
BCOM-5717: Wpeframework killed due to invalid memory access

arunvijas:
Reason for change: return statement added wrongly.
Test Procedure: Regression
Risks: None
Signed-off-by: Arun Vijay <asathy881@cable.comcast.com>

### DIFF
--- a/Source/core/Proxy.h
+++ b/Source/core/Proxy.h
@@ -588,6 +588,8 @@ namespace Core {
         {
             ASSERT(a_Index < m_Current);
             ASSERT(m_List != nullptr);
+            ASSERT(a_Entry.IsValid() != false);
+            ASSERT(m_List[a_Index] != nullptr);
 
             // Remember the item on the location, It should be a relaes and an add for
             // the new one, To optimize for speed, just copy the count.
@@ -595,6 +597,7 @@ namespace Core {
 
             // If it is taken out, release the reference that we took during the add
             m_List[a_Index]->Release();
+            m_List[a_Index] = nullptr;
 
             // Delete one element.
             Core::InterlockedDecrement(m_Current);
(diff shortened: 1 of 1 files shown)


pwielders (approved):

Source/core/Proxy.h
```diff
@@ -604,6 +607,10 @@ namespace Core {
                 // Kill the entry, (again dirty but quick).
                 memmove(&(m_List[a_Index]), &(m_List[a_Index + 1]), (m_Current - a_Index) * sizeof(IReferenceCounted*));
             }
+	    else if (a_Index == m_Current)
```
pwielders: Do not think this is needed. this is already done on line 600

Source/core/Proxy.h
```diff
@@ -614,9 +621,12 @@ namespace Core {
         {
             ASSERT(a_Index < m_Current);
             ASSERT(m_List != nullptr);
+            ASSERT(a_Entry.IsValid() != false);
```
MFransen69: Think you didn't build this in debug :) a_Entry is not available here...

MFransen69 (changes requested):

Source/core/Proxy.h
```diff
@@ -588,13 +588,16 @@ namespace Core {
...
```
MFransen69: Perhaps I'm missing something but I think this ASSERT is incorrect.
a_Entry is only assigned to below. E.g. I think here it would always be triggered now : https://github.com/rdkcentral/Thunder/blob/eadbd1e71ba415ed04247e9e3314b3ad2e49682b/Source/core/Proxy.h#L659

MFransen69 (changes requested):

karuna2git (approved):

PR #832, opened 2021-11-22
Abstract functionality for passing data over process boundaries

lukiwa:


Source/core/CyclicBuffer.cpp
```diff
@@ -265,7 +277,7 @@ namespace Core {
         AdminUnlock();
     }
 
-    uint32_t CyclicBuffer::Read(uint8_t buffer[], const uint32_t length)
+    uint32_t CyclicBuffer::Read(uint8_t buffer[], const uint32_t length, bool partialRead = false)
```
pwielders: I think this default parameter (for partialRead) should go inito the CyclicBuffer.h as otherwise the interface did change :-)

Source/core/DataElementFile.cpp
```diff
@@ -46,6 +46,8 @@ namespace Core {
         // The file needs to be prepared in the same way as we request the Memorymapped file...
         ASSERT(m_File.IsOpen() == false);
 
+        m_File.Open((type & File::USER_WRITE) == 0);
```
pwielders: This is changing the interface behaviour. It now opens the file, if it was not yet open. We need to eleborate a bit on this (probably open in the MessageDispatchr) to keep the default bahaviour like it was.

Source/core/MessageDispatcher.h
```diff
@@ -0,0 +1,422 @@
...
+                //did not even receive length of the full message
+                if (length < 2) {
+                    TRACE_L1("Inconsistent message\n");
+                    _dataBuffer.Flush();
+                } else if (length > outLength) {
+                    TRACE_L1("Lost part of the message\n");
+                    result = Core::ERROR_GENERAL;
+
+                    //first two bytes are sizeof full messsage.
+                    std::copy_n(outValue + sizeof(outLength), outLength - sizeof(outLength), outValue);
```
pwielders: Suggest to adapt this from the ReadSize() using the cursor from the Cyclic buffer perspective to avoid this "memcpy". Lets discuss this.

pwielders (changes requested):
Minor ones, but lets first discuss this in a conf call..

Source/core/DataElementFile.cpp
lukiwa: In this line ``` ASSERT(m_File.IsOpen() == false);``` an assertion is performed to check if file is not open. Then, the function in the next line ```if (IsValid()) ``` internally checks if the file is open. Shouldn't the file be opened between those checks?

Source/core/MessageDispatcher.h
```diff
@@ -0,0 +1,415 @@
...
+        * @param instanceId number of instance
+        * @return std::tuple<string, string, string> 
+        *         0 - doorBellFilename
+        *         1 - dataFileName
+        *         2 - metaDataFilename
+        */
+        static Filenames PrepareFilenames(const bool initialize, const string& baseDirectory, const string& identifier, const uint32_t instanceId)
+        {
+            if (initialize) {
+                if (Core::File(baseDirectory).IsDirectory()) {
```
pwielders: I do not htink you can do this. This is only allowed if the "first" MessageDispatcher is build, but if the second on gets created (like we saw with the WPEProcess (WebServer out-of-process plugin) it will create another file.
Make the clearing of the directory a responsibility from a higher layer, e.g. the WPEframework application, priot to opening the first MessageDispatcher (PluginHost.cpp, opening the first TraceUnit should do this)

Source/core/MessageDispatcher.h
```diff
@@ -0,0 +1,415 @@
...
+                    //if directory exists remove it to clear data (eg. sockets) that can remain after previous creation
+                    Core::Directory(baseDirectory.c_str()).Destroy(false);
+                }
+
+                if (!Core::Directory(baseDirectory.c_str()).CreatePath()) {
```
pwielders: Same for creating the Directory, leave this up to the Higher Layer. Suggest to ASSERT here if the Directory does not exist :-)

pwielders (approved):

PR #840, opened 2021-11-26
[FRAME] Extend the functionality of the Frame template.

pwielders:
Sometimes the ordering in a binary serialzed frame is LSB, sometimes it is MSB. Allow the template to
cope with both situations, the default is MSB as that was as it was used sofar.
Also sometimes we only need to work with very small frames <255 bytes. using the the uint16_t for maintaining
location and size is a bit of overkill. Maybe in the future we would like to enlarge the 64Kb boundary to
4Gb, and the variables than need to be uint32_t :-)
Make the variable sizes for creating the buffers a template parameter, again making the default equal to
what it was.

### DIFF
--- a/Source/core/Frame.h
+++ b/Source/core/Frame.h
@@ -343,9 +343,11 @@ namespace Core {
         ~FrameType() = default;
 
         FrameType<BLOCKSIZE, BIG_ENDIAN_ORDERING, SIZE_CONTEXT>& operator=(const FrameType<BLOCKSIZE, BIG_ENDIAN_ORDERING, SIZE_CONTEXT>& rhs) {
-            Size(rhs.Size());
-            if (Size() > 0) {
-                ::memcpy(&(_data[0]), rhs.Data(), Size());
+            if (this != &rhs) {
+                Size(rhs.Size());
+                if (Size() > 0) {
+                    ::memcpy(&(_data[0]), rhs.Data(), Size());
+                }
             }
             return(*this);
         }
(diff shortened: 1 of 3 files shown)


Source/core/Frame.h
```diff
@@ -303,59 +310,78 @@ namespace Core {
...
             , _data(buffer, length)
         {
             static_assert(BLOCKSIZE == 0, "This method can only be called if you pass a buffer that can not be extended");
         }
-        ~FrameType()
-        {
+        ~FrameType() = default;
+
+        FrameType<BLOCKSIZE, BIG_ENDIAN_ORDERING, SIZE_CONTEXT>& operator=(const FrameType<BLOCKSIZE, BIG_ENDIAN_ORDERING, SIZE_CONTEXT>& rhs) {
+            Size(rhs.Size());
```
MFransen69: Add check for assigning to self? (behaviour memcpy in that case is undefined)

Source/core/Frame.h
```diff
@@ -582,15 +657,15 @@ namespace Core {
...
         {
             TYPENAME result;
 
             if ((offset + sizeof(TYPENAME)) > _size) {
                 ::memset(&result, 0, sizeof(result));
-            } else
+            }
+            else if (BIG_ENDIAN_ORDERING == true) {
 #ifdef LITTLE_ENDIAN_PLATFORM
-            {
```
MFransen69, line 593: Didn't check very thoroughly but wouldn't it be better to put the endian handling in some helper method? Code is now duplicated it seems and an issue would need to be fixed in two places

MFransen69 (changes requested):

Source/core/Frame.h
pwielders: That was quicker to just copy/paste ;-) But created endian handling in separate methods,

Source/core/Frame.h
pwielders: added.

MFransen69 (approved):

PR #847, opened 2021-11-29
WebTransfer: Add support to get current hash context and reload it back

HaseenaSainul:


HaseenaSainul:
@mhughesacn, the implementation in Hash.cpp/Hash.h is open source code based, hence the License details is based on that.

HaseenaSainul:
Thanks @mhughesacn

Source/cryptalgo/Hash.h
```diff
@@ -96,11 +96,27 @@ namespace Crypto {
...
+            uint8_t messageBlock[64]; // 512-bit message blocks
+            uint32_t messageIndex; // Index into message block array
+        } Context;
+
     public:
         inline SHA1()
         {
             Reset();
         }
+        inline SHA1(const uint8_t context[])
```
pwielders: To avoid confusion, why not use the Context as the input parameter. This way the ContextLength() method becomes obsolete and the signature is much clearer, it expects a inline SHA1(const Context& context); and it is used to create a SHA1, from an "existing" context.
Update: As I continued the review there is also a load. Lets drop this method (constructor) and only add the Load(const Context& context); method

Source/cryptalgo/Hash.h
```diff
@@ -115,17 +131,33 @@ namespace Crypto {
         static const EnumHashType Type = HASH_SHA1;
         static const uint8_t Length = 20;
 
+        static uint32_t ContextLength()
```
pwielders: If you make the interface (constructor with the uint8_t) typesafe (take a SHA1::Context ) this method is not needed.

Source/cryptalgo/Hash.h
```diff
@@ -115,17 +131,33 @@ namespace Crypto {
...
+        {
+            return static_cast<uint32_t>(sizeof(_context));
+        }
+
+        inline const uint8_t* CurrentContext() const
```
pwielders: Why not return here the inline const Context& CurrentContex() const. Leave it up to the receiver on how to interpret and store this.

Source/cryptalgo/Hash.h
```diff
@@ -115,17 +131,33 @@ namespace Crypto {
...
+        {
+            return reinterpret_cast<const uint8_t*>(&_context);
+        }
+
+        inline void LoadContext(const uint8_t context[])
```
pwielders: As the constructore. make this type safe, you expect here a Context type, so set it on the interface like: inline void Load(const Context& contex); This way you also do not have to re-iteraet in the method name that it is a COntext as the parmater already depicts this. This makes it all type safe and more self-documenting.

pwielders (changes requested):
Also looked at some of the code, the result for SHA224 returns as the hash, the _block[0] variable but the SHA256, returns the _sha[0] variable. I think that is an error, not sure but could we create some unit tests for the SHA claculations? Also I think that the SHA256 does not need to use the Hash as it could also copy back the final hask into the block variable, effectively saving the hash object in the context and thus 20 bytes :-)

Source/cryptalgo/Hash.h
HaseenaSainul: done

Source/cryptalgo/Hash.h
HaseenaSainul: done

Source/cryptalgo/Hash.h
HaseenaSainul: done

Source/cryptalgo/Hash.h
HaseenaSainul: done

HaseenaSainul:
> Also looked at some of the code, the result for SHA224 returns as the hash, the _block[0] variable but the SHA256, returns the _sha[0] variable. I think that is an error, not sure but could we create some unit tests for the SHA claculations? Also I think that the SHA256 does not need to use the Hash as it could also copy back the final hask into the block variable, effectively saving the hash object in the context and thus 20 bytes :-)

Changed different hash based Context to common one

Source/cryptalgo/Hash.h
```diff
@@ -90,41 +90,40 @@ namespace Crypto {
         HASH_SHA384 = 48,
         HASH_SHA512 = 64
     };
+    typedef struct EXTERNAL Context {
+        uint64_t length;
+        uint64_t h[8];
+        uint32_t index;
+        uint32_t a, b, c, d;
```
pwielders: Can you not change the md5 from using a,b,c,d to start using the h[0], h[1],h[2],h[3] ? Than the fields a,b,c,d can also be dropped from the struct.

Source/websocket/WebTransfer.h
```diff
@@ -308,9 +308,9 @@ namespace Web {
         {
             return (_fileBody.Position());
         }
-        inline void LoadHashContext(const uint8_t context[])
+        inline void LoadHash(Crypto::Context& context)
```
pwielders: You know I am a bif fan of const ;-) Can we make the context here a const parameter ?

pwielders (changes requested):
Looks good, happy to see 1 context, just some minor remarks..

Source/cryptalgo/Hash.h
HaseenaSainul: done

Source/websocket/WebTransfer.h
HaseenaSainul: done

Source/cryptalgo/Hash.h
```diff
@@ -360,6 +388,18 @@ namespace Crypto {
         void Reset();
         static const EnumHashType Type = HASH_SHA224;
         static const uint8_t Length = 28;
+
+        inline const Context& CurrentContext() const
+        {
+            return _context;
+        }
+
+        inline void Load(Context& context)
```
pwielders: missing const on the context parameter ?

pwielders (approved):

PR #854, opened 2021-12-06
[ONEM-23840] Support-for-serving-gzip-compressed-files

mikolaj-staworzynski-red:
*.gz files served from filesystem will have GZIP Content-Encoding set and Conntent-Type remains same.
bundle.js.gz from filesystem would have the following:
* Content-Type: application/javascript
* Content-Encoding: GZIP


Source/websocket/WebSerializer.cpp
```diff
@@ -359,8 +359,17 @@ namespace Web
         return (Core::EnumerateType<type>(value).Data());
     }
 
+    bool endsWithCaseInsensitive(const string& mainStr, const string& toMatch)
```
pwielders: Minor, but to keep code consistent, Methods are always PascalCase, so could you please name it:
EndsWithCaseInsensitive

mikolaj-staworzynski-red: Done

pwielders (approved):

PR #859, opened 2021-12-09
[HIDDEN] Make sure the MODULE_NAME becomes hidden agin and do expose …

pwielders:
…the MetaDataProperty interfaces on the module.


HaseenaSainul (approved):
LGTM

PR #870, opened 2021-12-17
Core::Time update to add boundary validation for Time Ctor and UT updates based on latest Time changes

HaseenaSainul:


Source/core/Time.cpp
```diff
@@ -639,19 +639,23 @@ namespace Core {
...
-        _time.wMinute = minute % 60;
-        _time.wSecond = second % 60;
-        _time.wMilliseconds = millisecond;
-        _time.wDayOfWeek = static_cast<WORD>(~0);
-
-        if (localTime) {
-            SYSTEMTIME convertedTime;
-            TzSpecificLocalTimeToSystemTime(nullptr, &_time, &convertedTime);
-            _time = convertedTime;
+        if ((year >= 1900) && ((month >= 1) && (month <= 12)) && ((day >= 1 ) && (day <= 31)) &&
```
MFransen69: Can we perhaps make this a bit smarter? Either use a system API to check if the date is valid or otherwise create a small algorithm which checks how many days there are in a month (and also take the leap years into account). There is for example already an is_leap_year method in the cpp file.
Also can we have hours > 23? (e.g. would 24:30:00 be a valid time?)
And not very important as in the Thunder use case older dates are not relevant but is there a technical reason to limit years > 1900? Don't know if you tested but did a quick search and for the years in struct tm I found "tm_year is a signed value, therefore years before 1900 may be represented.". For the Windows presentation I'm pretty sure it is supported

Source/core/Time.cpp
```diff
@@ -639,19 +639,23 @@ namespace Core {
...
+            _time.wMinute = minute % 60;
+            _time.wSecond = second % 60;
+            _time.wMilliseconds = millisecond;
+            _time.wDayOfWeek = static_cast<WORD>(~0);
+
+            if (localTime) {
+                SYSTEMTIME convertedTime;
+                TzSpecificLocalTimeToSystemTime(nullptr, &_time, &convertedTime);
+                _time = convertedTime;
+            }
```
MFransen69, line 663: as _time is not set initialized in the constructor initializer list should we do that or have an else here for the IsValid to work?

MFransen69 (changes requested):

Source/core/Time.cpp
```diff
@@ -968,21 +972,25 @@ namespace Core {
...
         struct tm source{};
 
-        source.tm_year = year - 1900;
-        source.tm_mon = month - 1;
-        source.tm_mday = day;
-        source.tm_hour = hour;
-        source.tm_min = minute;
-        source.tm_sec = second;
-        source.tm_isdst = -1; // make sure dst is calculated automatically
+        if ((year >= 1900) && ((month >= 1) && (month <= 12)) && ((day >= 1 ) && (day <= 31)) &&
```
MFransen69: And perhaps better not to copy these kind of algorithms but to make a helper for them so we only have the code once (could be of use later on somewhere else anyway)

MFransen69 (changes requested):

Source/core/Time.cpp
HaseenaSainul: @MFransen69, will add algorithm to validate days in a month case. 
24 will be set to 0, hence why added this case.
<<is there a technical reason to limit years > 1900>> according to this https://www.cplusplus.com/reference/ctime/tm/ it should be start from 1900 and we are subtracting 1900 from our code base too :source.tm_year = year - 1900;

MFransen69: To my opinion It is not about what (at least the lunix implementation) accepts but what seems logical. Times bigger than 23:59:59.999 do not make much sense or am I missing something?

Source/core/Time.cpp
HaseenaSainul: Initialise with some default values you mean ? Currently it is working with default values. But as you suggested, I think better to init with default to avoid garbage as well

Source/core/Time.cpp
HaseenaSainul: Sure, will use a method to handle this

Source/core/Time.cpp
HaseenaSainul: Sure, I will make the valuation accordingly, since in the windows we are doing %24, %60, %60, I added upto the max range. Will restrict the limit to 23:59:59.999

Source/core/Time.cpp
HaseenaSainul: Added method to handle this and called from both Windows and posix version

Source/core/Time.cpp
HaseenaSainul: I found initialisation of _time is already there in the code, expect with this windows based contractor. Added here as well

Source/core/Time.cpp
HaseenaSainul: Validation logic added for Time inputs as an API and called from both Windows and posix version

Source/core/Time.h
```diff
@@ -343,7 +358,17 @@ namespace Core {
         {
             return (static_cast<uint16_t>(time.tm_yday));
         }
+        inline bool IsLeapYear(const uint16_t year) const
```
MFransen69: There was already a static is_leap_year in the c++ file, please use that or remove that one. Now we have two implemenations (hope they are the same :) )

Source/core/Time.h
```diff
@@ -343,7 +358,17 @@ namespace Core {
...
+        {
+            return (((year % 4 == 0) && (year % 100 != 0)) || (year % 400 == 0));
+        }
+        inline bool IsValidTime(const uint16_t year, const uint8_t month, const uint8_t day, const uint8_t hour, const uint8_t minute, const uint8_t second, const uint16_t millisecond) const
```
MFransen69: Although it is naming and subjective I would either only let it check time and remove the date checks or call it IsValidDateTime

Source/core/Time.cpp
```diff
@@ -1173,5 +1171,33 @@ namespace Core {
         return (seconds);
     }
 
+    bool Time::IsValidDate(const uint16_t year, const uint8_t month, const uint8_t day) const
+    {
+        uint8_t totalDays = 0;
```
MFransen69, line 1176: Perhaps it would be easier (and a bit faster although it would be nice to check that :) ) to do a month%2 for selecting even/uneven months for number of days and then have one if in case month%2 == 0 for feb as the special case

MFransen69 (changes requested):

Source/core/Time.h
HaseenaSainul: Sorry I missed to push my last commit, I used the one available with CPP and renamed it

Source/core/Time.cpp
HaseenaSainul: For months 4,6,9 and 11 are 30, So that logic will not work :-)

Source/core/Time.h
HaseenaSainul: name changed to IsValidDateTime

Source/core/Time.h
```diff
@@ -343,7 +358,13 @@ namespace Core {
         {
             return (static_cast<uint16_t>(time.tm_yday));
         }
+        inline bool IsValidDateTime(const uint16_t year, const uint8_t month, const uint8_t day, const uint8_t hour, const uint8_t minute, const uint8_t second, const uint16_t millisecond) const
```
MFransen69: adding the inline here does not make much difference, better to add it to IsValidDate below

Source/core/Time.cpp
```diff
@@ -37,6 +37,11 @@ namespace {
 namespace WPEFramework {
 namespace Core {
 
+    static int IsLeapYear(const uint16_t year)
```
MFransen69: like we discussed

MFransen69 (changes requested):

Source/core/Time.cpp
HaseenaSainul: done

Source/core/Time.h
HaseenaSainul: removed from here, and added the inline in the definition of IsValidDate in CPP

MFransen69 (approved):

PR #871, opened 2021-12-17
WorkerPool: Make Submit() return bool instead of void for error handling

jfifranca:
Changing the API of IWorkerPool::JobType::Submit from

void Submit()

to

bool Submit()

so to reflect the success of the Submit operation towards ThreadPool.
Current implementation will trigger silent errors, where a failed submission
will pass unnoticed.

### DIFF
--- a/Source/core/WorkerPool.h
+++ b/Source/core/WorkerPool.h
@@ -47,13 +47,17 @@ namespace Core {
             }
 
         public:
-            void Submit()
+            bool Submit()
             {
+                bool result = false;
                 Core::ProxyType<Core::IDispatch> job(ThreadPool::JobType<IMPLEMENTATION>::Aquire());
 
                 if (job.IsValid()) {
                     Core::IWorkerPool::Instance().Submit(job);
+                    result = true;
                 }
+
+                return result;
             }
             bool Schedule(const Core::Time& time)
             {


pwielders:
What is the use case for this change? It is volatile information and one should not base any assumptions on hatis returned here. I can image that it is interesting for debugging/loggin purpose but that would be it.
Coulld you eleborate on what you will be using this for? If the job will not be submitted, it is because it is already submitted, so it is not an error. This construction will just make sure that the job gets scheduled for running only once. So it is not consodred an error if it is *not* submitted.

jfifranca:
The main drive for this change was for logging purposes. In the use case, it is categorized as a `Warning` if `Submit` returns `false`. 
Another reason is consistency when compared with the `Schedule` method.

rjpdasilva:
Hi @pwielders, I think the argument on making this consistent wiith `Schedule()` is a strong one.
On the other hand, even if just for diagnostics purposes, do you see any disadvantage of making this change?

pwielders:
@rjpdasilva and @jfifranca nope, consitency and than use it only for logging is s a valid reason. Just want to make sure that there will be no functional logic based on it. We resolved a nasty race condition a couple of weeks ago where the LastRef() on the ProxyObject was used. Method is now removed ;-) 
Lets merge this than into the R2, On R3, I will make sure that it gets aligned see next remark below.
@jfifranca / @rjpdasilva also be aware that we refactored the Workerpool last week since there was a possibility that the job (with the old implementation could run multiple times concurrently. Resolvedn in R2.

pwielders (approved):

karuna2git (approved):

karuna2git:
@pwielders ,
You mentioned in the comment that there is a change made to master to address the race condition.
`there will be no functional logic based on it. We resolved a nasty race condition a couple of weeks ago where the LastRef() on the ProxyObject was used. Method is now removed ;-)`

Is that already brought to R2?

PR #872, opened 2021-12-17
ThreadPool: Fix missing virtual declaration to JobType destructor

jfifranca:
Adding virtual declaration to the destructor of Core::ThreadPool::JobType class.
This class serves as base class to Core::IWorkerPool::JobType, hence if one uses a pointer of the base class type, the correct destruction sequence will not be met.

### DIFF
--- a/Source/core/ThreadPool.h
+++ b/Source/core/ThreadPool.h
@@ -85,7 +85,7 @@ namespace Core {
             #pragma warning(default: 4355)
             #endif
 
-            ~JobType()
+            virtual ~JobType()
             {
                 ASSERT (_state == IDLE);
                 _job.CompositRelease();


MFransen69 (dismissed):

Source/core/ThreadPool.h
```diff
@@ -85,7 +85,7 @@ namespace Core {
             #pragma warning(default: 4355)
             #endif
 
-            ~JobType()
+            virtual ~JobType()
```
MFransen69, line 88: Hi,

I fully understand it is used as a base class for Core::IWorkerPool::JobType but I would not expect we do a delete somewhere on the base class (so on a JobType<IMPLEMENTATION> pointer). If we for semantic correctness add the virtual here all the jobs grow in size (as there are no virtuals in JobType yet). You could I guess argue if in this case the public inheritance from JobType is technically the best solution but at this moment we would not prefer to rewrite that.
So was this change created because there really is an issue and somewhere a delete on a JobTYpe<IMPLEMENTATION>* is done or just to make it semantically correct? In the later case we would like to propose to leave it like it is to not grow the size of the JobType objects unnecessarily.

MFransen69 (changes requested):

Source/core/ThreadPool.h
jfifranca: Hi @MFransen69! Yes, the main drive for the change was for semantic correctness (no issues detected).

jfifranca: Agree to drop this MR

MFransen69:
As agreed let's close this Pull Request and not merge this change.

PR #879, opened 2022-01-04
CyclicBuffer: Updates

HaseenaSainul:
1. Allow write if the requested size is less than free size, to avoid infinite loop
2. Updated AssureFreeSpace logic to avoid inifinite looping with free and required with same size condition
3. Set/Reset Overwritten flag once freeing the space for write (in overwrite enabled mode)
4. Add assert if buffer create with readonly mode
5. Check _administrator pointer also for buffer validity
6. Lock deadlock issue fixes

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -41,6 +41,7 @@ namespace Core {
         , _alert(false)
         , _administration(nullptr)
     {
+        ASSERT((mode & Core::File::USER_WRITE) != 0);
 #ifdef __WINDOWS__
         string strippedName(Core::File::PathName(_buffer.Name()) + Core::File::FileName(_buffer.Name()));
         _mutex = CreateSemaphore(nullptr, 1, 1, (strippedName + ".mutex").c_str());
@@ -95,7 +96,7 @@ namespace Core {
         uint32_t actual_offset = RoundUp(offset, sizeof(void*));
         uint32_t actual_bufferSize = 0;
         if (bufferSize == 0) {
-            actual_bufferSize = actual_offset <= static_cast<uint32_t>(_buffer.Size()) ? static_cast<uint32_t>(_buffer.Size() - actual_offset) : 0;
+            actual_bufferSize = actual_offset <= _buffer.Size() ? (_buffer.Size() - actual_offset) : 0;
         } else {
             actual_bufferSize = bufferSize <= _buffer.Size() ? bufferSize : 0;
         }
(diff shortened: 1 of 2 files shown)


Source/core/CyclicBuffer.cpp
```diff
@@ -506,14 +519,10 @@ namespace Core {
 
                 _administration->_agents++;
 
-                AdminUnlock();
-
-                timeLeft = SignalLock(timeLeft);
+                result = SignalLock(timeLeft);
```
pwielders, line 518: Removing the AdminUnlock here, just before the waiting for a new lock might easily lead to deadlocks.
Can we associate these  "lock" changes at lease with a test case that showed that it was failing?

Source/core/CyclicBuffer.cpp
```diff
@@ -430,22 +438,27 @@ namespace Core {
         uint32_t tail = oldTail & _administration->_tailIndexMask;
         uint32_t free = Free(_administration->_head, tail);
 
-        while (free <= required) {
-            uint32_t remaining = required - free;
-            Cursor cursor(*this, oldTail, remaining);
-            uint32_t offset = GetOverwriteSize(cursor);
-            ASSERT((offset + free) >= required);
+        if (free <= required) {
```
pwielders: The cyclic buffer implementation is based on optimistic concurrency, assume it works unless at the end when we check it, the numbers do not add up, than we repeat.
The free size, might by the time we reach the check point already be occupied again and thus free space can not be assured, the while loop will make sure we will try again untill we have what we expected. If the cursor (line 445) does not ensure enough free space, than that is the actual error.
So here I am interested which test case led to these changes.

Source/core/CyclicBuffer.h
```diff
@@ -175,7 +175,7 @@ namespace Core {
         }
         inline bool IsValid() const
         {
-            return (_buffer.IsValid());
+            return (_buffer.IsValid() && (_administration != nullptr));
```
pwielders: Can we not make the buffer invalid if there no administration _administration == nullptr ? Otherwise I guess we need more checks throughout the code if the administratin can be null but we do have a valid buffer..
Or, if we can not invalidate the buffer if there is ni _administration, only base our Valid object on the _administration.
I think it is good to have 1 source of thruth and not 2...

Source/core/CyclicBuffer.h
```diff
@@ -252,7 +252,7 @@ namespace Core {
         void AdminLock();
         void AdminUnlock();
         void Reevaluate();
-        uint32_t SignalLock(const uint32_t waitTime);
+        uint32_t SignalLock(uint32_t& waitTime);
```
pwielders, line 255: What is the reason for changing this interface ? Why do we need to return the amount of time we waited ?

pwielders (changes requested):
Changing the locking schema is a bit dangerous and I would like to see the related test cases. Also the while (in optimistic concurrency) to an if should be very well tested as it does not match the theory of optimistic concurrency. 
Can we split this PR in simple to test functional issues (added asserts, always good, length >= Free(), and, if there is a good reason for returning the duration from the SIgnalLock, feel free to add it) and then we do the changing of the locks in a different PR and the while -> if in a seperate PR, cuase I think we should definitily add some test cases around those..

HaseenaSainul:
Thanks @pwielders, for the comments, will look and update the details

HaseenaSainul:
Splitting this PR changes, hence closing this and will create new one with initial set changes : https://github.com/rdkcentral/Thunder/pull/890

PR #885, opened 2022-01-13
[JSONRPC] Extend the JSONRPC templates to get the context.

pwielders:
For extensions required by GAP, the JSONRPC call sequance should be enhanced to get access
to the internal administartion of the channel over which the JSONRPC request was issued.
This is called the call context. The call context can, with this pull request be requested
by providing a first parameter of the type Core::JSONRPC::Context, if this parameter is
supplied, this prameter will receive the actual Call context that goes with this JSONRPC
Request. This context contains the following information:

1) ChannelId, this is the channel by which the comminucation is established (TCP connection
   websocket or HTTP reques) with the thunder framework and thus can be used to send replys
   to.
2) SequenceId, this is the Id that was part of the message that arrived. It should be used
   the response so that the request can combine the response and the request.
3) Token, this is the payload that was part of the JWT token generated and passed when the
   connection was setup (for websocket) or attached to one of the HTTP headers in case of
   a HTTP JSONRPC request.

### DIFF
--- a/Source/core/JSONRPC.h
+++ b/Source/core/JSONRPC.h
@@ -289,32 +289,26 @@ namespace Core {
             Info Error;
         };
 
-        class EXTERNAL Connection {
-        private:
-            Connection() = delete;
-
+        class EXTERNAL Context {
         public:
-            Connection(const uint32_t channelId, const uint32_t sequence)
-                : _channelId(channelId)
-                , _sequence(sequence)
-            {
+            Context& operator=(const Context& rhs) = delete;
+
+            Context() 
+                : _channelId(~0)
+                , _sequence(~0)
+                , _token() {
             }
-            Connection(const Connection& copy)
+            Context(const Context& copy) 
                 : _channelId(copy._channelId)
                 , _sequence(copy._sequence)
-            {
+                , _token(copy._token) {
             }
-            ~Connection()
-            {
+            Context(const uint32_t channelId, const uint32_t sequence, const string& token)
+                : _channelId(channelId)
+                , _sequence(sequence)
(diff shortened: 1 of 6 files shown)


Source/core/JSONRPC.h
```diff
@@ -289,32 +289,26 @@ namespace Core {
...
-            {
+            Context& operator=(const Context& rhs) = delete;
+
+            Context() 
+                : _channelId(~0)
+                , _sequence(~0)
+                , _token() {
             }
-            Connection(const Connection& copy)
+            Context(const Context& copy)
```
MFransen69, line 301: Why not also add an operator=? It is always unexpected that you can only do one, or is there in this case a good reason?

Source/plugins/JSONRPC.h
```diff
@@ -29,11 +29,11 @@ namespace WPEFramework {
 namespace PluginHost {
 
     struct EXTERNAL IDispatcher : public virtual Core::IUnknown {
-        virtual ~IDispatcher() {}
+        virtual ~IDispatcher() override = default;
```
MFransen69, line 32: better to remove the virtual?

Source/plugins/JSONRPC.h
```diff
@@ -79,7 +79,13 @@ namespace PluginHost {
         };
 
     public:
-        typedef std::function<bool(const string& token, const string& method, const string& parameters)> TokenCheckFunction;
+        enum classification {
```
MFransen69, line 82: is there a reason not to make it a class enum?

MFransen69 (approved):

PR #886, opened 2022-01-13
SystemInfo: RawDeviceId creation: Ethernet card name configuration provision added

HaseenaSainul:
1. Moved RawDeviceId API from Core to WPEFramework, to limit its usage inside it. And try to ensure Network subsystem is up before trying to read MAC Ids
2. Added Ethernet Card name configuration


Source/WPEFramework/PluginHost.cpp
```diff
@@ -545,7 +545,7 @@ namespace PluginHost {
             SYSLOG(Logging::Startup, (_T(EXPAND_AND_QUOTE(APPLICATION_NAME))));
             SYSLOG(Logging::Startup, (_T("Starting time: %s"), Core::Time::Now().ToRFC1123(false).c_str()));
             SYSLOG(Logging::Startup, (_T("Process Id:    %d"), Core::ProcessInfo().Id()));
-            SYSLOG(Logging::Startup, (_T("SystemId:      %s"), Core::SystemInfo::Instance().Id(Core::SystemInfo::Instance().RawDeviceId(), ~0).c_str()));
+            SYSLOG(Logging::Startup, (_T("SystemId:      %s"), Core::SystemInfo::Instance().Id(SystemInfo::RawDeviceId(_config->EthernetCard()), ~0).c_str()));
```
pwielders: We should now get this information only through the ISubSystem->IIdentifier interface. The SystemInfo::RawDeviceId, should dissapear as a visible interface.

Source/WPEFramework/SystemInfo.h
```diff
@@ -373,8 +375,17 @@ namespace PluginHost {
                         _identifier->Release();
                     }
 
+                    uint32_t timeLeft = WaitTimeForInterfaceReady;
```
pwielders: Lets not wait for an interface to become available. If we need to do so we should do it in the PluginHost.cpp (where it is already waiting for the localhost interface) But if we do it here we delay the full startup of all the other plugins.

Source/WPEFramework/SystemInfo.h
```diff
@@ -27,11 +27,124 @@ namespace WPEFramework {
 namespace PluginHost {
 
     class SystemInfo : public PluginHost::ISubSystem {
-    private:
+    public:
         SystemInfo() = delete;
         SystemInfo(const SystemInfo&) = delete;
         SystemInfo& operator=(const SystemInfo&) = delete;
 
+    private:
```
pwielders, line 35: Looks like this is a copy of the core, why not just use the core functionality, why copy it here ?

pwielders (changes requested):

Source/plugins/IPlugin.h
```diff
@@ -223,6 +223,14 @@ namespace PluginHost {
         virtual ISecurity* Officer(const string& token) = 0;
     };
 
+    struct EXTERNAL IConfigure : virtual public Core::IUnknown {
```
pwielders: There is already an IConfigure you can use for the plugins, no need to define a new one that is part of the core. See interfaces/IConfigure..

Source/com/Ids.h
```diff
@@ -54,7 +55,7 @@ namespace RPC {
         ID_SUBSYSTEM_LOCATION = 0x00000036,
         ID_SUBSYSTEM_IDENTIFIER = 0x00000037,
         ID_SUBSYSTEM_TIME = 0x00000038,
-		ID_SUBSYSTEM_SECURITY = 0x00000039,
+        ID_SUBSYSTEM_SECURITY = 0x00000039,
```
pwielders, line 57: Already exists as an interface in the ThunderInterfaces, no need to make this part of the Thunder.

HaseenaSainul: Sure, will use that from interfaces

Source/plugins/IPlugin.h
HaseenaSainul: Sure will use that from interfaces

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: I think at this point system Identifier is not configured, so do we need to query for ISubSystem->IIdentifier

Source/WPEFramework/SystemInfo.h
HaseenaSainul: Moved this to DeviceIdentification plugin

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: Moved this after _dispatcher->Open as per our discussion

Source/WPEFramework/SystemInfo.h
HaseenaSainul: removed wait, as per the discussion now try for whatever ready in the deviceidentificaiton plugin not configured case.

Source/com/Ids.h
HaseenaSainul: done

Source/plugins/IPlugin.h
HaseenaSainul: done

Source/WPEFramework/PluginHost.cpp
```diff
@@ -573,6 +595,11 @@ namespace PluginHost {
             // If we have handlers open up the gates to analyze...
             _dispatcher->Open();
 
+            string id = GetDeviceId(_dispatcher->Services().SubSystemsInterface());
```
pwielders: We are now losing a reference. the returned SubSystemsInterface() is an Addref'ed interface, it is never Released, so it is currently leaking. Move this call (SubSystemsInterface()) into the GetDeviceId method so you can release the interface, if it is given..

pwielders (changes requested):
Interface leak on ISubSystem

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: done

pwielders (approved):

PR #887, opened 2022-01-14
[JSONRPC] Extend the JSONRPC templates to get the context. (#885)

pwielders:
For extensions required by GAP, the JSONRPC call sequance should be enhanced to get access
to the internal administartion of the channel over which the JSONRPC request was issued.
This is called the call context. The call context can, with this pull request be requested
by providing a first parameter of the type Core::JSONRPC::Context, if this parameter is
supplied, this prameter will receive the actual Call context that goes with this JSONRPC
Request. This context contains the following information:

1) ChannelId, this is the channel by which the comminucation is established (TCP connection
   websocket or HTTP reques) with the thunder framework and thus can be used to send replys
   to.
2) SequenceId, this is the Id that was part of the message that arrived. It should be used
   the response so that the request can combine the response and the request.
3) Token, this is the payload that was part of the JWT token generated and passed when the
   connection was setup (for websocket) or attached to one of the HTTP headers in case of
   a HTTP JSONRPC request.


gladish (approved):

karuna2git (approved):

scthunderbolt:

PR #889, opened 2022-01-17
DecoupledJob: Move DecoupledJob class to WorkerPool.h

jfifranca:
`DecoupledJob` class is better suited to be placed as a common component in the framework core. It will also facilitate maintenance since the implementation is copied throughout several packages.
In the process of moving it from Bluetooth plugins to `WorkerPool.h` file, a fix is introduced that will facilitate the destruction process of an object of this type. Original implementation has a caveat when a Job is added to the `ThreadPool` `MessageQueue` and the submitter process is destroyed/deactivated in-between. For this use case, if `Revoke` is not exercised, it can lead to misbehaviors.

With this new implementation :
 * the destructor was modified in order to call `Revoke` method instead of being
   left as `default`.
 * `Submit` method return code was changed from void to bool so that the caller
   can know whether the requested job was successfully submitted or not.

### DIFF
--- a/Source/core/WorkerPool.h
+++ b/Source/core/WorkerPool.h
@@ -258,5 +258,73 @@ namespace Core {
         mutable Metadata _metadata;
         ::ThreadId _joined;
     };
+
+    class DecoupledJob : private Core::ThreadPool::JobType<DecoupledJob&> {
+    public:
+        using Job = std::function<void()>;
+
+        DecoupledJob(const DecoupledJob&) = delete;
+        DecoupledJob& operator=(const DecoupledJob&) = delete;
+
+        DecoupledJob()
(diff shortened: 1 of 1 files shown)


pwielders:
Agree to move this, originally started in Bluetooth plugin, to core. However some issues have been resolved on R3 (multiple submits of this in certain scenarios) and @sebaszm solved this for the Bluetooth plugins. Hence why I have added him as a reviewer to see if those issues are in here as well. I will raise a Jira ticket to do the same in R3.

rjpdasilva:
Hi @pwielders, @sebaszm,

> and @sebaszm solved this for the Bluetooth plugins

Can you share the link to that commit?

karuna2git (approved):

PR #891, opened 2022-01-21
Dev/connector lib

sramani-metro:


pwielders (dismissed):

Source/connector/Connector.h
```diff
@@ -0,0 +1,59 @@
...
+
+#pragma once
+
+namespace WPEFramework {
+
+namespace PluginHost {
+    struct IShell;
+}
+
+namespace Connector {
```
pwielders: Although technically there is nothing wrong with this, I would like to hide it from the external interface and force all users to the C functions, to make sure that there is only one way for accessing for these methods. Less exposure is less maintenance and less portability risk. 

So lets hide this part: https://github.com/rdkcentral/Thunder/pull/891/files#diff-276434645d979f1b866322ebbce0d324c54b0a94e89821f4f50afe7d77107b1cR28-R51 and only allow the C functions to be accessed. This class off cource can be sused interbnally to this implementation thats fine, so it is just to move the class to the cpp file...

Source/connector/Connector.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "Module.h"
+#include "Connector.h"
+#include "plugins/IShell.h"
+
+namespace WPEFramework {
+namespace Connector {
+
```
pwielders, line 25: Move the class declaration in here

Source/WPEFramework/Controller.cpp
```diff
@@ -70,6 +71,9 @@ namespace Plugin {
 
         _resumes.clear();
         _service = service;
+        
+        Connector::Connector::Instance().Announce(service);
```
pwielders: Use the C function i.s.o. the C++ functionality, so remark for this class declaration.

Source/WPEFramework/Controller.cpp
```diff
@@ -110,6 +114,8 @@ namespace Plugin {
     {
         ASSERT(_service == service);
 
+        Connector::Connector::Instance().Revoke(service);
```
pwielders: Use the C function i.s.o. the C++ functionality, so remark for this class declaration.

Source/plugins/Types.h
```diff
@@ -267,13 +271,13 @@ namespace RPC {
         {
             ASSERT(_controller == nullptr);
 
+            _controller = Connector::Connector::Instance().Controller();
```
pwielders: Use the C function i.s.o. the C++ functionality, so remark for this class declaration.

pwielders (changes requested):

Source/connector/Connector.cpp
sramani-metro: Done

Source/WPEFramework/Controller.cpp
sramani-metro: Done

Source/plugins/Types.h
sramani-metro: Done

Source/connector/Connector.h
sramani-metro: Done

pwielders (approved):

PR #894, opened 2022-01-27
PluginServer: Release conneciton id if activation sequence is failed

HaseenaSainul:


pwielders (changes requested):
The root-casue here is in the plugin. The plugin has toresolve this, not the core. There mght be more connections that need to be termonated, only the plugin should know this and resolve this. 
We need to fix this in the plugin that is causing the issue.

HaseenaSainul:
Closing this, since I found there is an issue with AVS plugin sequence as well. After handling the segfault in the plugin, this ASSERT is not observed

PR #897, opened 2022-02-01
[Tracing][Logging] Use common unit to manage controls and push messages

lukiwa:
Introduce MessageUnit which, similarly to the TraceUnit,  is supposed to push messages using MessageDispatcher and manage Tracing and Logging controls. Tracing and logging controls are now instantiation of the generic Control class. 

These changes allow outputting Logging messages to a buffer, from where they can be popped in a plugin and then outputted to WebSocket or UDP socket. Set trace categories will not reset, after deinitializing and re-initializing a plugin. Documentation and unit tests were added. 

In the future, WarningReporitng should also use this concept, instead of creating its own Unit.

Related changes:
- https://github.com/rdkcentral/ThunderUI/pull/62
- https://github.com/rdkcentral/ThunderNanoServices/pull/539
- https://github.com/rdkcentral/ThunderInterfaces/pull/123
- https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/pull/127
- https://github.com/WebPlatformForEmbedded/buildroot/pull/414


pwielders (approved):
I guess we did a lot of reviewing during the dvelopment of this, so lets just get it in tehre and start testing.. Tooo much to review now :-)

PR #901, opened 2022-02-11
[Tracing] Increase TracingBuffer size to avoid discarding messages

lukiwa:
The current buffer size is too small in real-world applications and can be increased, to fit more trace messages.


MFransen69 (dismissed):

Source/tracing/TraceUnit.h
```diff
@@ -35,7 +35,7 @@ namespace Trace {
     struct ITraceControl;
     struct ITrace;
 
-    constexpr uint32_t CyclicBufferSize = ((8 * 1024) - (sizeof(struct Core::CyclicBuffer::control))); /* 8Kb */
+    constexpr uint32_t CyclicBufferSize = ((16 * 1024) - (sizeof(struct Core::CyclicBuffer::control))); /* 8Kb */
```
pwielders: Maybe we should than also update the comment at the end of the line :-)

pwielders (approved):

PR #902, opened 2022-02-15
RemoteConnection: Release connection after activation failure case also

HaseenaSainul:


MFransen69 (dismissed):

Source/WPEFramework/PluginServer.h
```diff
@@ -1040,6 +1040,11 @@ namespace PluginHost {
                     // Could be that we can now drop the dynamic library...
                     Core::ServiceAdministrator::Instance().FlushLibraries();
                 }
+
```
pwielders: This should be released prior to the FlushLibraries, as potentially the flush might be ignored if the Connection is still holing a reference to anything in that librray. SO please move this to  line 1031

pwielders (approved):

PR #903, opened 2022-02-16
[JsonGen] Support including of revision in the interface source file link

sebaszm:
Use "sourcerevision" in xxxPlugin.json "info" section or --source-revision command line option to specify the commit id. If not specified will pick up the default value from generator configuration (currently set to "master") or "main" if not configured anywhere. 

This works if the source location is configured to use a revision. Currently the interface location is specified in the generator configurationas follows:
"https://github.com/rdkcentral/ThunderInterfaces/blob/{revision}/interfaces/{interfacefile}"
where the {revision} token will be replaced accordingly.
This can also be overriden by "sourcelocation" in xxxPlugin.json "info" section, or @sourcelocation tag, or "sourcelocation" entry in .json or --source-location command line option,


sourcelocation:
why ping me >:)

MFransen69 (approved):

PR #908, opened 2022-02-23
DELIA-54331: Don't set the receive buffer

npoltorapavlo:
Reason for change: TCP delays caused by the receive buffer
set after connect/accept to a value less than the localhost
MSS and the TCP window advertised on connect/accept.
Test Procedure: Check performance.
Risks: Low


karuna2git (commented):
Change looks good to me. Merged to R2

PR #911, opened 2022-03-02
[JsonGen] Allow JSON property getters to pass argument also via a pointer

sebaszm:
ProxyStubs also allow now passing interfaces as ISomething** instead of ISomething*&.


pwielders:
@sebaszm just to make sure, this is backwards compatible, right? Old Stuff (Get/Set definitions for JSONRPC in COMRPC header files) will still work, for new stuff we can now propose to use pointer i.s.o. references to make C++ development easier and JSONRPC generation is still in line.

sebaszm:
> @sebaszm just to make sure, this is backwards compatible, right? Old Stuff (Get/Set definitions for JSONRPC in COMRPC header files) will still work, for new stuff we can now propose to use pointer i.s.o. references to make C++ development easier and JSONRPC generation is still in line.

Yes, it's fully backwards compatible. No need to modify the existing interfaces.

pwielders (approved):

PR #913, opened 2022-03-04
[PATH] Normalize path on a file basis, not only on directory basis.

pwielders:


Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
     {
         Close();
     }
+
+    /* static */ string File::Normalize(const string& location, bool& valid)
+    {
+        string result(location);
+
+        valid = true;
```
MFransen69, line 64: Guess not very important but should empty be valid?

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                    result[index] = '/';
+                }
+#endif
+
+                if ((result[index] == '/') && (index >= 1) ) {
+                    if (result[index - 1] == '/') {
+                        if (index >= 2) {
+                            // We have a double slash, clear all till the beginning
+                           result = result.substr(index - 1);
+                           index = 1;
```
MFransen69, line 83: You almost got met here, I thought shouldn't it be 2? But indeed we now skip to the end and do ++ and start the next iteration at 2 :)

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                        }
+                        else if ((result[index - 2] == '.') && ((index == 2) || (result[index - 3] == '/'))) {
+                            if (index <= 3) {
+                                valid = false;
+                                result.clear();
+                            }
+                            else {
+                                // Seems like we are moving up a directory... execute that on the result... if we can...
+                                // there is data we can drop, drop it, drop it till the '/' is found
+                                uint32_t offset = (index <= 3 ? 0 : index - 4);
```
MFransen69: small improvement perhaps for readability: index <= 3 cannot happen here as you already (rightfully) decided above that that would be invalid as you would move out of the allowed path

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                                while ((offset > 0) && (result[offset] != '/')) {
+                                    offset--;
+                                }
+                                result.erase(offset, index - offset);
```
MFransen69, line 105: Think something might be wrong here in case you have something like "aaa/../bbbb" in that case if I'm right (doing this from my head I'm not a compiler/debugger :) ) the result would be "/bbbb". I understand that on any other then the first folder you want to keep the/ but when it is not present in the original I guess you do not want to introduce it (or perhaps it is required that is always should start with a / but the better check already at the start of the method)

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                                index = offset;
+                            }
+                        }
+                    }
+                }
+                index++;
+           }
+
+            // Also do not allow .. or /.. at the beginning or at the end.. not detected by the loop
```
MFransen69: Not completely correct, think you detect ../ and /../ already at line 94 (and think indeed the code below does not check that again)

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                   // We have a ".." and nothing more
+                   valid = false;
+                   result.clear();
+               }
+               else if ((result.length() >= 2) && (result[result.length() - 2] == '/')) {
+                   result = result.substr(0, result.length() - 2);
+               }
+               else if ((result.length() >= 3) && (result[result.length() - 2] == '.') && (result[result.length() - 3] == '/')) {
+                    // there is data we can drop, drop it, drop it till the '/' is found
+                    uint32_t offset = (result.length() <= 3 ? 0 : result.length() - 4);
```
MFransen69: I would leave it in but I think you even would not need to check for the <= 3 as that (/../) would have already been detected by the while loop above

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,90 @@ namespace Core {
...
+                    while ((offset > 0) && (result[offset] != '/')) {
+                        offset--;
+                    }
+                    result = result.substr(0, offset);
+                    valid = result.length() > 0;
```
MFransen69: you could argue if this is completely correct. Something like "aaa/.." would result in an empty string. Guess this has to do with my question at the beginning. If passing an empty string is valid guess to be consistent also 'aaa/.." should be valid but I guess I'm nitpicking here, who would pass such a value :)

Source/core/FileSystem.h
```diff
@@ -595,7 +597,22 @@ namespace Core {
...
+            if (location.empty() == false) {
+
+                bool valid;
+                string result = File::Normalize(location, valid);
+
+                if ((valid == true) && ((result.empty() == true) || (result[result.length() - 1] != '/'))) {
+                    result += '/';
+                }
+
+                return (result);
```
MFransen69: understand why you did this but it is against the coding guidelines :)
Think an easy way to correct this and not copying location into result at the beginning is just to do "string result;" at the beginning and then always return result. In case location is empty the if will not be entered and you will now return an empty result like you would have done for location. In this case it will only cost the default (empty) string constructor which should be very cheap

MFransen69 (dismissed):

Source/core/FileSystem.cpp
pwielders: Yes, it would indicate the current active path.

Source/core/FileSystem.cpp
pwielders: oke.

Source/core/FileSystem.cpp
pwielders: Uhhh I guess you mean the comment is incorrect ?

Source/core/FileSystem.cpp
pwielders: oke

MFransen69 (approved):

PR #927, opened 2022-03-18
Compile warning: more flags added

HaseenaSainul:


cmake/common/CompileSettingsDebug.cmake
```diff
@@ -30,16 +30,16 @@ include(CMakePackageConfigHelpers)
 #
 if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_COMPILER_IS_GNUCXX} )
-        target_compile_options(CompileSettingsDebug INTERFACE -O0 -g -Wextra)
+        target_compile_options(CompileSettingsDebug INTERFACE -O0 -g)
```
MFransen69, line 33: Perhaps I'm missing something but should these not be in line with the options in release?

MFransen69 (changes requested):

cmake/common/CompileSettingsDebug.cmake
HaseenaSainul: This config file is specific to Debug mode

cmake/common/CompileSettings.cmake
```diff
@@ -101,11 +101,11 @@ endif()
 #
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     message(FATAL_ERROR "Compiling with Clang")
-    target_compile_options(CompileSettings INTERFACE -Weverything)
+    target_compile_options(CompileSettings INTERFACE -Weverything -Wextra -Wnon-virtual-dtor)
```
bramoosterhuis: What is our policy here? 

1. Enable warnings for all and make errors optional.
1. Make all warnings and errors optional.

Now there is a contradicting in this PR. In cmake/project.cmake.in you now made the warnings and errors optional. But adding warnings here, will force them on Core, so everyone gets these warnings, so having them optional in cmake/project.cmake.in will be useless. 

see:
https://github.com/rdkcentral/Thunder/blob/ae861aff03216a43907dd055592bb30ca226ec1e/Source/core/CMakeLists.txt#L207-L210

So you either add here the warnings here and remove them from the cmake/project.cmake.in file, or remove the warnings here and make all warnings and errors optional.

bramoosterhuis: @MFransen69 what is your view on this?

HaseenaSainul: Sorry, I missed to remove the part from CompileSettings.cmake

bramoosterhuis (approved):

MFransen69 (approved):

PR #930, opened 2022-03-21
[RDKDEV-262]: Changes related to DeviceIdentifcation plugin as part of sysncup 

HaseenaSainul:
1. SubSystem: Identification interface updates
2. RawDeviceId: implementation changed to make it based on given ethernet card name
3. Config option added for ethernet card name
4. Define MACSIZE constant and use it instead of magic number


pradeeptakdas:
Internal JIRA for tracking https://ccp.sys.comcast.net/browse/RDKCOM-2919

karuna2git (approved):

PR #933, opened 2022-03-24
Modified JSONRPC.h & JsonGenerator so as to make JSONRPC event code callable from const method.

ghost:


Source/plugins/JSONRPC.h
```diff
@@ -199,19 +199,19 @@ namespace PluginHost {
         //
         // Methods to send outbound event messages
         // ------------------------------------------------------------------------------------------------------------------------------
-        uint32_t Notify(const string& event)
+        uint32_t Notify(const string& event) const
         {
-            return (_handlers.front().Notify(event, Core::JSON::String()));
+            return (const_cast<JSONRPC*>(this)->_handlers.front().Notify(event, Core::JSON::String()));
```
MFransen69: Why not propagate the const further down instead of using the const cast here? I quickly checked (only briefly must admit) and perhaps a few iterators will be need to be replaced by const ones but I think potentially we could then do without const cast at all. At least let's try and let the compiler tell where a potential problem is :)

MFransen69 (changes requested):

Source/plugins/JSONRPC.h
ghost: @MFransen69 I have removed const_cast,

MFransen69 (approved):
C++ part looks good. Please wait with merging until Sebastian reviewed the python change

sebaszm (approved):
Looks good.

PR #938, opened 2022-03-30
IController changes syncup 

HaseenaSainul:


Source/WPEFramework/Controller.cpp
```diff
@@ -183,6 +183,132 @@ namespace Plugin {
...
+        // Normalise return code
+        if (result != Core::ERROR_NONE) {
+            result = Core::ERROR_GENERAL;
+        }
+
+        return result;
+
+    }
+
+    uint32_t Controller::Delete(const string& path)
```
MFransen69, line 213: I see that the delete code is now different then it was previously in endpoint_delete. For all the moved code let's just keep the implementation that was already in there (didnt check the others that were move3d, so please check these as well)

Source/WPEFramework/PluginServer.cpp
```diff
@@ -250,7 +250,10 @@ namespace PluginHost
 
         _adminLock.Unlock();
 
-        ASSERT(_notifiers.size() == 0);
+        TRACE_L1("Pending notifiers are %lu", _notifiers.size());
```
MFransen69: Let's also not add this code as it was not in R2 yet. I only added the unused variable in R3

Source/core/FileSystem.cpp
```diff
@@ -56,6 +56,99 @@ namespace Core {
     {
         Close();
     }
+
```
MFransen69, line 59: we should not add these changes to R2

Source/core/FileSystem.h
```diff
@@ -156,6 +156,7 @@ namespace Core {
             return (*this);
         }
 
+        static string Normalize(const string& input, bool& valid);
```
MFransen69: Let's also not add this

MFransen69 (changes requested):

Source/WPEFramework/PluginServer.cpp
HaseenaSainul: It is just a cleanup, hence added. will remove

Source/WPEFramework/Controller.cpp
HaseenaSainul: will check and update

Source/core/FileSystem.cpp
HaseenaSainul: removed

Source/core/FileSystem.h
HaseenaSainul: removed

Source/WPEFramework/Controller.cpp
HaseenaSainul: The update required for only Delete method, I removed Normalise call usage instead used older implementation. In Clone() method, the changes I added is just renaming variable to camel case

Source/core/FileSystem.cpp
HaseenaSainul: removed

Source/WPEFramework/PluginServer.cpp
HaseenaSainul: updated

MFransen69 (approved):

karuna2git (approved):

PR #941, opened 2022-04-04
LocalClosingWaitTime made configurable

HaseenaSainul:


Source/WPEFramework/GenericConfig.cmake
```diff
@@ -21,6 +21,8 @@ string(TOLOWER ${NAMESPACE} NAMESPACE_LIB)
 set(PORT 80 CACHE STRING "The port for the webinterface")
 set(BINDING "0.0.0.0" CACHE STRING "The binding interface")
 set(IDLE_TIME 180 CACHE STRING "Idle time")
+set(SOFT_KILL_CHECK 10  CACHE STRING "Soft kill check waiting time")
```
sramani-metro: May be adding _TIME at the end of both the configuration vars will help identify vars better?

Source/WPEFramework/Config.h
```diff
@@ -738,6 +746,12 @@ POP_WARNING()
             Core::SafeSyncType<Core::CriticalSection> scopedLock(_configLock);
             _idleTime = newValue;
         }
+        inline uint8_t SoftKillCheckWaitTime() {
```
MFransen69: please make both const methods

Source/WPEFramework/PluginServer.h
```diff
@@ -1725,12 +1725,13 @@ PUSH_WARNING(DISABLE_WARNING_THIS_IN_MEMBER_INITIALIZER_LIST)
...
                     config.AppPath(), 
                     config.ProxyStubPath(), 
-                    config.PostMortemPath(), 
+                    config.PostMortemPath(),
                     _engine)
                 , _server(server)
                 , _subSystems(this)
                 , _authenticationHandler(nullptr)
             {
+                _processAdministrator.ForcedDestructionTimes(config.SoftKillCheckWaitTime(), config.HardKillCheckWaitTime());
```
MFransen69, line 1734: Why not pass them on via the CommunicatorServer constructor?  And via that to the Communicator constructor

Source/com/Communicator.h
```diff
@@ -776,6 +784,11 @@ namespace RPC {
             }
 
         public:
+            void ForcedDestructionTimes(const uint8_t softKillCheckWaitTime, const uint8_t hardKillCheckWaitTime)
```
MFransen69: Perhaps I'm missing something but is this even called?

Source/com/Communicator.h
```diff
@@ -437,6 +437,10 @@ namespace RPC {
     };
 
     class EXTERNAL Communicator {
+    private:
+        static constexpr uint8_t SoftKillCheckWaitTime = 10; // In seconds
```
MFransen69, line 441: Think these are not used at the moment, right?
But perhaps it is a good alternative to passing these values around everywhere. We could make them static here (not constexpression) I guess and then set them (override basically then in that case) from the Communicator constructor. These values will never change, they are read from the config one and then never changed at runtime and they are the same for all usages, or am I missing something. And although they will never change anymore before even used, these are bytes even if they would change while being used that would be safe and their usage pattern (just single values) would not require any memory barrier. By having them static we do not have to copy these rather const values around to all classes/objects. They can just be used from a static method from the Communicator class everywhere else in this file.

MFransen69 (changes requested):

Source/WPEFramework/Config.h
HaseenaSainul: Sure will updated

Source/WPEFramework/PluginServer.h
HaseenaSainul: @pwielders, suggested to create method called ForcedDestructionTimes in Communicator class to set these values. Hence I used the method here

Source/com/Communicator.h
HaseenaSainul: It is used from PluginServer.h

Source/com/Communicator.h
HaseenaSainul: yes these are used as the default values to assign from the communicator constructor. 

Communicator::Communicator(const Core::NodeId& node, const string& proxyStubPath)
         : _connectionMap(*this)
         , _ipcServer(node, _connectionMap, proxyStubPath)
         , _softKillCheckWaitTime(SoftKillCheckWaitTime)
         , _hardKillCheckWaitTime(HardKillCheckWaitTime)

Source/com/Communicator.h
HaseenaSainul: Sorry it was redundant and removed

HaseenaSainul:
Closing this PR, since this has to be added on top of Pierre's changes in the Communicator. Will create new PR on top of it.

PR #942, opened 2022-04-06
Modified JSON.h to handle forward slash escape.

ghost:


karuna2git (approved):

npoltorapavlo:
@neeraj-deshpande I get "/" escaped in urls. I don't think it's expected behavior

PR #950, opened 2022-05-04
[WINDOWS] Get tracing in the new messaging setup working under window…

pwielders:
…s again.

Windows does not support Domain sockets so for windows it should be possible to
define a port number on the loopback interface (127.0.0.1) as a domain socket alternative.
This pull request adds the possibility (if the port is not 0) to use a loopback IP socket
for communication in stead of a domain socket. This allwos the messaging concept to also
work on a windows platform.


Source/core/MessageDispatcher.h
```diff
@@ -189,8 +189,8 @@ namespace Core {
          * @param initialize should dispatcher be initialzied. Should be done only once, on the server side
          * @param baseDirectory where to place all the necessary files. This directory should exist before creating this class.
```
lukiwa, line 190: Add socket port to docs

lukiwa, line 190: Add socket port to docs

Source/core/MessageDispatcher.h
```diff
@@ -385,13 +385,24 @@ namespace Core {
         *         1 - dataFileName
         *         2 - metaDataFilename
         */
-        static Filenames PrepareFilenames(const string& baseDirectory, const string& identifier, const uint32_t instanceId)
+        static Filenames PrepareFilenames(const string& baseDirectory, const string& identifier, const uint32_t instanceId, const uint16_t socketPort)
```
lukiwa, line 390: Also here, add socket port to the Doxygen docs.

lukiwa, line 390: Also here, add socket port to the Doxygen docs.

Source/core/MessageUnit.cpp
```diff
@@ -557,26 +557,28 @@ namespace Core {
         * @return uint32_t ERROR_NONE: opened sucessfully
         *                  ERROR_OPENING_FAILED failed to open
         */
-        uint32_t MessageUnit::Open(const string& pathName)
+        uint32_t MessageUnit::Open(const string& pathName, const uint16_t socketPort)
```
lukiwa, line 561: Also here, add to docs :)

lukiwa, line 561: Also here, add to docs :)

Source/messaging/MessageClient.cpp
```diff
@@ -27,9 +27,10 @@ namespace Messaging {
      * @param identifer identifier of the buffers
      * @param basePath where are those buffers located
      */
-    MessageClient::MessageClient(const string& identifer, const string& basePath)
+    MessageClient::MessageClient(const string& identifer, const string& basePath, const uint16_t socketPort)
```
lukiwa, line 31: And here, docs

lukiwa, line 31: And here, docs

lukiwa (changes requested):
It is OK, but since "socket port messaging" will not be used on Linux, maybe it would be good to somehow hide constructors and methods behind defines? Or create separate constructors allowing construction of the "messaging backend" in two distinct ways? Let me know what do you think

pwielders:
@lukiwa it could be used on linux as well however it is not recommended as the domain sockets are faster. So hiding it behind a Preprocessor might reduce code size but also hides a potential feature.
I am a bit in the middle. See advantages for left and see advantages for right. So your opinion is decisive. Where do you want to go?
And agree will update the documentation of Doxygen :-)

lukiwa:
Agree, maybe we can use this feature for something in the future. Personally, I would go for two constructors (or specific methods) - one for constructing domain socket (with argument basePath) and the second one with argument socketPort. 
Optionally, if the code size is an issue, maybe just set the default value for socketPort arguments as 0?

lukiwa (approved):

PR #953, opened 2022-05-13
RDKTV-15442, RDKTV-15390, DELIA-55298 : WPEFramework Crash in Core::ResourceMonitorType

nsahu910:
Reason for change: Call ResourceMonitor::Unregister explicitly during the socket close/destruction sequence
Risks: None
Signed-off-by: Neeraj Sahu <Neeraj_Sahu@comcast.com>

### DIFF
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -1129,6 +1129,7 @@ namespace Core {
             result = false;
         } else {
             DestroySocket(m_Socket);
+	    ResourceMonitor::Instance().Unregister(*this);
             // Remove socket descriptor for UNIX domain datagram socket.
             if ((m_LocalNode.Type() == NodeId::TYPE_DOMAIN) && 
                 ((m_SocketType == SocketPort::LISTEN) || (SocketMode() != SOCK_STREAM)) &&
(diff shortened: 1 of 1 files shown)


Source/core/SocketPort.cpp
```diff
@@ -373,7 +373,12 @@ namespace Core {
 
         // Make sure the socket is closed before you destruct. Otherwise
         // the virtuals might be called, which are destructed at this point !!!!
-        ASSERT(m_Socket == INVALID_SOCKET);
+        ASSERT((m_Socket == INVALID_SOCKET) &&(m_State == 0));
```
pwielders, line 376: Please see the comment above the ASSERT. Checking for m_State == 0 is not usefull as you already check that the m_Socket should be invalid and thus not usable, and thus m_state should not be checked/used.
Unregistering at this point is even dangerous and might lead to race conditions. If the socket is still open and the destructor is called, the VTABLE might be in a destructed mode and have dangling pointers (hence the ASSERT check).
This can be properly solved by making sure that the class inheriting and implementing the virtuals calls Close(Core::Infinite); in its destructor. This way the full close cycle is executed (and thus all implementations from the virtuals are called to signal a closure) before they are destructed.

nsahu910: but these changes are already available in master branch.
https://github.com/rdkcentral/Thunder/commit/149c241467fba1ca4ebc8d7c5345871feea64d86

PR #955, opened 2022-05-19
Warning: adding disable option for : non-virtual-dtor, along with updated windows part with empty defines

HaseenaSainul:


MFransen69 (approved):

pwielders:
The pull-request is oke, but where are these MACRO's used cause they sound scary to use...

HaseenaSainul:
this used in Spark: https://github.com/rdkcentral/ThunderNanoServices/commit/e6afc3c45318828655fd5e1efb22204bff2d0ec5#diff-9ed2454efed0181256f469b6355601fcbf22e747593b57a6cb2de82ebb542b46R134. Since this class is from pxCore.

PR #959, opened 2022-05-23
[METROL-523]: Block Controller from serving too many files

Karthick-Somasundaresan:
Before serving the file we are normalizing the file to
webserver file path to avoid serving file beyond the
configured path.

### DIFF
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -606,7 +606,7 @@ namespace PluginHost {
                     result = _unavailableHandler;
                 } else if (IsWebServerRequest(request.Path) == true) {
                     result = IFactories::Instance().Response();
-                    FileToServe(request.Path, *result);
+                    FileToServe(request.Path, *result, false);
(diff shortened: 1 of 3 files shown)


pwielders (changes requested):
This is now done on a generic level. We now block serving any file that might be from an unsafe path. It is not for this method to decide such a limitation but should be done by the callee. COuld you extend the signature with a boolean, like: void Service::FileToService(const string&, Web::Response&, bool safePath); and than signal with the last paramter whether it should make sure it is a safe path or not. So if the safePath variable is true, check for if it moves over the WebPrefix() path and if so, do not service it but if the variable is false, allow for moving outside of the WebPrefix().
Here: https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/PluginServer.h#L609 you can than pass as a last parameter true. This way the callee can decide whether he want a "safe" path or not.

pwielders (approved):

pwielders (approved):

PR #962, opened 2022-05-25
XIONE-8644 : [Rack][Xione UK][CUSTOM Test] : AV struck in blue screen…

nauka-programowania-MA:
… during continuos channel changes

Reason for change: Map wasn't clear correctly. It clear resource but doesn't clear itself. Now it correclty dealocate nodes and clear.
Test Procedure: Test if ThunderCall work properly and response is parsed
correclty
Risks: MID

Signed-off-by: mateusz.adamski <mateusz.adamski2@sky.uk>

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -3198,13 +3198,7 @@ namespace Core {
 
             void Reset()
             {
-                JSONElementList::const_iterator index = _data.begin();
-
-                // As long as we did not find a set element, continue..
-                while (index != _data.end()) {
-                    index->second->Clear();
-                    index = _data.erase(index);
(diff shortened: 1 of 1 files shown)


pwielders (approved):

Source/core/JSON.h
```diff
@@ -3225,6 +3214,12 @@ namespace Core {
                 }
             }
 
+        protected:
+            void Reset()
```
anand-ky: can we keep Reset Public? This is going to affect clients/plugins that call into Reset().

mateusz-adamski-red: Sure

PR #965, opened 2022-05-26
Add support for launching containers using spec file

filipe-norte:


filipe-norte:
Cherry picked from master
https://github.com/rdkcentral/Thunder/pull/961

PR #969, opened 2022-05-30
[JSONRPC] The full monty, including all risks, but fixes all issues s…

pwielders:
…ofar found on R3, backported to R2.

Bigger change set but fixing all issues so far **not** discovered on R2. Mainly on unicode usage in JSON. But with this fixes new issues might occur. Suggest to test this, if it works take it in on R2, but QA this thoroughly. 

A lower risk for the issue at hand might be taken in, see https://github.com/rdkcentral/Thunder/pull/968 that solves the issue in https://ccp.sys.comcast.net/browse/DELIA-56244

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -85,9 +85,11 @@ namespace Core {
 
         struct EXTERNAL IElement {
 
-            static char NullTag[];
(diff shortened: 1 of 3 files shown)


scthunderbolt:

PR #971, opened 2022-05-31
RDK-35564: Support Semantic Versioning for RDKServices

mateusz-adamski-red:


pwielders:
There was a meeting with @anand-ky and we proposed a POC to be build to automatically include the versioning for Plugin and JSONRPC Interfaces from the builds (SERVICE_REGISTRATION and the contract.json file) for this we do need enhancements in Thunder. These enhancements can be found here: https://github.com/rdkcentral/Thunder/pull/979

PR #981, opened 2022-06-09
RDKTV-17127 : WPEFramework crash at std::terminate

nsahu910:
Reason for change: Setting unused pointer to null
Risks: None
Signed-off-by: Neeraj Sahu Neeraj_Sahu@comcast.com

### DIFF
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -183,8 +183,11 @@ namespace PluginHost {
         virtual uint32_t Worker() override
         {
 	    _adminLock.Lock();
-            CloseDown(_destructor);
+	    PluginHost::Server* destructor = _destructor;
+	    _destructor = nullptr;
 	    _adminLock.Unlock();
+		
+            CloseDown(destructor);
             Block();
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
If this happens it means that there is something more serious going on. 
Scary part is that it means that the destruction can occure from two sides (double destruction) if the _destructor on line 185 occurs just before the pointer is pushed to the stack (and not null), the Thread will destruct in CloseDown *AND* it gets destrcuted on line 178.
so to fix this properly, you should lock the passing on of the _destrcutor member variable and clear it within the locl on line 185

nsahu910:
thanks @pwielders ,
i have added lock before closeDown, please check.

pwielders:
Is there a reproduction scenario for the issue that this pull request would fix ?

pwielders:
@nsahu910 does this fix the crash ? Is it 100% reproducible ?

This looks like colleteral damage from a crashing YouTube application. I can not imagine, but please prove me wrong, that this patch fixes the YouTube crash.

nsahu910:
changes not required.

PR #982, opened 2022-06-09
[virtualinput]: prevent auto-releasing of key modifiers (i.e: shift,c…

modeveci:
…trl,alt)


pwielders (commented):
Whats the usecase here? What are we solving. By adding this patch, key presses like only pressing and releasing the shift are now being dropped. It means that modifier keys are now nolonger send. On windows, for example, I know applications are using the modifier keys as a feature (like Sticky keys)

modeveci:
To be able send for instance SHIFT + 2 and to generate @. Without this PR, it is not possible to send two keys (one modifier, one other) at the same time. 

Without this PR, it behaves like this:
1. Press SHIFT and hold
2. Press 2 (this step releases previous "Pressed Shift" and presses 2).

So we are not able to send two keys at same time, which we need for Modifiers.

For sticky keys, Press SHIFT and Release SHIFT will work.

BTW, _pressedCode means the code pressed in previous iteration, not the current one. The current one is "sendCode"

pwielders:
Uhhh that example with the @ sign was indeed the one we tackled with this setup :-) However if step 2 in your setup occurs without physically releasing the shift thats wrong. Let me review the PR again with this info.
Just to make sure what I expect to see as messages:
1) SHIFT PRESSED EVENT
2) CHARACTER 2 PRESSED EVENT
3) CHARACTER 2 RELEASED EVENT
4) SHIFT RELEASED EVENT

So it is the responsibility of the receiver to know that the 2 was pressed with the shift still pressed in a previous cycle, the receiver needs to maintain state with respect to the modifiers.

modeveci:
[Thu, 09 Jun 2022 16:47:03]:[RemoteControl.cpp:360] UnknownKey: Unknown: [DevInput:0000002A] state pressed, blocked.
[Thu, 09 Jun 2022 16:47:06]:[RemoteControl.cpp:360] UnknownKey: Unknown: [DevInput:00000003] state pressed, blocked.
[Thu, 09 Jun 2022 16:47:07]:[RemoteControl.cpp:360] UnknownKey: Unknown: [DevInput:00000003] state released, blocked.
[Thu, 09 Jun 2022 16:47:08]:[RemoteControl.cpp:360] UnknownKey: Unknown: [DevInput:0000002A] state released, blocked.

0x2A is SHIFT, 0x3 is character "2" 
These are the sequence of the events.

pwielders (approved):

PR #983, opened 2022-06-09
DELIA-56537: Fix for WPEWebProcess Crash in __pthread_cond_timedwait

ksanth850:
Reason for change: Based on clock_gettime returns allowing pthread_cond_timedwait block (Fix included ticket RDKTV-15393)
Fix above issue: Checking clock_gettime value and printing error if not success
Test Procedure: .
Risks: High
Signed-off-by: Kumar Santhanam <Kumar_Santhanam@comcast.com>

### DIFF
--- a/Source/core/Sync.cpp
+++ b/Source/core/Sync.cpp
@@ -897,25 +897,31 @@ namespace Core {
 
             // We are not busy Setting the flag, so we can check it.
             if (m_blCondition == false) {
-                struct timespec structTime;
+                struct timespec structTime = {0};
 
-                clock_gettime(CLOCK_MONOTONIC, &structTime);
-                structTime.tv_nsec += ((nTime % 1000) * 1000 * 1000); /* remainder, milliseconds to nanoseconds */
-                structTime.tv_sec += (nTime / 1000) + (structTime.tv_nsec / 1000000000); /* milliseconds to seconds */
-                structTime.tv_nsec = structTime.tv_nsec % 1000000000;
+                if (clock_gettime(CLOCK_MONOTONIC, &structTime) == 0)
+                {
+                    structTime.tv_nsec += ((nTime % 1000) * 1000 * 1000); /* remainder, milliseconds to nanoseconds */
+                    structTime.tv_sec += (nTime / 1000) + (structTime.tv_nsec / 1000000000); /* milliseconds to seconds */
+                    structTime.tv_nsec = structTime.tv_nsec % 1000000000;
 
-                do {
-                    // Oops it seems that we are not allowed to pass.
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
Where is this issue coming from? What is the associated ticket?
If you look at the clock_gettime documentation there are only 2 issues possible for this call to fail:
EFAULT => tp points outside the accessible address space.
EINVAL => The clk_id specified is not supported on this system.

The first one means that the stack (that is where tp, struct timespec structTime is created) is not in accessible address space, e.g. a stack overwrite, so this is a severe error you can not recover from this, suggest an ASSERT or abort.

The second one means that clock type CLOCK_MONOTONIC is not supported, 100% reproducible cause the clock is supported on the system or not. If it is not. Thunder will not run at all. So bail out as well (Abort)

So if this happens the most likely root cause is the stack overflow, which means someone is allocating to much on a stack and this is just collateral damage. Find the root cause and decrease stack usage or increase the thread stack size.

Here, in case of a failing "clock_gettime" do not printf but abort as there is no recovery possible from this scenario! This PR does not fix an issue it just cover up a potential issue that requires a fix it will cause instable systems (a jibbering rack :-) )

ksanth850:
Hi @pwielders,

Actual ticket: https://ccp.sys.comcast.net/browse/RDKTV-15393
[RACK][HISENSE_PANEL][Stress Test][Reboot Test] : WPEWebProcess Crash in __pthread_cond_timedwait

Callstack:
```
[ 00 ] 0xb0dcdaa6
[ 01 ] __pthread_cond_timedwait
[ 02 ] WPEFramework::Core::Event::Lock
[ 03 ] WPEFramework::JSONRPC::LinkType<WPEFramework::Core::JSON::IElement>::Send<std::__cxx11::basic_string<char> >
[ 04 ] WPEFramework::JSONRPC::LinkType<WPEFramework::Core::JSON::IElement>::Get<WPEFramework::Core::JSON::ArrayType<WPEFramework::PluginHost::MetaData::Service> >
[ 05 ] TTSThunderClient::Service::isActive
[ 06 ] TTSThunderClient::Service::initialize
[ 07 ] TTSThunderClient::TextToSpeechService::initialize
[ 08 ] TTS::TTSClientPrivateJsonRPC::TTSClientPrivateJsonRPC
```
Reason comeup with this logic there is no much trace information in the callstack.
So, we suspected clock_gettime() returns must be failure and going for pthread_cond_timedwait block.

Suppose, if we have checks in clock_gettime() returns we come to know problem in clock_gettime() part or not.

Instead of printf i will go for assert based on return value(EFAULT, EINVAL) checks

With Regards,
Kumar S.

PR #990, opened 2022-06-15
Development/user permission fixes

bramoosterhuis:


Source/WPEFramework/Config.h
```diff
@@ -499,7 +502,7 @@ namespace PluginHost {
             inline Core::ProcessInfo::scheduler Policy() const {
                 return(_policy);
             }
-            inline uint16_t UMask() const {
+            inline Core::OptionalType<uint16_t> UMask() const {
```
pwielders: can we make this a const Core::OptionalType<uint16_t>& UMask() const {} signature

pwielders (approved):

PR #996, opened 2022-06-15
DELIA-56519: WPEFramework - undef __DEBUG__ unless passed in BUILD_TYPE

zrash:
Reason for change: undef __DEBUG__ unless passed in BUILD_TYPE

Test Procedure: Build and verify.
Risks: Low
Signed-off-by:Zameerun Rasheed M S <zexpen@gmail.com>

### DIFF
--- a/Source/com/Administrator.cpp
+++ b/Source/com/Administrator.cpp
@@ -294,7 +294,9 @@ namespace RPC {
                     }
                 } while ((loop->Decrement()) && (result == Core::ERROR_NONE));
 
+#ifdef __DEBUG__
                 ASSERT (loop->Flushed() == true);
+#endif
 
                 loop++;
             }
--- a/Source/core/IIterator.h
+++ b/Source/core/IIterator.h
@@ -174,7 +174,7 @@ namespace Core {
 
         inline ELEMENT operator->()
         {
-            ASSERT(IsValid());
+            ASSERT(((m_Index > 0) && (m_Index <= Count())));
 
             return (*m_Iterator);
         }
--- a/Source/core/Portability.h
+++ b/Source/core/Portability.h
@@ -396,7 +396,6 @@ inline void EXTERNAL SleepS(unsigned int a_Time)
 #endif
 
 #if defined(_THUNDER_DEBUG) || !defined(_THUNDER_NDEBUG)
-#define __DEBUG__
 #ifdef _THUNDER_PRODUCTION
 #error "Production and Debug is not a good match. Select Production or Debug, not both !!"
 #endif
--- a/Source/core/Trace.h
+++ b/Source/core/Trace.h
@@ -134,9 +134,22 @@ namespace WPEFramework {
 
 #define VERIFY(x, y) assert(x == y)
 #else
-#define ASSERT(x)
+
+#define ASSERT(expr)                                                                                            \
+    do {                                                                                                        \
+        if (!(expr)) {                                                                                          \
+            ASSERT_LOGGER("===== $$ [%d]: ASSERT [%s:%d] (%s)\n", TRACE_PROCESS_ID, __FILE__, __LINE__, #expr); \
+        }                                                                                                       \
+    } while(0)
+
 #define VERIFY(x, y) x
-#define ASSERT_VERBOSE(x, y, ...)
+#define ASSERT_VERBOSE(expr, format, ...)                                                                                                            \
+    do {                                                                                                                                             \
+        if (!(expr)) {                                                                                                                               \
+            ASSERT_LOGGER("===== $$ [%d]: ASSERT [%s:%d] (%s)\n         " #format "\n", TRACE_PROCESS_ID, __FILE__, __LINE__, #expr, ##__VA_ARGS__); \
+        }                                                                                                                                            \
+    } while(0)
+
 #endif
 
 #define LOG(LEVEL, MESSAGE)                                                         \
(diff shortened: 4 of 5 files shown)


Source/core/IIterator.h
```diff
@@ -174,7 +174,7 @@ namespace Core {
 
         inline ELEMENT operator->()
         {
-            ASSERT(IsValid());
```
pwielders, line 177: What is the use of this change? This is a copy of what is in IsValid(). So now the logic is spread.

Source/core/Portability.h
```diff
@@ -396,7 +396,6 @@ inline void EXTERNAL SleepS(unsigned int a_Time)
 #endif
 
 #if defined(_THUNDER_DEBUG) || !defined(_THUNDER_NDEBUG)
-#define __DEBUG__
```
pwielders, line 399: As discussed on slack this is an internal flag that is now exposed. Just use the external flags.

pwielders (commented):
What is the purpose of this change? What are you trying to resolve or what are you trying to achieve? Are you looking for an ASSERT that only reports but does not terminate executions?

zrash:
What is the purpose of this change? 
> This change is for the ticket DELIA-56519.  As explained in the ticket, when the other components which include the thunder header files for plugin implementation,  the `__DEBUG__` flag is automatically defined for them and the utility macros like ASSERT are being defined as abort().  So, the _DEBUG_ must be defined in CMakeList.txt based on the BUILD_TYPE not in the header file. So that the every components get the same header file for ref.

What are you trying to resolve or what are you trying to achieve?
> Explained above.

Are you looking for an ASSERT that only reports but does not terminate executions?
> Yes am also looking for ASSERT that reports only logs/info without terminating executions.

As discussed on slack this is an internal flag that is now exposed. Just use the external flags.?
> The __DEBUG__ flag is just passed from CMAKE and it is not exposed to yocto/rdk.  Still the yocto/rdk passed the flag BUILD_TYPE as any one of the following"
- Debug
- DebugOptimized
- ReleaseSymbols
- Release
- Production

pwielders:
See the Ticket DELIA-56519 where I explain why tinkering with internal thunder pre-processor flags is a very dangerous thing and can lead to unexpected crashes. If Thunder is build with different internal flags than the modules using Thunder,  code is excluded in one where the other might assume it is. So this is a very, very dangerous patch and should *not* be included.

pwielders:
"So, the DEBUG must be defined in CMakeList.txt based on the BUILD_TYPE not in the header file. So that the every components get the same header file for ref." 
Agree and it is. The DEBUG flag to which you are referring *is* based on the BUILD_TYPE since the build type sets the internal preprocessor defines here:  
https://github.com/rdkcentral/Thunder/blob/master/cmake/common/CompileSettings.cmake#L70-L100
Which in turn triggers this the flag you now expose externally, here: 
https://github.com/rdkcentral/Thunder/blob/master/Source/core/Portability.h#L506-L511

and it could be that that is not working correctly in R2, but than you still need to fix it as done on the master, internally and *not* expose this flag externally.

So looking at R2 and R3, I think this is a bug that has been solved long time ago and never ended up in R2, but on R3 it is solved properly (internally) and you should "copy" this solution so you do not ceate inconsistent builds with exposing the _DEBUG_ flag.

pwielders:
Finally :-) After a conf-call of 15 minutes (on this topic) the confusion is resolved. As can be seen in the last modified line, it is not passing the __DEBUG__ in CFLAGS, but the users have to pass the "BUILD_TYPE" to the CMakeList.txt. That makes me feel comfortable as now we are sending an external CMake variable which is oke (thats why they are external :-) ). So the internal Thunder flags are *not* send through the CFlags!

This issue is already been resolved in R3, so this does not require any follow up on R3.

Source/core/IIterator.h
pwielders: This (also learned it during the conf -call) is a compilation issue caused in GCC 9.3.0 but only in the RDK build environment (not in any other environment). Information is send, Metrological will also deep dive into this to see what the root casue is. For now it reolves the issue in the RDK build but does require a further investigation. INternal Jira ticket will be raised.

Source/com/Administrator.cpp
```diff
@@ -294,7 +294,9 @@ namespace RPC {
                     }
                 } while ((loop->Decrement()) && (result == Core::ERROR_NONE));
 
+#ifdef __DEBUG__
```
pwielders, line 297: If this is doen to supress the ASSERT, this also requires a follow up inverstigation. This assert should not be triggetred.

pwielders (approved):
However 2 tickets for further investigations required. 

1) Why is the RDK GCC 9.3.0 compilation failing on the IsValid() in the operator-> and no where else and fixed by copy pasting the contents of the IsValid() here?
2) If @zrash confirms that the ASSERT here: https://github.com/rdkcentral/Thunder/pull/996/files#diff-ed390913b30d98d0cb0820aab7e10a5efee16048581b83b4c8648d6039bf7360R298 fires, investigate why it fires.

Source/com/Administrator.cpp
pwielders: Why is this ```__DEBUG__``` flag around this? ASSERT by definition  is only functional if ```__DEBUG__``` is set, so I see no added value here, or is it to avoid that the ASSERT fires? If so we need to further investigate

karuna2git:
Thanks for the approval @pwielders .. Merging the change to R2.

PR #1006, opened 2022-06-27
Rdk 35564/3

mateusz-adamski-red:
Should be merged together with https://github.com/rdkcentral/rdkservices/pull/2735/


Source/core/SystemInfo.h
```diff
@@ -46,8 +46,8 @@ namespace Core {
         extern "C" const char* MODULE_NAME;
 
         extern "C" EXTERNAL uint32_t Reboot();
-        extern "C" EXTERNAL_EXPORT const char* ModuleBuildRef();
```
pwielders: This change might cause the issue we are observing with the ASSERT that fires. These methods should always be exported, regardless of the chosen module. Each module will include these to indicate the data they will expose. This should never use a exported symbol from another module.

pwielders:
@mateusz-adamski-red can we close the other pull request? https://github.com/rdkcentral/Thunder/pull/979 ?

pwielders (approved):

PR #1008, opened 2022-06-28
RDKTV-15390 : WPEFramework Crash  in Core::ResourceMonitorType

nsahu910:
Reason for change: closing the Socket
Risks: None
Signed-off-by: Neeraj Sahu Neeraj_Sahu@comcast.com

### DIFF
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -373,7 +373,11 @@ namespace Core {
 
         // Make sure the socket is closed before you destruct. Otherwise
         // the virtuals might be called, which are destructed at this point !!!!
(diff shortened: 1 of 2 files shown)


pwielders:
Taken from the master! Oke to go.

pwielders (approved):

karuna2git (commented):
Hey @nsahu910,
The better fix is to change the m_state from volatile to atomic. This is what needed in Multi threaded environment. Pl refer @pwielders change in master for ref.

Checking isClosed() is better here but in RDK, the WPEFramework is always built as PRODUCTION. So, the ASSERT is not going to block; it will proceed to the DestroySocket() even if socket is NOT closed.

PR #1015, opened 2022-07-05
Added stress test for Workerpool

Vinodsathyaseelan:


Tests/workerpool-test/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+add_executable(WorkerPoolTest
+    Module.cpp
+    WorkerPoolTest.cpp
+)
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread")
+
+target_link_libraries(WorkerPoolTest
+    PRIVATE
+        ${NAMESPACE}Core
+        ${NAMESPACE}Tracing
```
MFransen69: I would not expect tracing to be needed, what is it used for?
In the module.h I see however that com and plugins header files are also included. Are they really needed and if so should they not be included here?

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+            TesterTask& operator=(const TesterTask&) = delete;
+
+            TesterTask(WorkerPoolTestAdministrator& parent)
+                : _parent(parent)
+            {
+                _jobMetaData.JobId = _parent.getNextJobId();
+                _jobMetaData.RepeatCnt = _parent.getRandomValue(0, 5);
+                _jobMetaData.bIsRevoked = false;
+            }
+            ~TesterTask() = default;
```
MFransen69: please also make it override

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+        class TesterTask : public Core::IDispatch {
+        public:
+            TesterTask(const TesterTask&) = delete;
+            TesterTask& operator=(const TesterTask&) = delete;
+
+            TesterTask(WorkerPoolTestAdministrator& parent)
+                : _parent(parent)
+            {
+                _jobMetaData.JobId = _parent.getNextJobId();
```
MFransen69: Would be more correct to do it in the initializer list, make a c'tor for TestJobMetaData or I think you might be able to do it with {} initialization (never tried however)

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+                _jobMetaData.RepeatCnt = _parent.getRandomValue(0, 5);
+                _jobMetaData.bIsRevoked = false;
+            }
+            ~TesterTask() = default;
+
+            void Dispatch() override {
+                _parent.UpdateJobDetails(_jobMetaData.JobId);
+            }
+
+            uint32_t JobId() {
```
MFransen69: shouldn't this be const? Also for a lot of methods below

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+                return (JobId == other.JobId);
+            }
+        };
+
+        class TesterTask : public Core::IDispatch {
+        public:
+            TesterTask(const TesterTask&) = delete;
+            TesterTask& operator=(const TesterTask&) = delete;
+
+            TesterTask(WorkerPoolTestAdministrator& parent)
```
MFransen69: Better to make this explicit, Please also check other c'tors below

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+                Dispatcher& operator=(const Dispatcher&) = delete;
+
+                Dispatcher() = default;
+                ~Dispatcher() override = default;
+
+            private:
+                void Initialize() override { }
+                void Deinitialize() override { }
+                void Dispatch(Core::IDispatch* job) override {
+                    job->Dispatch();
```
MFransen69, line 119: add an assert to check for nullptr?

Tests/workerpool-test/WorkerPoolTest.cpp
```diff
@@ -0,0 +1,442 @@
...
+            WorkerPoolImplementation(const WorkerPoolImplementation&) = delete;
+            WorkerPoolImplementation& operator=(const WorkerPoolImplementation&) = delete;
+
+            WorkerPoolImplementation(const uint8_t threads, const uint32_t stackSize, const uint32_t queueSize)
+                : WorkerPool(threads, stackSize, queueSize, &_dispatcher)
+                , _dispatcher(){ }
+
+            ~WorkerPoolImplementation()
+            {
+                printf("%s[ %d ] Called \n", __FUNCTION__, __LINE__);
```
MFransen69: Are these kind of printfs useful for the user of the testtool or were these for helping in development. In case of the latter perhaps better to remove them?

MFransen69 (changes requested):

Tests/workerpool-test/CMakeLists.txt
Vinodsathyaseelan: Tracing is not used in WorkerPoolTest. Corrected CMakeLists.txt & Module.h by removing unused dependencies.

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Yes, JobId is const, but RepeatCnt and Revoked status are not.

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Done.

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Removed all the printf added for debugging purpose.

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Done

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Done

Tests/workerpool-test/WorkerPoolTest.cpp
Vinodsathyaseelan: Done.

MFransen69 (approved):

PR #1024, opened 2022-07-14
RDKTV-16992: Added timeout and synchronisation when stopping containers

willtallentiresky:
Reason for change: It was possible for successive calls to DobbyContainer::Stop
to be made due to the WPEFramework not being notified quickly enough
that a container had stopped.

This change implements the missing timeout functionality
within that method to prevent the opportunity for quick multiple calls.
When the timeout is zero this method behaves as previous.

Test Procedure: Stop containers with zero and non-zero timeouts and
ensure there are no occurrences of containers being stopped multiple
times.

Risks: Low

Signed-off-by: Will Tallentire <will.tallentire@sky.uk>

### DIFF
--- a/Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
+++ b/Source/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
@@ -371,15 +371,54 @@ namespace ProcessContainers {
 
     bool DobbyContainer::Stop(const uint32_t timeout /*ms*/)
     {
-        // TODO: add timeout support
         bool result = false;
+        bool stoppedSuccessfully = false;
         auto& admin = static_cast<DobbyContainerAdministrator&>(DobbyContainerAdministrator::Instance());
 
-        bool stoppedSuccessfully = admin.mDobbyProxy->stopContainer(_descriptor, false);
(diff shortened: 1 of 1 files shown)


willtallentiresky:
@pwielders In DobbyContainer::~DobbyContainer(), there is a Stop(Core::infinite), which does not make much sense as this could cause a hang. Passing zero instead would make the Stop method work as it did previously before this change. Would there be any objection to changing this to zero or a reasonable timeout value, and do you know where else this Stop method may be called?

pwielders (approved):

pwielders:
I guess this requires a cherry-pick to master. Could you create a pull request for the master?

karuna2git (approved):
>It was possible for successive calls to DobbyContainer::Stop to be made due to the WPEFramework not being notified quickly enough
that a container had stopped.

I have concern that the successive calls are still not blocked as every call is going to spawn a thread and attempting to call `mDobbyProxy->stopContainer(temp_descriptor, false);`.

sounds like we just added timeout support but successive calls are still a problem. We must return those successive `DobbyContainer::Stop()` calls saying stop is in progress.

But based on the conversation with @willtallentiresky where he mentioned that intention is to add timeout support and issues observed will be handled separate., I'm approving the changes.

PR #1027, opened 2022-07-20
Include GCC compilation flags

cmuhammedrafi:
-Wall and -Werror compilation flag included


pwielders:
As we are working in C++ Could we use C++ style castin. See master: https://github.com/rdkcentral/Thunder/blob/master/Source/core/Services.cpp#L69 where it has already been fixed.
See: https://riptutorial.com/cplusplus/example/19819/c-style-casting for more info

PR #1034, opened 2022-08-02
LLAMA-6991: Revert part of XIONE-8644 commit to avoid crash

mateusz-adamski-red:


pwielders (commented):
It is incorrect, not to remove the elements that have been added from the object on a clear as than, next time they are added, the old definition is used. This was a bug in R2 and this was a fix for it. I think the root-casue is the usage of these elements. If you drop the JSON object, also drop the refereneces that you have to it, or copy the value. So the rootcause is propbably in the using code.
This is *not* to be merged to the master/R3/R4.

PR #1036, opened 2022-08-03
RDKDEV-340: Thread: Fix handling of pthread_create() failures

rjpdasilva:
When pthread_create() fails, it returns a value different than zero, not
-1 as the current implementation was assuming. This was detected in a
use case where some threads were trying to be created with approximately
1GiB stack size and failing.

Additionally, when the thread creation fails, any waits or set states on
it must fail, otherwise the system may become deadlocked waiting for
some thread that could not even be created. For this, introduce a new
'FAILED' state and handle it accordingly.

Signed-off-by: Ricardo Silva <ricardo.josilva@parceiros.nos.pt>

### DIFF
--- a/Source/core/Thread.cpp
+++ b/Source/core/Thread.cpp
@@ -83,10 +83,11 @@ namespace Core {
 
         // If there is no thread, the "new" thread can also not free the destructor,
         // then it is up to us.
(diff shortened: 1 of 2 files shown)


pwielders (approved):

pwielders:
Good fix. WIll cherry pick it to R4/Master

anand-ky (approved):

PR #1037, opened 2022-08-05
Modified to disable low level & high level trace for production build.

ghost:


Source/core/Trace.h
```diff
@@ -49,6 +49,9 @@ namespace WPEFramework {
 #define TRACE_PROCESS_ID ::getpid()
 #endif
 
+#ifdef _THUNDER_PRODUCTION
```
pwielders: There should be no need to do this. This should be regulayed by using the TRACE_LEVEL macro which, for production, should just be set to 0.

ghost: Done. Here the _TRACE_LEVEL can be made 0 externally by using cmake variable ENABLED_TRACING_LEVEL.

pwielders (approved):

PR #1044, opened 2022-08-12
RDKTV-16258: WPEWebProcess crash due to WPEFramework lambda signature

cmuhammedrafi:
Reason for change: wpeframework lambda fuction crash issue fix
Test Procedure: Verify in Hisense TV
Risks: low

Signed-off-by: cmuhammedrafi <muhammed_rafi@comcast.com>


pwielders (commented):
If this is needed, it means there is a severe issue on a higher layer that should be checked. Just above is a line that generates an ASSERT (in debug) as defensive measurements to bail out. 
What this means is that a call from the Proxy on the otherside (other process) is creating a message. In this message it places a NULLPTR for the actual object on which the stub should call the method. If that is the case, There is no need for that proxy to call as he "knows" that the object/interface does not exist. It is a spec violation.
What could potentially trigger this would be a corrupted frame. Question is how it got corrupted.
For R2, fine to add. Please **do not** add it to the Master (the ASSERT is sufficient) and please do a further analyses on the root-cause as this is just masking a very severe issue.

karuna2git (approved):
Adding only to R2

PR #1049, opened 2022-08-22
RDKDEV-345i: Crypto: Random: Prefer the OpenSSL RNG if 'SECURESOCKETS…

rjpdasilva:
…_ENABLED' is set

Prefer the Random Number Generator (RNG) feature from OpenSSL to the C
library's one when OpenSSL is enabled ('SECURESOCKETS_ENABLED' is set),
both for seeding (Crypto::Reseed()) and random number generation (the
several Crypto::Random() overrides).

With OpenSSL, seeding is done using RAND_poll() and random number
generation with RAND_bytes().

The OpenSSL RNG is a Cryptographically Secure Pseudo Random Number
Generator (CSPRNG), which provides a better and more secure RNG than the
C library's one, so it should be preferred when available. In fact, with
tests around the SecurityAgent plugin and the tokens generated by it
(which use these Crypto::Random functions), it was observed that for a
given payload, the tokens generated where always the same across
different devices and reboots. This happens precisely because the C
library's RNG is always returning the same random numbers. Even though
this can be mitigated by forcing a Reseed() before requesting the random
numbers, still this Reseeding() is based on the current time, which is
not the best source of entropy for seeding. With OpenSSL RNG, this issue
with the tokens is not happening, even without the re-seeding.

Still falling back to the C library's implementations when OpenSSL is
not enabled, or when it's enabled but for some reason their RAND
functions fail.

Signed-off-by: Ricardo Silva <ricardo.josilva@parceiros.nos.pt>

### DIFF
--- a/Source/cryptalgo/Random.cpp
+++ b/Source/cryptalgo/Random.cpp
@@ -27,6 +27,10 @@
 #include "Winsock2.h"
 #endif // __WINDOWS__
 
+#if defined(SECURESOCKETS_ENABLED)
+#include <openssl/rand.h>
+#endif // SECURESOCKETS_ENABLED
+
 namespace WPEFramework {
 namespace Crypto {
     // --------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pwielders (approved):

pwielders:
@pwielders , reminder to Cherry pick this to the master!

karuna2git (approved):

pradeeptakdas:
Internal ticket created for tracking is https://ccp.sys.comcast.net/browse/RDKCOM-3276

PR #1063, opened 2022-09-13
[ARRISEOS-42363] Don't flush libraries in Dispatch

marcin-mielczarczyk-red:
When flush of libWPEFrameworkWebKitBrowserImpl.so library happens in Dispatch() there are cases where the other thread is still using it which leads to core dump of WPEProcess.

### DIFF
--- a/Source/WPEProcess/Process.cpp
+++ b/Source/WPEProcess/Process.cpp
@@ -75,8 +75,6 @@ namespace Process {
 
         public:
             void Dispatch() {
-                Core::ServiceAdministrator::Instance().FlushLibraries();
-
                 uint32_t instances = Core::ServiceAdministrator::Instance().Instances();
 
                 if (instances != 0) {


pwielders (commented):
Seen this but it might resolve an issue but it definitly has to be called sometime, so it probably needs a bit more of investigation. If it is not called, librar
ies will not be unloaded from memory although all refernces to the librraies have been dropped.

ganeshprasads:
Hi @marcin-mielczarczyk-red Can you please sign to the "Contributor License Agreement" from the link mentioned above

marcin-mielczarczyk-red:
> Seen this but it might resolve an issue but it definitly has to be called sometime, so it probably needs a bit more of investigation. If it is not called, librar ies will not be unloaded from memory although all refernces to the librraies have been dropped.

I remember that I checked that and I'm sure that library is unloaded when object is being destructed anyway. There are prints which confirm that. Unloading happens at the very end of destruction process.

pwielders:
After analyzing this pull-request in more detail, and combining it with other reports on potential crashes from LGI at the moment of deactvating/stopping plugins, I think the rootcause is the moment of unloading (releasing) the libraries (plugin code loaded dynamic libraries) as potentially there could still be work pending in the workerpools.
As a follow up we introduced this pull-request: https://github.com/rdkcentral/Thunder/pull/1108 to resolve it in a more deterministic way for WPEFramework and WPEProcess.
So this pull-request is not needed as 1108 fixes this issue and more..

pwielders:
> > Seen this but it might resolve an issue but it definitly has to be called sometime, so it probably needs a bit more of investigation. If it is not called, librar ies will not be unloaded from memory although all refernces to the librraies have been dropped.
> 
> I remember that I checked that and I'm sure that library is unloaded when object is being destructed anyway. There are prints which confirm that. Unloading happens at the very end of destruction process.

Thats correct at unloading it will always unload anythng still in there but this unload was also intended to "unload" as a dependency during runtime and unloaded during runtime. Really would love a whiteboard to try to explain this cause this is difficult to explain in txt but what I am trying to say:

1) Plugin -> OutOfProcess Plugin.so called [A]

in the WPEProcess that loaded the OutOfProcess Plugin.so [A] another COMRPC object [B] in librray [B] is added to this process by instantiateat runtime. The OutOfProcess Plugin [A] uses this object for a short period and than releases  COMRPC object [B], but the OutOfProcess Plugin continues to run, We do want the COMRPC object [B] in library [B] to be unloaded at the time we drop the last reference to OMRPC object [B] and not at the time that the OutOfProces Plugin [A] is unloaded. 
By removing the call in this PR, the COMRPC [B] object would nolonger be unloaded at the moment we drop the last reference to [B] but at the moment we unload [A]...

@marcin-mielczarczyk-red hope this makes sense, if not let talk on Skype

marcin-mielczarczyk-red:
@pwielders it makes sense but unfortunately I'm not able to test your changes because it doesn't apply on R2 branch.
I tried to solve conflicts but there are too many dependencies on additional refactor being made in the mean time.
Could you please backport your changes to R2 so I'll be able to test them?

pwielders:
Backporting will really be a pain in the *, ;-) since a lot around the Workerpool has changed (we would need to upgrade the whole workerpool stuff). So I suggest that we apply your change to the R2 branch, much smaller, but be aware the issue is a bit bigger than only in WPEProcess.

PR #1066, opened 2022-09-15
[DOWNLOADPATH] Add an additional path with the same sematics as the P…

PaviP98:
…ersistentPath.

The new functionality from the Packager (downloadable plugins) requires that a new location should be serached for plugins (the plugin binary) as well for starting that later installed plugin. The DoenloadPath allows for this to be indicated. Set the DownloadPath ("downloadpath":) in the /etc/WPEFramework/config.json file to search this path for a potential plugin.


pwielders:
No longer needed as we are changing the way of working!

PR #1067, opened 2022-09-15
User:Group: set persistent/data path permissions/user/group based on the given values

HaseenaSainul:


Source/WPEFramework/PluginHost.cpp
```diff
@@ -489,12 +489,23 @@ POP_WARNING()
                 if (_config->Process().Priority() != 0) {
                     myself.Priority(_config->Process().Priority());
                 }
-                if (_config->Process().User().empty() == false) {
-                    myself.User(_config->Process().User());
-                }
+
+                Core::Directory dataPath(_config->DataPath().c_str());
```
pwielders: DataPath is pointing to ROM (Read Only Memory) storage of the setup, so by definition, you should not (and if configured correctly) can not change proeprties on this path. This is something that needs to be done properly at image build time.
So this, if it works, it means that the config is incorrect (DATA Path is not pointing to the ROM storage) and if it solves an issue, it is hiding a root cause that the plugin did not properly use the dataPath. So please do not make any modifications to the DataPath and resolve the root cause in the plugin.

Source/WPEFramework/PluginHost.cpp
```diff
@@ -489,12 +489,23 @@ POP_WARNING()
...
+                dataPath.Permission((S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IWGRP | S_IXGRP));
+
+                Core::Directory persistentPath(_config->PersistentPath().c_str());
+                persistentPath.Permission((S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IWGRP | S_IXGRP));
+
                 if (_config->Process().Group().empty() == false) {
                     myself.Group(_config->Process().Group());
+                    dataPath.Group(_config->Process().Group());
```
pwielders: As above, do not change the DataPath...

Source/WPEProcess/Process.cpp
```diff
@@ -649,7 +649,22 @@ int main(int argc, char** argv)
             if (options.Group != nullptr) {
                 Core::ProcessCurrent().Group(string(options.Group));
             }
+
+            Core::Directory dataPath(options.DataPath.c_str());
+            dataPath.Permission((S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP));
```
pwielders: See previous remark, remove the changes on the dataPath

Source/WPEProcess/Process.cpp
```diff
@@ -649,7 +649,22 @@ int main(int argc, char** argv)
...
+
+            Core::Directory persistentPath(options.PersistentPath.c_str());
+            persistentPath.Permission((S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP));
+
             if (options.User != nullptr) {
+                string group = Core::ProcessCurrent().Group();
```
pwielders: Also here, ther is something incorrect on the way it is working. It is the PLugins, responsibility to make sure that the rights are handled correctly. Now we are overruling the settings of a plugin to, by definition, allow a group user to access is. It could be that the plugin developer wants the spawned out-of-process plugin part not to write to the Persistent directory. If so, we know killed that feature :-) So I think the settings on any directory from the spawned process should never happen. 
Lets investigate the root-casue that lead to this change (Amazon rights ? ) and lets fix it in the Amazon Plugin. Talked already also to @bramoosterhuis , There is a CreatePersistentPath that probably requires right settings.

Source/tracing/TraceUnit.cpp
```diff
@@ -51,6 +51,7 @@ namespace Trace {
         // Make sure the trace file opened proeprly.
         TRACE_L1("Opened a file to stash my traces at: %s [%d] and doorbell: %s", name.c_str(), CyclicBufferSize, doorBell.c_str());
         ASSERT (IsValid() == true);
+        Group(Core::ProcessCurrent().Group());
```
pwielders: Might be that this should be set from the PLuginHost file as well and not specific from here?

HaseenaSainul:
thanks @pwielders for the comments, will check and update

pwielders:
@HaseenaSainul do we still need the exposure of this config object, since we now moved the LinkLibraryPath/SystemRootPath. If nolonger needed, I prefer to keep it hiddden so it can not be abused.

Source/tracing/TraceUnit.cpp
HaseenaSainul: removed

Source/WPEProcess/Process.cpp
HaseenaSainul: removed

Source/WPEProcess/Process.cpp
HaseenaSainul: Added this settings from EnablePersistentPath Methods

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: moved from here to build system

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: removed

HaseenaSainul:
> @HaseenaSainul do we still need the exposure of this config object, since we now moved the LinkLibraryPath/SystemRootPath. If nolonger needed, I prefer to keep it hiddden so it can not be abused.

Now we need to get Group and User from plugins Initialise call to set permission through EnablePersistentStorage method. So moved to outside of Shell.cpp

Source/plugins/IShell.h
```diff
@@ -328,7 +328,7 @@ namespace PluginHost {
 
             return (handler == nullptr ? nullptr : handler->RemoteConnection(connectionId));
         }
-        inline uint32_t EnablePersistentStorage() {
+        inline uint32_t EnablePersistentStorage(uint32_t permission = 0, const string& group = {}, const string& user = {}) {
```
pwielders: Can we change the group and user from position, cause most commands have the user before the group, so this order is a bit different than everywhere else, so:
inline uint32_t EnablePersistentStorage(uint32_t permission = 0, const string& user = {}, const string& group = {}) {

Source/core/DataElementFile.cpp
```diff
@@ -62,6 +62,7 @@ namespace Core {
 
         if ((type & File::CREATE) != 0) {
             m_File.Create(type);
+            m_File.Permission(type & ~(File::CREATE));
```
pwielders: I think you need to blank out more? Like sharing ?

Source/core/NodeId.h
```diff
@@ -269,6 +261,8 @@ namespace Core {
         bool IsMulticast() const;
         uint8_t DefaultMask() const;
 
+        uint32_t EnablePermission() const;
```
pwielders: The NodeId is a data carrier object and should not have active methods, only carry/translate data. Permission setting is an active method that should from design perspective not be on a data carrier object. Suggest to "expose" the Group that might now be passed on the node (so a *const string& Group() const;* method) and leave the setting of the permissions up to new AccessControl.h class in the specific implementations (doorbel, socketport and others)

Source/core/AccessControl.h
```diff
@@ -0,0 +1,86 @@
...
+            mode_t mode = modeFlags & 0x0FFFFF;
+
+            if (::chmod(node.c_str(), mode) != 0) {
+                result = Core::ERROR_GENERAL;
+            }
+#else
+            result = Core::ERROR_NOT_SUPPORTED;
+#endif
+            return result;
+        }
```
pwielders, line 87: Suggest to add the NodeId::EnablePermission() (see comment furhter down the line) as a convenience method here like: *static uint32_t Apply(const NodeId& node);* Than this is a convenience method in the AccessControl space to actvely apply Ownership and Rights to a given NodeId. So in the other functions you could write: AccessControl::Apply(nodeId) == Core::ERROR_NONE.
This leaves the NodeId a pure Data Carrier object and rights/ownership managment in the AccessControl class..

pwielders (changes requested):

Source/core/AccessControl.h
HaseenaSainul: updated

Source/core/NodeId.h
HaseenaSainul: done

Source/plugins/IShell.h
HaseenaSainul: changed parameter order

Source/core/DataElementFile.cpp
HaseenaSainul: yes SHARING also flag also removed

pwielders (approved):
Perfect! Thanks !

PR #1070, opened 2022-09-20
Change loader paths to SystemLoaderPaths and use prefix this to generate dynamic library path for systemrootpath enabled plugins

HaseenaSainul:


Source/WPEFramework/PluginServer.h
```diff
@@ -1290,32 +1308,46 @@ namespace PluginHost {
...
+                    Add(_T("startup"), &Startup);
                 }
 
                 ~Plugin() override = default;
 
             public:
                 Core::JSON::Boolean AutoStart;
                 Core::JSON::String Configuration;
+                Core::JSON::String SystemRootPath;
+                Core::JSON::EnumType<PluginHost::IShell::startup> Startup;
```
pwielders, line 1343: By adding this value, I guess we can drop the Autpostart boolean.

Source/com/Communicator.cpp
```diff
@@ -452,11 +528,16 @@ PUSH_WARNING(DISABLE_WARNING_THIS_IN_MEMBER_INITIALIZER_LIST)
...
     }
 
-    void Communicator::Destroy(const uint32_t id) {
+    void Communicator::Destroy(const uint32_t id)
+    {
         // This is a forceull call, blocking, to kill that specific connection
         g_destructor.ForceDestruct(id);
     }
 
+    const std::vector<string> Process::DynamicLoaderPaths() const
```
pwielders: lets return a const std::vector<string>& (reference) since we do not need a copy here, we just need access to the orginal..

Source/com/Communicator.cpp
```diff
@@ -33,6 +33,82 @@ namespace RPC {
...
+            : _downloadLists()
+        {
+            ReadList(LoaderConfig, _downloadLists);
+            _downloadLists.emplace_back(_T("/usr/lib/"));
+            _downloadLists.emplace_back(_T("/lib/"));
+        }
+        ~DynamicLoaderPaths() = default;
+
+    public:
+        const std::vector<string> Paths() const
```
pwielders: Lets make this a const std::vector<string>& here as a return value. Otherwise a copy is created...

Source/com/Communicator.h
```diff
@@ -421,10 +425,35 @@ namespace RPC {
...
+                    newLDLibraryPaths += oldLDLibraryPaths.substr(start,
+                                        ((end != string::npos) ? (end - start + 1) : end));
+                    start = end;
+                    if (end != string::npos) {
+                        end = oldLDLibraryPaths.find(':', start + 1);
+                    }
+                } while (start != string::npos);
+            }
+
+            std::vector<string> loaderPaths = DynamicLoaderPaths();
```
pwielders: Just accept a ```const std::vector<string>&``` here to avoid a copy..

Source/WPEFramework/PluginServer.h
HaseenaSainul: done

Source/com/Communicator.h
HaseenaSainul: done

Source/com/Communicator.cpp
HaseenaSainul: done

Source/com/Communicator.cpp
HaseenaSainul: done

pwielders (approved):

PR #1101, opened 2022-10-07
[Stubgen] Fix redundant stub code when including other interfaces

sebaszm:


HaseenaSainul (approved):
Looks fine while testing

PR #1114, opened 2022-10-20
Development/fd passing

bramoosterhuis:


Source/core/PriviligedRequest.h
```diff
@@ -0,0 +1,360 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+// #define MODULE_NAME PriviligedRequest
+#include <core/core.h>
```
pwielders: This file is part of core, so the include should be "part of" the core, and not to be treated as an external include file. Please use: #include "core.h" and start by Including "module.h (see other header files)

pwielders (changes requested):
Minor request to change.

Source/core/PriviligedRequest.h
bramoosterhuis: Fixed!

pwielders (approved):

PR #1115, opened 2022-10-21
[JsonGen] Fix indexed properties

sebaszm:
Seems they got broken with the change of default compliance mode...


HaseenaSainul (approved):
LGTM

PR #1128, opened 2022-10-31
[WINDOWS] We are prepring the FileObserver for Windows OS.

pwielders:


Source/core/FileObserver.h
```diff
@@ -245,42 +249,173 @@ class FileSystemMonitor : public Core::IResource {
...
+            }
         }
 
     public:
-        HANDLE Event() {
-            return(_syncEvent);
+        void Register(ICallback* client) {
+            ASSERT(std::find(_clients.begin(), _clients.end(), client) == _clients.end());
+
+            _clients.push_back(client);
```
MFransen69, line 310: Not protected I guess for a reason? (performance and not used multithreaded)

Source/core/FileObserver.h
```diff
@@ -245,42 +249,173 @@ class FileSystemMonitor : public Core::IResource {
...
+                    TRACE_L1("Could not start observing: %s", _filename.c_str());
+                }
+
+                for (auto& client : _clients) {
+                    client->Updated();
+                }
+            }
+        }
+
+        /*
```
MFransen69: remove commented out code or do you still need it?

MFransen69 (changes requested):

Source/core/FileObserver.h
pwielders: It is protected by the outer class. Inner class can not be directly accessed. So not required to lock (unless we want to double lock :-) )

Source/core/FileObserver.h
pwielders: Don't know if we still need it, found it with an example on internet. Wanted to store it for reference later on. Can also link the example from which it was taken :-)

pwielders:
Eleborated a bit on the design decisions. Was a bit too much left to the reader :-) Hope it explains now.

MFransen69 (dismissed):

MFransen69 (approved):

MFransen69 (approved):

PR #1133, opened 2022-11-02
[JsonGen] Add bitmask enums

sebaszm:
Adds a `@bitmask` tag that can be put beside the enum definition, when it means all usage of the enum is meant to be a mask (ie. an array on JSON-RPC side), or beside a struct member or a method parameter when it means the bitmask only refers to the single particular usage. Also works with `enum struct` or `enum class` or if the underlying type is not specified (though with a warning).

The generator will give an early error if the enum marked with `@bitmask` has enumerator values not fitting power of 2 (including 0) or has no enumerator values at all.

@pwielders Please note the added JSON.h templates!

Examples:

       enum devicetype : uint8_t {      
            HEADPHONE = 1,
            SPEAKER = 2,
            RECORDER = 4,
            AMPLIFIER = 8
        };

        struct DeviceProperties {
            string Config;
            devicetype Type /* @bitmask */;
        };

        // @property
        virtual uint32_t SupportedDeviceTypes(devicetype& devicesMask /* @out @bitmask */) const = 0;

 Or

      enum devicetype : uint8_t /* @bitmask */ {      
            HEADPHONE = 1,
            SPEAKER = 2,
            RECORDER = 4,
            AMPLIFIER = 8
        };


pwielders (approved):
Wow, thats advanced template coding that took me some time to comprehend, but Looks good ! Nice feature on the JSON::ArrayType..

HaseenaSainul (commented):
First example is not working. It is not generating ArrayType

sebaszm:
> First example is not working. It is not generating ArrayType

Indeed I forgot the `@` sign in the struct. Is that what you mean?

HaseenaSainul:
> > First example is not working. It is not generating ArrayType
> 
> Indeed I forgot the `@` sign in the struct. Is that what you mean?

Not, I tried with @ but it is creating as enum only, not an array of enum. If @bitmask against enum definition then only it is worked for me

HaseenaSainul (approved):

sebaszm:
The tag needs to be in the same sequence point as the member definition, i.e. before the semicolon, otherwise code as this would put the tag with the wrong member:
```
struct {
  int x /* @bitmask */; int y; int z;
};
```
The same applies to enumerators in an enum definition.

HaseenaSainul:
> The tag needs to be in the same sequence point as the member definition, i.e. before the semicolon, otherwise code as this would put the tag with the wrong member:
> 
> ```
> struct {
>   int x /* @bitmask */; int y; int z;
> };
> ```
> 
> The same applies to enumerators in an enum definition.



> The tag needs to be in the same sequence point as the member definition, i.e. before the semicolon, otherwise code as this would put the tag with the wrong member:
> 
> ```
> struct {
>   int x /* @bitmask */; int y; int z;
> };
> ```
> 
> The same applies to enumerators in an enum definition.

yes, that fixed the issue

PR #1135, opened 2022-11-04
[OBSERVERS] Add the observers for Config and ProxyStubs to WPEFramework.

pwielders:


sebaszm (approved):
Still to try it on my own, but the mechanism looks ok!

PR #1160, opened 2022-11-28
CyclicBuffer:

HaseenaSainul:
1. AssureFreeSpace: start required + 1 to make sure that head pointer is not on the tail pointer
2. Allow write if the requested size is less than free size, to avoid infinite loop
3. Add assert if buffer create with readonly mode
4. Check _administrator pointer also for buffer validity

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -32,6 +32,7 @@ namespace Core {
         , _alert(false)
         , _administration(nullptr)
     {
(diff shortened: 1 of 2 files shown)


anand-ky:
@MFransen69 - Pls help add +1.

MFransen69:
> @MFransen69 - Pls help add +1.

No need. As it is R2 feel free to merge without our review

anand-ky (approved):

PR #1166, opened 2022-12-07
WebSocket: update control frames based on mask bit

HaseenaSainul:


pwielders:
Is this a bug raised somewhere ?

HaseenaSainul:
> Is this a bug raised somewhere ?

Yes got, while running fireboat test app against fireboat mock server. It expecting mask bit to be set for client packet. And even I create a web socket with MASK bit as true, it is crashing when it send control messages. Hence it also added to set based on the value of mask bit configured

pwielders (approved):

PR #1167, opened 2022-12-07
RDKDEV-433: Thunder: JSON config generator improvement

paulo-a-junior:
The current implementation of the JSON generation script has a shortcoming: it doesn't allow defining a number as a string.

This PR introduces a minor change, to allow for numbers defined as "'123456789'" to be translated to "123456789" on the final JSON. The default behavior translates "123456789" to 123456789 (i.e., into an integer).


pwielders:
Why would you like to have a number being represented as a string ? What is the use case for this ? If it is a usefull feature (@sebaszm ?) Please start on master and just backport it to R2, so it is not lost for upgrades.

paulo-a-junior:
One such use case: consider a JSON field that is a hex number, using the 0-9A-F symbols. If such a field is, by chance, made up of numeric values only, the resulting JSON file will contain an integer, whereas if the value is made of hex values, the resulting value would be a string.

Concrete use cases: a port value encoded as hex, MAC address that is _accidentally_ made of numbers only.
For our internal use case, it was very straightforward: we needed a number as a string because, at the present time it, was just a number, but, in the future, maybe it could be something else, so, it made sense to have it as a string.

Just a side note: While thinking of the use case to reply to you, I think that maybe it makes sense to accept not only numeric values but anything, thus normalizing the behavior.

For instance, consider two test cases:
```
kv(mac_address "'010203040506'")
kv(mac_address "'01020304BEEF'")
```
Would generate:
```
"mac_address": "010203040506"
"mac_address": "'01020304BEEF'"  (note enclosed single quotes)
```
Extending this train of thought, I am considering that generating ` "mac_address": "01020304BEEF"` - without the enclosed single quotes - would be more useful.

pwielders:
I do prefer explicit annotations. Hence why I think that typically people prefix Hex values with "0x" to avoid any confusion between base 10 numbers and base16 numbers. However I would like to leave it up to the JSON generator tool owner if he thinks it is smart functionality. I assume you have to explicitly mention that you want a number to be a string, cause I do not like the idea that now all numbers become strings.
@sebaszm could you cast your vote on this feature?

sebaszm:
I think it's a reasonable change and coincidentally how the new Python config generator operates. But let's check with @bramoosterhuis regarding the cmake coding.

pwielders:
As this is R2, it definitly is not using the new config generator. It is still using the confif from the CMake modules.

pwielders:
I included @sebaszm as I did not properly look at the commit :-) It is changing the "old" config generation tools, it is not the JSON code generators adapted here. So if this is applicable, we also need to apply this to the new way of config generating, since as of CMake 2.22, this code is not usable anymore.. @bramoosterhuis can you cast your vote on this and define action (if applicable) for the new config generators..

bramoosterhuis:
@pwielders this is only needed for the legacy generator. With the python based generator you can specify a number as a string. So I'm fine with this change since it does not impact future/current code.

pwielders (approved):

bramoosterhuis (approved):

MFransen69:
Hi @paulo-a-junior , 

Please note we can only merge this PR once the CLA has been signed.

paulo-a-junior:
Hello @MFransen69 , 
Just did it. Thanks!

PR #1169, opened 2022-12-07
Tracing: Add extra plugin specific tracing APIs

paulo-a-junior:
This patch adds the following helper APIS:

* TRACE_PLUGIN(CATEGORY, FORMATTER, ...)
* TRACE_PLUGIN_INFO(FORMATTER, ...)
* TRACE_PLUGIN_WARNING(FORMATTER, ...)
* TRACE_PLUGIN_ERROR(FORMATTER, ...)
* TRACE_PLUGIN_FATAL(FORMATTER, ...)
* TRACE_PLUGIN_FUNCTION_ENTRY(TRACE_TYPE)
* TRACE_PLUGIN_FUNCTION_EXIT(TRACE_TYPE)

These APIs provide shorter calls, and also provide specialized tracing functions for method entry and exit, which automatically figure the traced function name, easing maintainability.

Using such APIs would generate log messages by concatenating the MODULE_NAME to the logged message, to allow for a better tracing interpretation. E.g., one could `journalctl |grep MODULE_NAME` and very easily find all logs from a given plugin.

### DIFF
--- a/Source/tracing/TraceControl.h
+++ b/Source/tracing/TraceControl.h
@@ -52,6 +52,76 @@
             &__message__);                                                             \
     }
 
+// ---- Additional Plugin specific macros ----
+
+#ifdef MODULE_NAME  // The following API targets specifically WPEFramework Plugins, which should all define MODULE_NAME
+
+#ifdef __GNUC__ // Be aware: ##__VA_ARGS__ is not portable as it is a GCC extension.
+#define STRINGFY_MODULE_NAME(s) #s
+#define STRINGFY_2_MODULE_NAME(s) STRINGFY_MODULE_NAME(s)
+#define TRACING_PREFIX STRINGFY_2_MODULE_NAME(MODULE_NAME)"::"
+
+#define TRACE_PLUGIN(CATEGORY, FORMATTER, ...)                                                                                    \
+    if (WPEFramework::Trace::TraceType<CATEGORY, &WPEFramework::Core::System::MODULE_NAME>::IsEnabled() == true) { \
+        std::string __temp_string__{ TRACING_PREFIX };                                                             \
+        __temp_string__ += FORMATTER;                                                                              \
+        CATEGORY __data__(__temp_string__.c_str(), ##__VA_ARGS__);                                                 \
+        WPEFramework::Trace::TraceType<CATEGORY, &WPEFramework::Core::System::MODULE_NAME> __message__(__data__);  \
+        WPEFramework::Trace::TraceUnit::Instance().Trace(                                                          \
+            __FILE__,                                                                                              \
+            __LINE__,                                                                                              \
+            typeid(*this).name(),                                                                                  \
+            &__message__);                                                                                         \
+    }
+
+#define TRACE_PLUGIN_INFO(FORMATTER, ...) TRACE_PLUGIN(Trace::Information, FORMATTER, ##__VA_ARGS__)
+
+#define TRACE_PLUGIN_WARNING(FORMATTER, ...) TRACE_PLUGIN(Trace::Warning, FORMATTER, ##__VA_ARGS__)
+
+#define TRACE_PLUGIN_ERROR(FORMATTER, ...) TRACE_PLUGIN(Trace::Error, FORMATTER, ##__VA_ARGS__)
+
+#define TRACE_PLUGIN_FATAL(FORMATTER, ...) TRACE_PLUGIN(Trace::Fatal, FORMATTER, ##__VA_ARGS__)
+
+#define TRACE_PLUGIN_FUNCTION_EXIT(TRACE_TYPE)                                                                       \
+    if (WPEFramework::Trace::TraceType<TRACE_TYPE, &WPEFramework::Core::System::MODULE_NAME>::IsEnabled() == true) { \
+        std::string __temp_string__{ TRACING_PREFIX };                                                               \
+        __temp_string__ += "Exit Method:";                                                                           \
+        __temp_string__ += static_cast<const char*>(__func__);                                                       \
+        TRACE_TYPE __data__(__temp_string__.c_str());                                                                \
+        WPEFramework::Trace::TraceType<TRACE_TYPE, &WPEFramework::Core::System::MODULE_NAME> __message__(__data__);  \
+        WPEFramework::Trace::TraceUnit::Instance().Trace(                                                            \
+            __FILE__,                                                                                                \
+            __LINE__,                                                                                                \
+            typeid(*this).name(),                                                                                    \
+            &__message__);                                                                                           \
+    }
+
+#define TRACE_PLUGIN_FUNCTION_ENTRY(TRACE_TYPE)                                                                      \
+    if (WPEFramework::Trace::TraceType<TRACE_TYPE, &WPEFramework::Core::System::MODULE_NAME>::IsEnabled() == true) { \
+        std::string __temp_string__{ TRACING_PREFIX };                                                               \
+        __temp_string__ += "Entered Method:";                                                                        \
+        __temp_string__ += static_cast<const char*>(__func__);                                                       \
+        TRACE_TYPE __data__(__temp_string__.c_str());                                                                \
+        WPEFramework::Trace::TraceType<TRACE_TYPE, &WPEFramework::Core::System::MODULE_NAME> __message__(__data__);  \
+        WPEFramework::Trace::TraceUnit::Instance().Trace(                                                            \
+            __FILE__,                                                                                                \
+            __LINE__,                                                                                                \
+            typeid(*this).name(),                                                                                    \
+            &__message__);                                                                                           \
+    }
+
+#else
+#warning "Currently, custom tracing is only available for GCC compilers. Tracing calls will resolve to NOPs.\n\
+          for future development, one may consider compiler specific extensions as well."
+#define TRACE_PLUGIN_INFO(...)
+#define TRACE_PLUGIN_ERROR(...)
+#define TRACE_PLUGIN_WARNING(...)
+#define TRACE_PLUGIN_FUNCTION_ENTRY(...)
+#define TRACE_PLUGIN_FUNCTION_EXIT(...)
+#endif // __GNUC__
+
+#endif // MODULE_NAME
+
 // ---- Helper functions ----
 
 // ---- Class Definition ----


pwielders (changes requested):
Thunder is a layered architecture where there are *no* cyclic dependencies. By introducing these convenience methods, you are using the PLUGIN concept on the layer where the is no notion of the PLUGINS yet.  
So if these convenenience methods should be usefull, they must be placed on the right layer (in the plugins module, not tracing)
The offered functionality is quite limited. Tracing categories can be cerated in any shape or form, and good programming does request more categories) However, by only limiting the amount of convenience functions to INFO/WARNING/ERROR/FATAL/FUNCTION_ENTRY/FUNCTION_EXIT, forces the developper to believe that he is back in the old days, where there where only these 5 categories. So I think these convenience functions that change lines from TRACE (Trace::Error, (_T(".."))) to TRACE_PLUGIN_ERROR(_T(""))) brings more limitations than that it resolves.....

paulo-a-junior:
Thanks for your feedback. It sure makes sense to place it inside the `plugins` module.

As for usability, whether or not having the trace type in the macro is not exactly the point I'd like to enhance, and I do agree with you that having these in the macro name might be a bit crippling.

In fact, the main two points I'd like to enhance are:
1 - Being able to tell which plugin is generating a given trace log. This is especially useful when having multiple active logging plugins;
2 - Making the formatter explicit, thus not requiring the second argument to be globbed up in a single argument made of a list of argument.

As a bonus goal, if I could achieve a smaller code line to run the log, I definitely would go for it.

pwielders:
> In fact, the main two points I'd like to enhance are: 
1 - Being able to tell which plugin is generating a given trace log. This is especially useful when having multiple active logging plugins; 
2 - Making the formatter explicit, thus not requiring the second argument to be globbed up in a single argument made of a list of argument.
> 
> As a bonus goal, if I could achieve a smaller code line to run the log, I definitely would go for it.

1)  the information you are printing, the MODULE_NAME (plugin name), is always part of the trace and requires no explicit mentioning in the trace text as it will only use more resources (space) and it requires parsing on the receiving side to get it out. See code: https://github.com/rdkcentral/Thunder/blob/R2/Source/tracing/TraceUnit.cpp#L315 where it is forwarded to the receiver. So taking it up in the message only has disadvantages. 
If it is not send forward in the logs, it is just a matter of adding it to the logline on the publishing side, here: https://github.com/rdkcentral/rdkservices/blob/sprint/22Q4/TraceControl/TraceControl.h#L284 you see that the Modulename is taken from the traceline again, you just need to forward it here: https://github.com/rdkcentral/rdkservices/blob/sprint/22Q4/TraceControl/TraceControl.cpp#L236 and the module name is available in a much more efficient way. and on the specific publishers forward it.

2) is due to the limited look you apply to the, as already very limited set, you expose with these convenience preprocessor statements. The second argument is not always a formatter. There are traces that do not take a string but only parameters the logic to form the actual trace message is in the category. See: https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.h#L236 or https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.h#L286  
So forcing the formatter explicitly for a limited set of TRACES (general traces we discourage to use) will only make the use of properly designed traces unreadable..

"As a bonus goal, if I could achieve a smaller code line to run the log, I definitely would go for it." 
The convenience pre-processor statements are more code (now we only have 1 called TRACE), so the number of lines of code are now bigger, We are only formatting the TRACE lines for , INFO/WARNING/ERROR/FATAL/FUNCTION_ENTRY/FUNCTION_EXIT differently with effectively a saving of "Trace::", space, comma and the elipses, so saving of 12 characters on a single line but creating confusion for TRACE statements with other trace categories as these have no convenience pre-processor definitions.
So to be honoust, these conevenience macros, to me, create more noise and confusion than the 12 characters they save on a line for a limited set of traces.  

Taken all the disadvantages, I suggest not to add these convenience macros and enhance the TraceControl plugin to simply output the Module name that is part of the Trace Statement. As a **bonus** you get also the module names for all the other traces in the system and not only for these few...

As always love to get a second opnion on this, so including Marcel Fransen for a review.

pwielders:
Sorry accidently closed it..

MFransen69:
Indeed I would not add these macro's as is in this PR, so also not in the Plugin folder. I think indeed it is an oversight that the module name is not added to the trace message, but let's solve that in a generic way, so it is always available and not only when you would use these specific macro's. By the way doing it with these macro's would make the use case where you would use grep to find the lines for a specific module not reliable as you cannot trust all lines you are looking for do actually contain the module name. And by the way that is what we try to prevent with our tracing system that you need to search through thousands lines of trace lines. Therefore it is very simple so (even dynamically) turn on tracing for a specific plugin and/or specific category so you only get the relevant trace lines.
So in short I think it is a good idea to add printing the module name for every trace line to Tracing, similar to what we already do for WarningReporting. 
And good to know, before we can accept any PR, you have to sign the CLA, otherwise we are not allowed to merge these.

paulo-a-junior:
Hello, @MFransen69 , @pwielders . Took me a while, but, I've found an alternative for this. It is to path the following two files:

- Source/tracing/Logging.cpp:SysLog() on this repository

- TraceControl/TraceOutput.h:WPEFramework::Plugin::TraceOutput:Output on the https://github.com/rdkcentral/rdkservices repository.

The idea is to extract the `information->Module()` info from the `Trace::ITrace* information` object and add to the syslog line, with the following anatomy:

`[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>:<data>`

How do you feel about this? Would this be more acceptable?

pwielders:
> Hello, @MFransen69 , @pwielders . Took me a while, but, I've found an alternative for this. It is to path the following two files:
> 
> * Source/tracing/Logging.cpp:SysLog() on this repository
> * TraceControl/TraceOutput.h:WPEFramework::Plugin::TraceOutput:Output on the https://github.com/rdkcentral/rdkservices repository.
> 
> The idea is to extract the `information->Module()` info from the `Trace::ITrace* information` object and add to the syslog line, with the following anatomy:
> 
> `[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>:<data>`
> 
> How do you feel about this? Would this be more acceptable?

I thnk that is indeed what I am proposing. The module information is already in all the whole flow, it is just not exposed at the last step. See the remark from two weeks ago where I looked up where the module name is dropped (at the end) https://github.com/rdkcentral/rdkservices/blob/sprint/22Q4/TraceControl/TraceControl.h#L284 where it is still available, but not forwarded to the publishers here https://github.com/rdkcentral/rdkservices/blob/sprint/22Q4/TraceControl/TraceControl.cpp#L236 The proposal is indeed to add the module name on this publsiher interface and than any publisher (e.g. the syslog publisher) can than add it to the output.

paulo-a-junior:
Ok, now I get it (_it takes some time for one to understand a complex package such as WPEFramework and surroundings_ :smiling_face_with_tear: )

I'd have to adapt the ITraceMedia::Output signature [here](https://github.com/rdkcentral/Thunder/blob/93e1f16a222b80234a93343439cb946fb85d9e53/Source/tracing/ITraceMedia.h#L30) to include the module name, explicitly. 

How would you propose the signature to look?

~~Considering existing APIs, one possibility is to have it like `void Output(const char fileName[], const uint32_t lineNumber, const char className[], const ITrace* information, const char moduleName[] = _T(""))`~~

Considering the overall signature, I think this one would be preferred `void Output(const char moduleName[], const char fileName[], const uint32_t lineNumber, const char className[], const ITrace* information)`

Edit:
Although I thought of defining a default argument value for the `const char moduleName`, this will yield no benefit, as the derived classes would have to adapt also their signature. 

This means also changes to the concrete classes, instead of only the method calls, as I first thought of.

Concrete classes that would be affected would be TraceMedia [here](https://github.com/rdkcentral/Thunder/blob/93e1f16a222b80234a93343439cb946fb85d9e53/Source/tracing/TraceMedia.h#L98) and on the `TraceMedia.cpp`. In the context of rdkservices, the TraceOutput class, [here](https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L46).

pwielders:
> _it takes some time for one to understand a complex package such as WPEFramework and surroundings_

That is why we *always* suggest to first reach out to one of the Metrological team members who are more experienced with this product with the issue so we can guide you :-)

On the interface: Yes indeed that is the interface I suggest to change indeed and all implementations need to be adapted, however, and thats why the first line above is already applicable, good that you reach out cause I think the information is already available, see https://github.com/rdkcentral/Thunder/blob/93e1f16a222b80234a93343439cb946fb85d9e53/Source/tracing/ITraceControl.h#L43-L44 Seems like the ITrace interface, passed through this interface, already has the Category()  and Module() name, so I guess the publishers do get the information already. So I guess the only change you have to make is wite it to the syslog here: https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L55 by adding information->Module() and you are done. So the whol change will probably only require adding a parameters (and  formatter parameter) in these lines:
https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L52
https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L55
https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L62
https://github.com/rdkcentral/rdkservices/blob/ad2635316dbf0b62197eadd5d881976355ee0dc3/TraceControl/TraceOutput.h#L65
with the moduleName that you can extact via calling information->Module()
and than *alll* traces will publish the moduleName as well.....

paulo-a-junior:
Yes, I also saw that `ITrace` already included the module name. I was a bit confused about why would the interface require changes as well.

So, to wrap up, the suggestion is just to change the TraceOutput.h, right?

I also would suggest changing the SysLog function  [here](https://github.com/rdkcentral/Thunder/blob/d57aace34672278dce1bd5e3a569711ff8299076/Source/tracing/Logging.cpp#L60), so that all the logs, coming from WPEFramework itself and from its plugins, would have the same anatomy:
`[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>:<data>`

pwielders:
> Yes, I also saw that `ITrace` already included the module name. I was a bit confused about why would the interface require changes as well.
> 
> So, to wrap up, the suggestion is just to change the TraceOutput.h, right?
> 
> I also would suggest changing the SysLog function [here](https://github.com/rdkcentral/Thunder/blob/d57aace34672278dce1bd5e3a569711ff8299076/Source/tracing/Logging.cpp#L60), so that all the logs, coming from WPEFramework itself and from its plugins, would have the same anatomy: `[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>:<data>`

From an architectural perspective, Thunder makes a distinction between logging an tracing. 
1) Logging is for the operator audience, it indicates issues that are expected but can be resolved by taking action. E.g. diskfull, action: Remove files..
2) Tracing is for developer audience, it shows the flow through the code to help debug the system.

Logging is on by default, tracing is off by default, Tracing could event be removed from the comilation in case you select build_type =production to make teh code even smaller. Logging can not be removed from the code base by selecting a different build_type.

Now Logging has a *fixed* module name, it is always the same, in contrast to tracing. If for clarity you want to add the ModuleName for Logging to Logging statements that is fine by me but please consider that the LOGGING and the TRACING are for different audiences. 

Just as a heads-up: In R4, the mechanisms for both Tracing and Logging will be the same as for WarningReporting, third flow of non-functional information, so there the mechanism is the same but the type (LOGGING/TRACING/WARNING_REPORTING) is indicated by an enum. It could be that the internals might change that the ModuleName is nolonger send for LOGGING messages but than you have the enum. SO all this will require a different implementation on R4.

pwielders:
> So, to wrap up, the suggestion is just to change the TraceOutput.h, right?

Missed to answer the question :-) Correct !

paulo-a-junior:
Created two other PRs to supersede this one:

* https://github.com/rdkcentral/Thunder/pull/1178
* https://github.com/rdkcentral/rdkservices/pull/3555

PR #1176, opened 2022-12-20
[DESTRUCTION] Delay submission of the Library associated with a servi…

pwielders:
…ce till the service is fully destructed.

If the destruction of a service takes some time (e.g. more than a few time slices) the previous setup would already drop the Library reference at the begining of the destruct sequence of the service while the destruction still had to take place. At the end this could cause the Library unloading take place prior to the full destruction of the service (e.g. wait for a thread to be completed in the service destruction), which in turn could lead to the segmentation error as the code was removed while it was executing. Now the submission of the Library reference is done *after* the destruction is completed.


HaseenaSainul (approved):
LGTM

PR #1178, opened 2022-12-21
Logging: Add module name to logs

paulo-a-junior:
The current implementation of the TraceControl plugin won't output the origin of the trace.

With this commit, the resulting log message will have the form: 

`[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>: <data>`

This would allow for easier handling of log messages in a scenario where multiple plugins have their tracing flag enabled at the same time.

This change is targeting only the non-abbreviated form of the logs, as the understanding is that adding the module name would make it "less abbreviated".

To capture syslog messages for a given module, use `journalctl |grep <module-name>`.

### DIFF
--- a/Source/tracing/Logging.cpp
+++ b/Source/tracing/Logging.cpp
@@ -65,11 +65,11 @@ namespace Logging {
 #ifndef __WINDOWS__
         if (_syslogging == true) {
             string time(now.ToRFC1123(true));
-            syslog(LOG_NOTICE, "[%s]:[%s:%d]: %s: %s\n", time.c_str(), Core::FileNameOnly(fileName), lineNumber, information->Category(), information->Data());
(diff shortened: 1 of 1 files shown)


pwielders (approved):
See you only add the ModuleName if the trace output is directed to the syslog and not to the console. Are you sure that it is only needed for Syslog, and not for Console?

paulo-a-junior:
I've done that way because the console log was already too much different the syslog line, by only outputting the timestamp and the log data, dropping both file name and line number. 

Do you think it might be a good time to normalize that?

On the rdkservice PR, on the other hand, I've added it to both the console and syslog.

pwielders:
On R4, we will align/normalize it, for R2, I prefer it, but if not, it is fine as well..

paulo-a-junior:
Hello,
I've just pushed the changes for the console log as well =)

karuna2git (approved):

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3520

PR #1179, opened 2022-12-21
Controller: Allow configuring if the UI is enabled or not

rjpdasilva:
Currently, the Controller's UI, serving as base for the Thunder UI, is enabled by default, without allowing a configuration option to disable it.

Since having the Controller's UI enabled requires Web Servicing (accepting and responding to HTTP requests) and that might bring security concerns, and also considering that the UI is mostly for development/debug, add a way to configure the Controller to have the UI disabled.

This way, add a cmake boolean option (`PLUGIN_CONTROLLER_UI_ENABLED`) which, when set to `false`, disables the Controller's UI. Set the new option as `true` by default, enabling the UI by default, for keeping backwards compatibility.

Tested by:
* Confirming the JSON generation for the Controller plugin succeeds in both the cases of Thunder UI enabled and disabled.
* Confirming the Thunder UI is still enabled by default.
* Confirming that, when the Thunder UI is disabled, even though the web server still responds (it is expected), it never actually tries to serve a file by using gdb and confirming that `PluginHost::Server::Service::Evaluate()`, which is the method that processes the requests, will never call `PluginHost::Server::Service::FileToServe()` in this case, and this is the only method that could actually leak some file out.

RDKDEV-458

### DIFF
--- a/Source/WPEFramework/Controller.cpp
+++ b/Source/WPEFramework/Controller.cpp
@@ -111,7 +111,11 @@ namespace Plugin {
 
         _service->Register(&_systemInfoReport);
 
-        _service->EnableWebServer(_T("UI"), EMPTY_STRING);
+        if ((config.Ui.IsSet() == true) && (config.Ui.Value() == true)) {
+            _service->EnableWebServer(_T("UI"), EMPTY_STRING);
+        } else {
+            _service->DisableWebServer();
+        }
 
         // On succes return a name as a Callsign to be used in the URL, after the "service"prefix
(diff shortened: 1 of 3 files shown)


pwielders (approved):
Good functionality. Good implementation. Will track it to move to R4 as well.

pradeeptakdas:
Internal Ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3538

karuna2git (approved):

PR #1198, opened 2023-01-16
[LEAKS] Add  notification if a potential leak (dengling proxy) is det…

pwielders:
…ected.

Going through the code, it was observed that in the non-happy day scenario, proxies of processes that close abnormally mght never be released by the crashing process. Examples of these are notification sinks that are registered by the out-of-process part in the plugin (the Thudner main process).
In such  case, the proxy created in the Thunder process, on behalf of the sink offered for registration by the out-of-process part, is never cleared (maintains a refcount of 1 and thus would leak.
Now if the Trace::Activity in the application, is turned on these "potential" leaks are sned to the trace output. If observed, please check if they can properly destructed if not lready done.

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -215,7 +215,7 @@ namespace PluginHost
     {
         _adminLock.Lock();
 
-        std::map<const string, Core::ProxyType<Service>>::iterator index(_services.begin());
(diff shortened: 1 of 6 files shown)


MFransen69 (dismissed):

volkan-aslan (dismissed):
LGTM

Divya-563 (approved):

PR #1199, opened 2023-01-17
[COM] Unify byte order on the announce message, introduce hresult

sebaszm:


Source/WPEFramework/PluginServer.h
```diff
@@ -961,7 +961,7 @@ namespace PluginHost {
             virtual uint8_t Minor() const override;
             virtual uint8_t Patch() const override;
 
-            uint32_t Submit(const uint32_t id, const Core::ProxyType<Core::JSON::IElement>& response) override;
+            Core::hresult Submit(const uint32_t id, const Core::ProxyType<Core::JSON::IElement>& response) override;
```
pwielders: Why is the Submit, which due to the fact that it has objects, can not be a COMRPC call, have as a return value Core::hresult ?

Source/com/Messages.h
```diff
@@ -190,16 +190,25 @@ namespace RPC {
 
         class Init {
         private:
-            uint32_t ParentId() const 
+            uint32_t ParentId() const
             {
                 uint32_t exchangeId = 0;
-                string value; Core::SystemInfo::GetEnvironment(_T("COM_PARENT_EXCHANGE_ID"), value);
+                string value;
```
pwielders: Can we merge the pieces of information into 1 Environment variable, to avoid a large range of environment variables and to keep associated data in 1 place. Suggest `COM_PARENT_INFO=<ExchangeId>,<Callsign>` ?

Source/plugins/IShell.h
```diff
@@ -259,16 +259,16 @@ namespace PluginHost {
...
-        virtual uint32_t Hibernate(const reason) = 0;
+        virtual Core::hresult Activate(const reason) = 0;
+        virtual Core::hresult Deactivate(const reason) = 0;
+        virtual Core::hresult Unavailable(const reason) = 0;
+        virtual Core::hresult Hibernate(const reason) = 0;
         virtual reason Reason() const = 0;
 
         // Method to access, in the main process space, the channel factory to submit JSON objects to be send.
         // This method will return a error if it is NOT in the main process.
         /* @stubgen:stub */
```
pwielders: As a follow up on the inital remark, I think this could be omit? We can not cross process boundaries with this object anyway..

Source/core/Portability.h
```diff
@@ -736,7 +737,7 @@ namespace Core {
     struct EXTERNAL IReferenceCounted {
         virtual ~IReferenceCounted() = default;
         virtual void AddRef() const = 0;
-        virtual uint32_t Release() const = 0;
+        virtual hresult Release() const = 0;
```
pwielders: Maybe start a skype discussion on this. Wonder how we can do this with a ProxyType? Maybe we can keep this an uint32_t if it fails, we can always return a specific Core::Error code. If this fails we have a potential memory leak...

pwielders (commented):
Lets do a skype chat ;-)

Source/com/Messages.h
sebaszm: Agree.

Source/plugins/IShell.h
sebaszm: We need the stub to be created so the we don't end up with an abstract class instantiation.

Source/core/Portability.h
sebaszm: As discussed, a bit at the crossroads, but let's leave it as uint32_t.

Source/WPEFramework/PluginServer.h
sebaszm: Keeping uint32_t here; not to give a false idea that Submit() can be successful COM call.

sebaszm:
Combined exchange ID and callsign into one environment variable, reverted IReferenceCounted::Release() and IShell::Submit() to uint32_t, but left the latter as stubgen::stub to keep having the stub generated.

pwielders (approved):

PR #1201, opened 2023-01-19
Deleting repetitive condition

VeithMetro:


Source/WPEFramework/Controller.cpp
```diff
@@ -301,7 +301,7 @@ namespace Plugin {
 
         ASSERT(_pluginServer != nullptr);
 
-        if ((basecallsign.empty() == false) && (basecallsign.empty() == false) && (basecallsign != controllerName) && (newcallsign != controllerName)) {
+        if ((basecallsign.empty() == false) && (basecallsign != controllerName) && (newcallsign != controllerName)) {
```
sebaszm: I believe the idea was to make sure that neither basecallsign nor newcallsign is an empty string.

sebaszm (changes requested):

sebaszm (approved):

PR #1205, opened 2023-01-23
[Core] Removing memrcpy implementation and calling memmove instead

VeithMetro:
Benchmarking proved that memrcpy is about 5 times slower than both memmove and memcpy.

### DIFF
--- a/Source/core/Portability.cpp
+++ b/Source/core/Portability.cpp
@@ -191,19 +191,6 @@ uint32_t GetCallStack(const ThreadId threadId, void* addresses[], const uint32_t
 
 #endif // BACKTRACE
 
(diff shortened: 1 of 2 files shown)


sebaszm (commented):
Why leave out EXTERNAL? I suppose this is also DEPRECATED?

pwielders (changes requested):
Agree with @sebaszm . The EXTERNAL is needed (it is not deprecated) as this, in a build, identifies the function as being visible from outside of the compilation unit. The EXTERNAL might seem superfluous as this is a Header file only metod, but it does indicate the intent of this function, it is visible for the outside world.

pwielders (approved):

PR #1208, opened 2023-01-26
Changes for dumping requests on workerpool and pending on signal

madanagopalt:


pwielders:
@madanagopalt really love the feature, just think that in this form it is a bit too limited. Major concern is that solution only targets JSONRPC requests while the Workerpool can execute much more (COMRPC, jobs from plugins etc). That is also the reason why I think that this call: https://github.com/rdkcentral/Thunder/pull/1208/files#diff-4b1e1b742592ba42e2b6a6f5bbf2ccdac6a7d947b11b7f0f704388ac8b19370dR3632 is very scary and potentially might lead to segmentations faults as only a small percentage of the job types in the workerpool is indeed derived from the Channel::Job.
We will pick up the requirement and realize the feature for *all* potential jobs present in the workerpool. Will keep you posted.

pwielders:
Just created a new pull. See: https://github.com/rdkcentral/Thunder/pull/1209

The path where the dump is made, is using the already existing PostMortumPath(), See: https://github.com/rdkcentral/Thunder/pull/1209/files#diff-4b1e1b742592ba42e2b6a6f5bbf2ccdac6a7d947b11b7f0f704388ac8b19370dR4144 
So dump location is different, compared to this pull but now all diagnostic info ends up on the same location.

The JSONRPC call to get this information was already available, it is this one: https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/ControllerJsonRpc.cpp#L47
So the JSONRPC call to execute is different as in this pull, but as this already was available, reused it.

The file contents are JSON objects pushed to file, so the layout might be different compared to this pull request.

The JSONRPC job description, tried to mimic it as good as possible, compared to this pull, but as explained, wit the 1209 pull request we are trying to target all jobs, not only JSONRPC hence why I added the type of Job, easier to later on "read" the file. The metadata/job info for JSONRPC jobs can be found and modified here:
https://github.com/rdkcentral/Thunder/pull/1209/files#diff-4b1e1b742592ba42e2b6a6f5bbf2ccdac6a7d947b11b7f0f704388ac8b19370dR3382-R3391
For COMRPC jobs here:
https://github.com/rdkcentral/Thunder/pull/1209/files#diff-6f6452176a9404fdbaf124522cfcbcd009cb3467c359120a8841f05c20078072L315-R319

Potentially you might need to update more Job types to get more distinct information like here:
https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/blob/eb54220754a3267ebd940fd17cc3735a634d410c/LocationSync/LocationService.h#L46
If you grep for Core::ThreadPool::JobType< you can find all types of Jobs that are submitted to the WorkerPool for which you might need a more descriptive name. You can do so on these jobs by adding a method (no need to make it virtual, it is using SFINAE ti find it) with the following signature: _**string JobIdentifier() const**_ (See: https://github.com/rdkcentral/Thunder/blob/master/Source/core/ThreadPool.h#L346-L359)

PR #1213, opened 2023-01-30
Added ERROR_INPROC, timeout from request, logs and comments

adrianM27:


Source/WPEFramework/PluginServer.cpp
```diff
@@ -701,7 +702,10 @@ namespace PluginHost
                 result = Core::ERROR_NONE;
                 #endif
                 if (result == Core::ERROR_NONE) {
+                    const string className(PluginHost::Service::Configuration().ClassName.Value());
```
pwielders: As you are now in the class Service (which inherits from Plugin::Service) There is no need to dive in the config. The service already has convenience methods to return Classname and Callsign, see: https://github.com/rdkcentral/Thunder/blob/master/Source/plugins/Service.h#L192-L199
As potentially this might change in the future, suggest not to copy it;s content but just use these convenience methods.. So you can just code:
`SYSLOG(Logging::Notification, ("Hibernated plugin [%s]:[%s]", ClassName().c_str(), Callsign().c_str()))`
and drop all the temporary variables.

Source/WPEFramework/PluginServer.cpp
```diff
@@ -739,7 +743,10 @@ namespace PluginHost
...
```
pwielders: See remark above..

Source/WPEFramework/json/JsonData_Controller.h
```diff
@@ -211,6 +211,23 @@ namespace JsonData {
             Core::JSON::Boolean Active; // Denotes whether the subsystem is active (true)
         }; // class SubsystemsParamsData
 
+        class HibernateParamsInfo : public Core::JSON::Container {
```
pwielders, line 214: Not sure if this always works... This is the old way of working but this file is generated during the build from here: https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/json/Controller.json but loooks like this is not part of the generation today, for which I will raise an internall error :-)
aLSO SEE THESE LINES IN THIS FILE: https://github.com/adrianM27/Thunder/blob/development/Hibernation/Source/WPEFramework/json/JsonData_Controller.h#L21
And indeed it was turned off. I will raise a ticket so it gets instantiated again (this line) so just leave this remark for now.. (https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/CMakeLists.txt#L25)

Source/WPEFramework/PluginServer.cpp
adrianM27: gotcha, changed per your suggestion

PR #1218, opened 2023-02-01
Added Memcr libcheckpoint linked

adrianM27:


Source/WPEFramework/PluginServer.cpp
```diff
@@ -507,16 +511,26 @@ namespace PluginHost
...
             const string callSign(PluginHost::Service::Configuration().Callsign.Value());
 
             _reason = why;
 
+            if(currentState == IShell::state::HIBERNATED)
+            {
+                Unlock();
+                if(Wakeup() != Core::ERROR_NONE)
+                {
+                    //TODO: should we force termination?
```
adrianM27, line 529: What if Wakeup failed? Shouldn't we force to kill the plugin's process?

Source/hibernate/checkpointlib/CheckpointLib.c
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#define MODULE "CheckpointLib"
+
+#include "../hibernate.h"
+#include "../common/Log.h"
+
+//TODO: #include <memcr.h>
```
adrianM27, line 24: memcr lib has to be still aligned and changed on top of the latest changes, here only skeleton to call certain memcr functions

PR #1220, opened 2023-02-01
Creating GitHub Action for building Thunder

VeithMetro:
Automatically building Thunder in Debug, Release and MinSizeRel (Production) versions when anyone creates a PR or tries to push into master. The whole process is done concurrently and takes about 2 minutes in total. If the build is unsuccessful, there will be a notification just like in case of other checks (e.g. license/cla).


pwielders (approved):
Well you are the expert on these actions, so I guess what you deliver is oke by definition :-)

PR #1239, opened 2023-02-15
[TYPE] Add a new type (ConCatString -> ccstring) for strings in COMRP…

pwielders:
…C that might, due to concatenation, exceed 64Kb.


Source/plugins/IDispatcher.h
```diff
@@ -38,16 +38,16 @@ namespace WPEFramework {
...
+                virtual Core::hresult Response(const uint32_t channel, const uint32_t id, const ccstring& response) = 0;
 
                 virtual Core::hresult Subscribe(const uint32_t channel, const string& event, const string& designator) = 0;
                 virtual Core::hresult Unsubscribe(const uint32_t channel, const string& event, const string& designator) = 0;
             };
 
-            virtual Core::hresult Validate(const string& token, const string& method, const string& paramaters) const = 0;
-            virtual Core::hresult Invoke(ICallback* callback, const uint32_t channelId, const uint32_t id, const string& token, const string& method, const string& parameters, string& response /* @out */) = 0;
+            virtual Core::hresult Validate(const string& token, const string& method, const ccstring& paramaters) const = 0;
+            virtual Core::hresult Invoke(ICallback* callback, const uint32_t channelId, const uint32_t id, const string& token, const string& method, const string& ccparameters, string& ccresponse /* @out */) = 0;
```
sebaszm: ccstring here!

sebaszm (changes requested):

Source/plugins/IDispatcher.h
pwielders: For 2 reasons:
1) Nice test for the Tools that require adaptions.
2) This is for linking up multiple WPEFrameworks through COMRPC, than the next step is that an attempt will be made to forward the receiving JSONRPC through COMRPC to the other WPEFramework instance and I think.... given the RDKSevices and how the interface design is in those components, within a day we get a call that the JSONRPC can not be forwarded..

At the end, the COMRPC (Proxy Stubs) are only used if we hookup two WPEFrameworks and even than it is, under normal conditions just an additional 4 bytes (2 for parameters, 2 for response)  we need to send, so I guess it is not noticeable in any part of the Metrological system deployments..

MFransen69 (dismissed):

sebaszm (approved):

PR #1252, opened 2023-02-23
[HIBERNATION] If we endup on the controller for JSONRPC, make sure th…

pwielders:
…e actual plugin reports hibernation..


Source/WPEFramework/Controller.cpp
```diff
@@ -776,19 +776,26 @@ namespace Plugin {
             result = _pluginServer->Services().FromIdentifier(callsign, service);
 
             if (result == Core::ERROR_NONE) {
-                ASSERT(service.IsValid());
```
MFransen69, line 779: Think this ASSERT should stay here

MFransen69 (changes requested):

MFransen69 (approved):

PR #1256, opened 2023-02-27
Controller JSONRPC implementation generation based on IControllerExt interface added

HaseenaSainul:


Source/WPEFramework/Controller.cpp
```diff
@@ -804,5 +812,375 @@ namespace Plugin {
...
+    {
+        uint32_t result = Core::ERROR_NONE;
+        return result;
+    }
+    uint32_t Controller::Configuration(const string& callsign, const string& configuration)
+    {
+        uint32_t result = Core::ERROR_NONE;
+        return result;
+    }
+#endif
```
sebaszm: Dead code....?

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
 
         virtual Core::hresult Clone(const string& basecallsign, const string& newcallsign) = 0;
     };
+} // PluginHost
 
-} // namespace PluginHost
-} // namespace WPEFramework
+namespace Exchange {
+    // @stubgen:omit
+    // @json @uncompliant:extended
```
sebaszm, line 67: Do we really want to promote the uncompliant JSON-RPC mode for Controller API?

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
 
         virtual Core::hresult Clone(const string& basecallsign, const string& newcallsign) = 0;
     };
+} // PluginHost
 
-} // namespace PluginHost
-} // namespace WPEFramework
+namespace Exchange {
+    // @stubgen:omit
```
sebaszm: Why disable stubs generation?

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
...
+    // @json @uncompliant:extended
+    struct EXTERNAL IControllerExt : public virtual Core::IUnknown {
```
sebaszm: It seems to me that this could be merged with IController. Using the `@name` and `@alt` tags to provide backwards compatibility at JSON-RPC level, if needed. Since now we have a C++ interface dedicated to JSON-RPC only. Unless we really don't want to expose the other methods via COM-RPC>

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
...
+        // @property
+        virtual uint32_t Environment(const string& index /* @index */, string& environment /* @out @opaque */ ) const = 0;
+        // @property
+        virtual uint32_t Configuration(const string& callsign /* @index */, string& configuration /* @out @opaque */) const = 0;
+        virtual uint32_t Configuration(const string& callsign /* @index */, const string& configuration /* @opaque */) = 0;
+        // @property
+        virtual uint32_t CallStack(const string& index /* @index */, string& callstack /* @out @opaque */) const = 0;
+        // @property
+        virtual uint32_t Version(string& response /* @out @opaque */) const = 0;
+    };
```
sebaszm, line 141: Some properties got lost in translatation.. Were prefix, latitude, longitude idle time removed on purpose?

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
...
+        virtual uint32_t Activate(const string& callsign) = 0;
+        virtual uint32_t Deactivate(const string& callsign) = 0;
+        virtual uint32_t Unavailable(const string& callsign) = 0;
+        virtual uint32_t Suspend(const string& callsign) = 0;
+        virtual uint32_t Resume(const string& callsign) = 0;
+        virtual uint32_t Hibernate(const string& callsign, const uint32_t timeout) = 0;
+        virtual uint32_t StartDiscovery(const uint8_t& ttl) = 0;
+        virtual uint32_t StoreConfig() = 0;
+        virtual uint32_t Delete(const string& path /* @in @opaque */) = 0;
+        virtual uint32_t Harakiri() = 0;
```
sebaszm: example have a Reboot() c++ method but use `@alt harakiri` to enable an alternative deprecated JSON-RPC method

Source/plugins/IController.h
```diff
@@ -43,7 +46,59 @@ namespace PluginHost {
...
+        virtual uint32_t ProcessInfo(string& response /* @out @opaque */) const = 0;
+        // @property
+        virtual uint32_t Subsystems(string& response /* @out @opaque */) const = 0;
+        // @property
+        virtual uint32_t DiscoveryResults(string& response /* @out @opaque */) const = 0;
+        // @property
+        virtual uint32_t Environment(const string& index /* @index */, string& environment /* @out @opaque */ ) const = 0;
+        // @property
+        virtual uint32_t Configuration(const string& callsign /* @index */, string& configuration /* @out @opaque */) const = 0;
+        virtual uint32_t Configuration(const string& callsign /* @index */, const string& configuration /* @opaque */) = 0;
```
sebaszm: All documentation is lost :(

sebaszm (changes requested):

Source/plugins/IController.h
HaseenaSainul: Since this is used for only JSONRPC interface

Source/WPEFramework/Controller.cpp
HaseenaSainul: sorry, will remove

Source/plugins/IController.h
HaseenaSainul: it is added to get the exact format for configuration setting case

Source/plugins/IController.h
HaseenaSainul: @pwielders, already suggested to split the interfaces logically. So looking to that

Source/plugins/IController.h
HaseenaSainul: I have referred : https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/ControllerJsonRpc.cpp#L61-L81 these items. I could not see it here

Source/plugins/IController.h
HaseenaSainul: let me check

Source/plugins/IController.h
HaseenaSainul: will add

Source/plugins/IController.h
sebaszm: Fair enough, I suppose the implementation takes precedence here.

Source/plugins/IController.h
sebaszm: Perhaps it is acceptable to break the backward compatibility for this one call? @pwielders?

Source/plugins/IController.h
HaseenaSainul: added for StoreConfig and Harakiri

Source/plugins/IController.h
pwielders: What is the method, maybe it is better to ask for forgiveness that for permission ;-)

sebaszm: Property for setting `configuration`, being compliant with JSON-RPC it needs a parameter name for the configuration object, however previously it did not.

pwielders (dismissed):
Mainly focused on the interfaces, They look very good so no remarks there. @sebaszm I think we should keep the uncompliant mode on Configurations stuff as it is used to intensively already. We need a lot of forgiveness if we would push it :-)

sebaszm (dismissed):
Great that only IConfiguration needs to be uncompliant, the others are nicely compliant.

pwielders (approved):

PR #1262, opened 2023-03-03
[Actions] Reviewing tests in Thunder and enabling automatic testing

VeithMetro:
Went through all tests available in Thunder, had to disabled a bit of them, since some were either not building/linking, or simply failing. JIRA ticket will be made with a list of tests that need to be fixed. On the other hand, some tests that were disabled could've been enabled again, since they seem to be working properly. Besides, this is a separate action cuz it needs to run on Ubuntu 20.04 as most of our tests are not updated to work with the newest version of libgtest.

### DIFF
--- a/Tests/unit/core/test_message_dispatcher.cpp
+++ b/Tests/unit/core/test_message_dispatcher.cpp
@@ -251,7 +251,7 @@ namespace Tests {
     }
 
     //doorbell (socket) is not quite working inside test suite
-    TEST_F(Core_MessageDispatcher, ReaderShouldWaitUntillRingBells)
+    TEST_F(Core_MessageDispatcher, DISABLED_ReaderShouldWaitUntillRingBells)
     {
         auto lambdaFunc = [this](IPTestAdministrator& testAdmin) {
             Core::MessageDispatcherType<METADATA_SIZE, DATA_SIZE> dispatcher(this->_identifier, this->_instanceId, false, this->_basePath);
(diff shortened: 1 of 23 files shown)


Tests/unit/core/test_message_dispatcher.cpp
```diff
@@ -251,7 +251,7 @@ namespace Tests {
     }
 
     //doorbell (socket) is not quite working inside test suite
-    TEST_F(Core_MessageDispatcher, DISABLED_ReaderShouldWaitUntillRingBells)
+    TEST_F(Core_MessageDispatcher, ReaderShouldWaitUntillRingBells)
```
HaseenaSainul: It seems you have disable test_message_dispatcher.cpp in the cmakelist. Is this due to enabling these tests ?

VeithMetro: Good catch, but no, it was because test_message_dispatcher wasn't building at all:
```
/home/runner/work/Thunder/Thunder/Thunder/Tests/unit/core/test_message_dispatcher.cpp:71:31: error: ‘MessageDispatcherType’ is not a member of ‘WPEFramework::Core’
   71 |         std::unique_ptr<Core::MessageDispatcherType<METADATA_SIZE, DATA_SIZE>> _dispatcher;
      |                               ^~~~~~~~~~~~~~~~~~~~~
```
No need to enable this test tho, you're right, will keep it disabled.

HaseenaSainul (approved):

PR #1268, opened 2023-03-08
[SocketPort] Close half-way open sockets

sebaszm:


pwielders (approved):
This probably resolves a nasty bug I have been chasing for months but was/is hard to reproduce!

PR #1270, opened 2023-03-10
Updating syslog of dynamic COMRPC/configs when disabled

VeithMetro:
It scared both me and Sebastian that something has failed, whereas in reality it is simply disabled (not configured), so let's not scare others trying to build and run Thunder after the upcoming readme update.


sebaszm:
The only thing that scares me is a fridge without beer.

sebaszm (changes requested):
If the path is empty the observer can never be valid. You don't need two conditions to come to this conclusion.

sebaszm (approved):

pwielders (approved):

PR #1272, opened 2023-03-11
ThunderTools: path variable added for Tools cmake path

HaseenaSainul:


bramoosterhuis (approved):

bramoosterhuis:
Approved since it doesn't break other code, that is the rationale behind this?

HaseenaSainul:
yes, in yocto build tools are installing is sysroot-native path, since this is native tool. So we cannot use sys root for both tools and thunder in OE build

PR #1282, opened 2023-03-20
[COM] Frame anti-tampering security

sebaszm:


Source/com/Communicator.h
```diff
@@ -1698,16 +1719,31 @@ POP_WARNING()
 
                 _announceMessage->Parameters().Set(Core::ProcessInfo().Id(), INTERFACE::ID, instance_cast<void*>(offer), Data::Init::REVOKE);
 
+                Core::ProxyType<Core::IPCChannel> baseChannel(*this);
+                ASSERT(baseChannel.IsValid() == true);
+
+                const RPC::InstanceRecord localInstances[] = { { RPC::instance_cast(offer), INTERFACE::ID }, { 0, 0 } };
```
pwielders, line 1725: If you do this, which is correct since the Revoke, might act upon this interface and he did not during the offer, should we than not also at the end check for a CACHED_ADDREF ? 
Or do we want to "forcefully" not allow the Revoke() to do anything with this interface and than not temporary add it here as well ?

pwielders (commented):
See question ;-)

Source/com/Communicator.h
sebaszm: The idea was to give them a chance of Release()-ing and do anything else during the callback. Not sure what they would want to accomplish with AddRef()-ing on a Revoke...?

pwielders: Agree, but if they had done an AddRef during the Offer, it would by in the ProxyAdministration cache and thus would the local cache not be nessecary and it would prevent them from doing an AddRef  in the Revoke ?
If the did not do an Addref on the Offer and do the Addref on the Revoke (for god knows what reason, but they could) what would happen than ?

sebaszm: Well, they could also try to Release on Offer. Hence during Offer Release is a nop and during Revoke AddRef is a nop. In debug it will assert, so this unconventional behaviour is not completely silently ignored.

And local cache is necessary in Offer because they could access the pointer asynchronously (that does currently happen in Compositor server)

sebaszm: Ok, perhaps there is something to it. They could theoretically Addref the pointer on Revoke and asynchronously do an action on it (synchronous action will lock up I think) and only then Release.

pwielders: Agree that theoretically they could do a Release on the Offer but than we can argue that one is violating reference counted lifetime rules, you can only Release, what you addref'ed. Since the offer is the first time you get the interface it could not yet have been Addref'ed. For the Revoke, you can argue that you are not allowed to AddRef an interface you get.
So I think the optimization on the Offer is justifiable by reference counting lifetime rules, I think the optimization on the Revoke can be, using the official rules, by error prone....

pwielders:
Oops I see you already addressed, it but still a good discussion !

pwielders (approved):

PR #1283, opened 2023-03-21
Development/type

pwielders:


sebaszm:
Thout of doing the same, but code like this will fail
https://github.com/rdkcentral/Thunder/blob/7cf788e27697b76f178fa2acbd9d03ea30ef966f/Source/core/Frame.h#L676-L688

pwielders:
Hmmm thought about going "short trough the curve" and leave testing up to someone else, how wrong could this go.... But I think you are right. Guess I need to fix this wit explicit template specialization. Will do so.

sebaszm (dismissed):

sebaszm (approved):

PR #1291, opened 2023-03-24
Hibernate: Support hibernate and wakeup of multi-process plugins

adrianM27:


Source/WPEFramework/PluginServer.cpp
```diff
@@ -681,8 +681,41 @@ namespace PluginHost
             }
             else {
                 #ifdef HIBERNATE_SUPPORT_ENABLED
-                result = HibernateProcess(timeout, local->ParentPID(), _administrator.Configuration().HibernateLocator().c_str(), _T(""), &_hibernateStorage);
-                #else
+                std::map<string, Core::process_t> pidsWithName;
```
pwielders: I think what you are looking for here is:
Core::ProcessInfo::Iterator  children (ParentPid); and then you can use the children to iterate over all the processes that are spawned by the parent pid. 
Is that the idea? or are is it missing something ?

Source/WPEFramework/PluginServer.cpp
```diff
@@ -680,11 +683,24 @@ namespace PluginHost
                 result = Core::ERROR_BAD_REQUEST;
             }
             else {
-                #ifdef HIBERNATE_SUPPORT_ENABLED
```
pwielders, line 683: @adrianM27 sorry for the late remark but for Thunder we have coding guidelines. IN these coding guidelines it is a rule, for reason of maintainability,  that it is only allowed for 1 return in a method. 
Could we reschuffle this method and the next a bit to only have 1 return statement in the method ? Thanks!

pwielders (commented):
Code looks oke, just a small coding guidilen/ rule violation.
I can share the coding guidelines if you need!

Source/WPEFramework/PluginServer.cpp
adrianM27: Not a problem, aligned per your suggestion

pwielders (approved):
@adrianM27  Thanks. Very good contribution!

PR #1304, opened 2023-04-13
Use unlink instead of remove for MessageDataBuffer in removal sequence

HaseenaSainul:


Source/core/FileSystem.h
```diff
@@ -370,6 +370,21 @@ namespace Core {
...
+        {
+            bool result = true;
+
+            if (Exists() == true) {
+                Close();
+#ifdef __POSIX__
+                result = (unlink(_name.c_str()) == 0);
+#endif
+#ifdef __WINDOWS__
+                result = (::DeleteFile(_name.c_str()) != FALSE);
```
pwielders, line 383: Think it is better to put an ASSERT here as this is also on windows not UNLINK functionality but realy deletes it all.. If used on windows we can find out what the logic should be.

pwielders (changes requested):

Source/core/FileSystem.h
HaseenaSainul: as per the DeleteFile documentation, it do the unlink the file and the file deletion does not occur until the last handle to the file is closed. So expect it also behave same as Posix unlink. In this case do we need ASSERT ?

pwielders (approved):

PR #1305, opened 2023-04-18
<RDK-40121>[THUNDER] Self Healing Phase 1

Dosakaya:
Changes Done:
1. Backtrace of crashed thread
2. Logging the Json RPC Request string
3. Logging the Register information
4. Added the crash monitor handler under THUNDER_CRASH_HANDLER profile.

Note:
To enable/disable this feature, we need to set CRASH_MONITOR=ON/OFF in meta-rdk-video/recipes-extended/rdkservices/wpeframework_1.0.bb file.

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -162,6 +162,56 @@ namespace PluginHost
         const string _controllerName;
     };
 
+#ifdef THUNDER_CRASH_HANDLER
+    void Server::WorkerPoolImplementation::UpdateRequestsList(Core::IDispatch* dispatchRequestJob,  const string& jsonString)
+    {
+       _adminLock.Lock();
+       _requestStringList.emplace_back(dispatchRequestJob, jsonString);
+       _adminLock.Unlock();
+    }
+
+    void Server::WorkerPoolImplementation::Dispatcher::SetCrashMonitor(Core::ThreadPool::ICrashMonitor* crashMonitor)
+    {
+        ASSERT(NULL != crashMonitor);
+        ASSERT(NULL == _crashMonitor);
+        _crashMonitor = crashMonitor;
+    }
+
+    void Server::WorkerPoolImplementation::Dispatcher::StoreRequestString(Core::IDispatch* job)
+    {
+        _requestListLock.Lock();
+        auto ptr = _requestList.cbegin();
+
+        while (ptr != _requestList.cend())
+        {
+            if(ptr->iDispatchJob == job)
+            {
+                if (!ptr->requestString.empty())
+                {
+                    ASSERT(NULL != _crashMonitor);
+
+                    if (_crashMonitor)
+                    {
+                        _crashMonitor->StoreRequestString(Core::Thread::ThreadId(), ptr->requestString);
+                    }
+                    else
+                    {
+ #ifndef __WINDOWS__
+                        syslog(LOG_NOTICE, "_crashMonitor is NULL");
+ #endif
+                    }
+                }
+
+                ptr = _requestList.erase(ptr);
+                break;
+            }
+            ptr++;
+        }
+
+        _requestListLock.Unlock();
+    }
+#endif /* THUNDER_CRASH_HANDLER */
+
     void Server::WorkerPoolImplementation::Dispatcher::Dispatch(Core::IDispatch* job) /* override */ {
     #ifdef __CORE_EXCEPTION_CATCHING__
         string callsign(_T("Unknown"));
@@ -173,6 +223,9 @@ namespace PluginHost
         WARNING_REPORTING_THREAD_SETCALLSIGN(callsign.c_str());
 
         try {
+#ifdef THUNDER_CRASH_HANDLER
+            StoreRequestString(job);
+#endif /* THUNDER_CRASH_HANDLER */
             job->Dispatch();
         }
         catch (const std::exception& type) {
@@ -182,6 +235,9 @@ namespace PluginHost
             Logging::DumpException(_T("Unknown"));
         }
     #else
+#ifdef THUNDER_CRASH_HANDLER
+        StoreRequestString(job);
+#endif /* THUNDER_CRASH_HANDLER */
         job->Dispatch();
     #endif
     }
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -34,6 +34,10 @@
 #error "Please define the name of the COM process!!!"
 #endif
 
+#ifdef THUNDER_CRASH_HANDLER
+#define MAX_LENGTH_JSON_REQUEST_PARAM 200 /* maximum length of json request param to store */
+#endif
+
 #define MAX_EXTERNAL_WAITS 2000 /* Wait for 2 Seconds */
 
 namespace WPEFramework {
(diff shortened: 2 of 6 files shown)


pwielders:
@Dosakaya / @anand-ky : The title sounds like an interesting feature, can I get the design document (CR ?) for this? I assume it is temporary ? It is dropped if you move to R4.
If needed in R4, please supply the use-case/CR!

anand-ky:
@pwielders - Here is the document - https://ccp.sys.comcast.net/secure/attachment/5601122/ThunderCrashHandlerPoC%20v39.pdf
This is tracked under https://ccp.sys.comcast.net/browse/RDK-40121

Source/WPEFramework/PluginServer.cpp
```diff
@@ -162,6 +162,56 @@ namespace PluginHost
         const string _controllerName;
     };
 
+#ifdef THUNDER_CRASH_HANDLER
+    void Server::WorkerPoolImplementation::UpdateRequestsList(Core::IDispatch* dispatchRequestJob,  const string& jsonString)
+    {
+       _adminLock.Lock();
+       _requestStringList.emplace_back(dispatchRequestJob, jsonString);
```
pwielders, line 169: As I do not have the design documentation but looking at the inserted code, I think this is a *very* dangerous action. The Core::IDispatch is not a reference counted object. The workerpool relies on the submitter to revoke the Core::IDispatch *before* destructing it. I do not see the  _requestStringList being updated in the Remove, so the IDispatch might be a dead object by the time you use it.

Source/WPEFramework/PluginServer.cpp
```diff
@@ -162,6 +162,56 @@ namespace PluginHost
...
+       _adminLock.Unlock();
+    }
+
+    void Server::WorkerPoolImplementation::Dispatcher::SetCrashMonitor(Core::ThreadPool::ICrashMonitor* crashMonitor)
+    {
+        ASSERT(NULL != crashMonitor);
+        ASSERT(NULL == _crashMonitor);
+        _crashMonitor = crashMonitor;
```
pwielders, line 177: According to the asserts this is an a-symetrical function. I suggest to:\
ASSERT ((nullptr != crashMonitor) ^ (nullptr != _crashMonitor));
so you can also remove the crashMonitor through this interface. 
Also NULL is a C pre-processor statement, this is a C++ file, use the c++ nullptr for null checking

Source/WPEFramework/PluginServer.h
```diff
@@ -34,6 +34,10 @@
 #error "Please define the name of the COM process!!!"
 #endif
 
+#ifdef THUNDER_CRASH_HANDLER
+#define MAX_LENGTH_JSON_REQUEST_PARAM 200 /* maximum length of json request param to store */
+#endif
```
pwielders, line 39: Preprocessor statements are from the old days :-) 
The disadvantage of preprocessor statements are that the do not use namespacing and thus might conflict easilly. Why not use a proper const definition within the PluginServer class namespace ?

Source/WPEFramework/PluginServer.h
```diff
@@ -103,27 +107,79 @@ namespace PluginHost {
...
+                string requestString;
+
+                requestStrListStruct (Core::IDispatch* Job,  const std::string &rString)
+                {
+                    iDispatchJob = Job;
+                    requestString = rString;
+                }
+            };
+
+            typedef std::list<requestStrListStruct> RequestStrList;
```
pwielders, line 123: 1) Risk of the iDispatchJob being dead. I did not find any cleanup upon revoke of the job.
2) std::vector for performance is applicable here. Do not thinkg you really need an std::list

Source/WPEFramework/PluginServer.h
```diff
@@ -103,27 +107,79 @@ namespace PluginHost {
...
             WorkerPoolImplementation& operator=(const WorkerPoolImplementation&) = delete;
 
+#ifdef THUNDER_CRASH_HANDLER
+            WorkerPoolImplementation(const uint32_t stackSize)
+                : Core::WorkerPool(THREADPOOL_COUNT, stackSize, 16, &_dispatch)
+                , _adminLock()
+                , _dispatch(_requestStringList, _adminLock)
+            {
+                CreateCrashMonitor();
+               _dispatch.SetCrashMonitor(GetThreadPool().GetCrashMonitor());
```
pwielders, line 174: _requestStrinList is, I think not in the initializer list?

Source/WPEFramework/PluginServer.h
```diff
@@ -2513,6 +2575,21 @@ namespace PluginHost {
             ~Channel() override;
 
         public:
+#ifdef THUNDER_CRASH_HANDLER
+            void GetRequestString(const Core::JSONRPC::Message &message, std::string& jsonRPCString)
```
pwielders, line 2579: The concept behind thinder it to have a portable piece of code. WIth the change of a build flag we can turn in from char <->wchar. to do so we made a Portability.h where the strin is declared. Depending on the buildflag string -> std::string or std::wstring. By declaring std::string this feature is broken. Never use the std::string but always use string.

Source/WPEFramework/PluginServer.h
```diff
@@ -2740,7 +2817,18 @@ namespace PluginHost {
...
                             job->Set(Id(), &_parent, service, baseRequest, _security->Token(), !request->ServiceCall());
+#ifdef THUNDER_CRASH_HANDLER
+                            std::string strJsonRpcRequest("");
+                            /* write the json rpc request in new string varaible */
+                            if(!request->ServiceCall() && (baseRequest->HasBody() == true))
+                            {
+                                Core::JSONRPC::Message &message(*(baseRequest->Body<Core::JSONRPC::Message>()));
+                                GetRequestString(message, strJsonRpcRequest);
+                            }
+                            _parent.Submit(Core::proxy_cast<Core::IDispatchType<void>>(job), strJsonRpcRequest);
```
pwielders, line 2828: ???? Oke now I am intersted in the design :-) the workerpool is agnostic (and should be agnostic) to the type of work it is conducting. The scary feeling I get is that the generic job agnostic functionality of the Workerpool is becoming a JSONRPC aware job handler.

Source/core/ThreadPool.cpp
```diff
@@ -0,0 +1,234 @@
...
+        void ThreadPool::CrashMonitor::DeleteCrashMonitor()
+        {
+            if (NULL != _workerThreadInfo)
+            {
+                delete[] _workerThreadInfo;
+                _workerThreadInfo = NULL;
+            }
+        }
+
+        void ThreadPool::CrashMonitor::DumpBacktrace()
```
pwielders, line 110: This functionality is already in the THunder core. Do not repeat!! See: https://github.com/rdkcentral/Thunder/blob/R2/Source/core/Portability.cpp#L209-L244

Source/WPEFramework/PluginServer.h
pwielders: Digging down, looks like ot is limited to the PluginServer, there is an additional method to submit. A bit lost why it would than not be part of this callstack, but I do think that on R4, we redesign this as I see a lot of code duplication and I   seriously doubt the lifetime management of the IDispatch jobs held in the different lists!

pwielders (commented):
Not suitable for R4. Code duplication and potentially lifetime issues on the jobs held in the lists..
If needed in R4, suggest to create a CR describing this feature and suggest that it is picked up in collaboration with the Thunder Development team.

Source/WPEFramework/PluginServer.cpp
pwielders: As I am not familiar with the intend of the functionality at hand here but I am fairly knowledgeable on the workerpool/threadpool, I would like to explain that if the submitter of the Job called: https://github.com/rdkcentral/Thunder/blob/R2/Source/core/WorkerPool.h#L198-L207 which in turn on the ThreadPool issues https://github.com/rdkcentral/Thunder/blob/R2/Source/core/ThreadPool.h#L351-L369 the submitter is free to "delete" the dispatchRequestJob held in this queue. I can not find any code in these methods that take care of cleaning up the _requestStringList. Maybe the pointer is only used for comparison to find the proper JSONRPC string and the fact that it is deleted it will probably not really matter as it is just a key. But extensions in the future might be using it for the Job name and than we have a crash for sure :-)

pwielders:
From the code I can see that the feature is probably intended to output JSONRPC messages in case of an exception/crash occurs during the handling of this JSONRPC request.
To do so, I think, a list is build of all currently scheduled jobs on the Workerpool where the IDispatch is probably used as the key. In case of the crash the exception handler in the Workerpool start looking in the queue to find the actual JSONRPC request and output that for POSTMORTEM analyses.
If that is the case, it means that all JSONRPC requests coming in and are submitted on the workerpool, on forehand create a long string with the full JSON message store it in the list (not searchable), in case of a crash it is searched and reported, if no crash (hopefully 99.9999% of the message) has to find it again and remove it form this list.

Have we thought about **TLS** ? Which is already available as an abstraction and is already used for the Callsign ?

Where we could store the JSONRPC::Message and only if there is a crash (hopefully 0.00000001% of the message) Do all the difficult work without all the lists and string formatting on forehand.

Maybe good if we pickup this feature for R4 ?

PR #1312, opened 2023-05-09
[Messaging] Separating Logging and Tracing, unifying code formatting

VeithMetro:
Initial changes to Messaging, along with changes to the MessageControl plugin (in ThunderInterfaces and ThunderNanoServicesRDK).

### DIFF
--- a/Source/core/MessageStore.cpp
+++ b/Source/core/MessageStore.cpp
@@ -255,18 +255,15 @@ namespace Core {
             }
         }
 
-        /* static */ void IControl::Revoke(IControl* control)
-        {
+        /* static */ void IControl::Revoke(IControl* control) {
             _registeredControls.Revoke(control);
         }
 
-        /* static */ void IControl::Iterate(IControl::IHandler& handler)
-        {
+        /* static */ void IControl::Iterate(IControl::IHandler& handler) {
             _registeredControls.Iterate(handler);
         }
 
-        /* static */ IStore* IStore::Instance()
-        {
+        /* static */ IStore* IStore::Instance() {
             return (_storage);
         }
         
@@ -277,4 +274,4 @@ namespace Core {
         }
     } // namespace Messaging
 } // namespace Core
-} // namespace WPEFramework
+}
--- a/Source/core/MessageStore.cpp
+++ b/Source/core/MessageStore.cpp
@@ -92,7 +92,6 @@ namespace Core {
 
         const char* MODULE_LOGGING = _T("SysLog");
         const char* MODULE_REPORTING = _T("Reporting");
-        const char* MODULE_INVALID = _T("Invalid");
 
         uint16_t Metadata::Serialize(uint8_t buffer[], const uint16_t bufferSize) const
         {
@@ -130,7 +129,6 @@ namespace Core {
             if (bufferSize > (sizeof(_type) + (sizeof(_category[0]) * 2))) {
                 Core::FrameType<0> frame(const_cast<uint8_t*>(buffer), bufferSize, bufferSize);
                 Core::FrameType<0>::Reader frameReader(frame, 0);
-                _type = frameReader.Number<type>();
                 _category = frameReader.NullTerminatedText();
 
                 if (_type == TRACING) {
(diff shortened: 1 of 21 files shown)


Source/core/MessageStore.cpp
```diff
@@ -85,19 +87,32 @@ namespace WPEFramework {
     }
 
 namespace Core {
+
     namespace Messaging {
 
+        const char* MODULE_LOGGING = _T("SysLog");
+        const char* MODULE_REPORTING = _T("Reporting");
+        const char* MODULE_INVALID = _T("Invalid");
```
sebaszm: Remove Invalid, this is developer-error situation, no need to cater for it other than with an ASSERT.

Source/messaging/MessageClient.h
```diff
@@ -70,5 +71,6 @@ namespace Messaging {
         Clients _clients;
         Factories _factories;
     };
-}
+
+} // namespace WPEFramework
```
sebaszm: Don't add this.

Source/messaging/MessageClient.cpp
```diff
@@ -159,9 +160,11 @@ namespace Messaging {
      *
      * @param function function to be called on each of the messages in the buffer
      */
-    void MessageClient::PopMessagesAndCall(std::function<void(const Core::Messaging::IStore::Information& info, const Core::ProxyType<Core::Messaging::IEvent>& message)> function)
+    void MessageClient::PopMessagesAndCall(std::function<void(const Core::Messaging::Metadata& metadata, const Core::ProxyType<Core::Messaging::IEvent>& message)> function)
     {
-        Core::Messaging::IStore::Information information;
+        Core::Messaging::Metadata metadata;
+        Core::Messaging::IStore::Logging log;
+        Core::Messaging::IStore::Tracing trace;
```
sebaszm: Initialize variables only when needed.

Source/messaging/MessageClient.cpp
```diff
@@ -176,21 +179,46 @@ namespace Messaging {
                     size = sizeof(_readBuffer);
                 }
 
-                auto length = information.Deserialize(_readBuffer, size);
+                if (metadata.Deserialize(_readBuffer, size) != 0) {
+
+                    if (metadata.Type() == Core::Messaging::Metadata::type::TRACING) {
+                        auto length = trace.Deserialize(_readBuffer, size);
```
sebaszm: Seems the same data is deserialized again!

sebaszm: What is the type of the length variable?

Source/messaging/MessageClient.cpp
```diff
@@ -176,21 +179,46 @@ namespace Messaging {
...
+                            if (factory != _factories.end()) {
+                                message = factory->second->Create();
+                                message->Deserialize(_readBuffer + length, size - length);
+                                function(log, message);
+                            }
+                        }
+                        else {
+                            client.second.FlushDataBuffer();
+                        }
+                        size = sizeof(_readBuffer);
```
sebaszm: Looks like same code is repeated, move to a function.

Source/core/MessageStore.h
```diff
@@ -25,10 +25,15 @@
...
 
 namespace WPEFramework {
+
 namespace Core {
 
     namespace Messaging {
 
+        extern EXTERNAL const char* MODULE_LOGGING;
+        extern EXTERNAL const char* MODULE_REPORTING;
+        extern EXTERNAL const char* MODULE_INVALID;
```
sebaszm: Why are these still needed externally?

sebaszm (changes requested):
See comments.
Please do not adjust the formatting of the code if not necessary, it obstructs the important changes! :)

Source/messaging/MessageClient.cpp
VeithMetro: Don't we need to first Deserialize() with just Metadata, as we don't know if that's Tracing or Logging type message in the buffer, and after we do that, then Deserialize again, but this time on a real object?
Alternatively to deserializing this part of the data twice, we could remove the call of Metadata::Deserialize() from Logging/Tracing::Deserialize(), and then after knowing the type, just read the extra stuff like timeStamp, file, class and line
/edit: Resolved in a private conversation - now only the Type is check instead of Deserializing the whole Metadata

VeithMetro: It used to be auto before my changes, but it is simply uint16_t, so I guess there is no point to keep it as auto. Should I just change it to uint16_t?

sebaszm: yes please

Source/messaging/MessageClient.cpp
VeithMetro: Is it more efficient to initialize these variables multiple times inside the loop, but only when necessary, rather than just doing it once at the start?

Source/core/MessageStore.h
VeithMetro: Just tried it without the EXTERNAL macro and I'm getting such buildroot error:
`[ 85%] Linking CXX executable WPEProcess`
`/home/veith/buildroot/output/host/lib/gcc/arm-buildroot-linux-gnueabihf/9.2.0/../../../../arm-buildroot-linux-gnueabihf/bin/ld: ../messaging/libWPEFrameworkMessaging.so.1.0.0: undefined reference to WPEFramework::Core::Messaging::MODULE_LOGGING`

sebaszm: Ok since this is used during control creation as well, let's leave it as it is. Perhaps we will address this another time.

sebaszm (approved):

PR #1314, opened 2023-05-11
[POSTMORTEM] Enhance the dumping of information as required by the Th…

pwielders:
…under Crash Handler POC [RDK-40121]

From a Thunder perspective this is a usefull part of this POC. Self-healing, as also explained in the meetings, is an undesirable feature as it break more than it solves. Recovering in a process is unwanted because:
1) Thunder has the concept of out-0of-process. This is for instable or security sensitive plugins.
   Use this feature if you cannot solve the root-cause, develop stable plugins. A process restart resolves all following objections that come with the described self-healing feature.

2) Attempting to self heal the WPEFramework process will push the cause untrackable issues later on because:
   - memory leakages will occure for objects that have been alllocated (but are not freed due to the self healing)
   - synchronisation object might casue deadlock later on (very hard to track the root-cause than) because the self-healing process can not revert locked synchronisation objects.
   - Files might get corrupted as they might have partially written, Self healing can not close these and recover these.
   - State changes on which other components might rely are never send and thus other components might assume the incorrect state as the change might have occured but the notification was not yet end.

Self healing might suggets fast recovery but as explained in the call, the deadlocking mutex might frustrate the end0user even more as the system becomes than irresponsive for a undefined time.

Fault finding with the spamming RDK Service plugins becomes a nightmare as you need to first check if a self-healing process occured to place the latest crash in the right perspective....

So self healing is a bad idea!

### DIFF
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -266,8 +266,11 @@ POP_WARNING()
 
     ExitHandler* ExitHandler::_instance = nullptr;
     Core::CriticalSection ExitHandler::_adminLock;
+    
+    #ifndef __WINDOWS__
     struct sigaction _originalSegmentationHandler;
     struct sigaction _originalAbortHandler;
+    #endif
 
     static string GetDeviceId(PluginHost::Server* dispatcher)
     {
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -266,6 +266,7 @@ POP_WARNING()
 
     ExitHandler* ExitHandler::_instance = nullptr;
     Core::CriticalSection ExitHandler::_adminLock;
+    struct sigaction _originalSegmentationHandler;
 
     static string GetDeviceId(PluginHost::Server* dispatcher)
     {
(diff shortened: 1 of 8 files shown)


Source/WPEFramework/PluginHost.cpp
```diff
@@ -316,6 +317,24 @@ POP_WARNING()
 
             ExitHandler::StartShutdown();
         }
+        else if (signo == SIGSEGV) {
+
+            // From here on we do the best we can do. Have no clue what failed, try to log as much as possible and on
+            // a subsequent segmentation fault, just handle it the old fashion way. The root cause has been logged
+            // by than!
+            sigaction(SIGSEGV, &_originalSegmentationHandler, nullptr);
```
MFransen69, line 329: Just thinking wouldn't it be better to turn this and the next line around? I don't know if on Linux if we get here the process is frozen but on Windows I think that is not the case and a different thread could cause an access violation then trigger the original handler and just abort the process before the first one was logged. Or does this work completely different on Linux?

Source/WPEFramework/PostMortem.h
```diff
@@ -0,0 +1,152 @@
...
+            , Module(copy.Module)
+            , Line(copy.Line)
+        {
+            Add(_T("address"), &Address);
+            Add(_T("function"), &Function);
+            Add(_T("module"), &Module);
+            Add(_T("line"), &Line);
+        }
+
+        ~CallstackData() = default;
```
MFransen69: Nitpicking but for consistency it should be an override just like in Callstack below :)

MFransen69 (changes requested):

Source/WPEFramework/PluginHost.cpp
pwielders: Reinstalling the basic handler is mandatory before we proceed, otherwise we could get recursively called over and over again (creating an endless loop) So swapping with the next line could cause serious issues.

Source/WPEFramework/PostMortem.h
pwielders: agree. Fixed

MFransen69 (dismissed):

MFransen69 (approved):

anand-ky (approved):

PR #1317, opened 2023-06-05
[MessageExchange] Fix deadlock during flushing

sebaszm:
It has been observed that ocassionally write on serial port can lock up with flush. This hangs BluetoothControl and thus Thunder startup.

SerialPort::Write takes SerialPort mutex (A), waits on MessageExchange mutex (B) via SendData
MessageExchange::Flush takes MessageExchange mutex (B), waits on SerialPort mutex (A) via SocketPort::Flush


pwielders (approved):
Thats a nasty bug to tackle/catch, I do see the deadlock, but what if, between the flush and the lock a new character comes in (so it is not really flushed) would that be oke? As you are deeper into this, than I am, going to approve it and assume you thought about that situation ;-) If not think about it and if oke merge and otherwise fix :-)

pwielders (approved):
Thats a nasty bug to tackle/catch, I do see the deadlock, but what if, between the flush and the lock a new character comes in (so it is not really flushed) would that be oke? As you are deeper into this, than I am, going to approve it and assume you thought about that situation ;-) If not think about it and if oke merge and otherwise fix :-)

PR #1328, opened 2023-06-14
[COMRPC] Adding multiple COMRPC Servers need to cope each with hteir …

pwielders:
…own AnnounceMessage..

This is a flaw in the implementation that surfaces now as now we have a multitude of COMRPC servers if a plugin has a dedicated COMRPC port.


Source/com/Administrator.h
```diff
@@ -298,27 +298,34 @@ namespace RPC {
...
         }
-        Job(Core::IPCChannel& channel, const Core::ProxyType<Core::IIPC>& message, Core::IIPCServer* handler)
+        Job(Core::IPCChannel& channel, const Core::ProxyType<Core::IIPC>& message)
             : _message(message)
             , _channel(channel)
-            , _handler(handler)
+        {
+        }
+        Job(Job&& move)
+            : _message(move._message)
```
MFransen69: I think you want a std:move here as well (as in the next line)

Source/com/Communicator.h
```diff
@@ -317,6 +317,23 @@ namespace RPC {
         virtual Core::instance_id ParentPID() const = 0;
     };
 
+    class EXTERNAL InvokeHandler : public Core::IIPCServer {
+    public:
+        InvokeHandler(InvokeHandler&&) = delete;
+        InvokeHandler(const InvokeHandler&) = delete;
+        InvokeHandler& operator=(const InvokeHandler&) = delete;
```
MFransen69, line 325: Not that it is really required but we discussed to in these cases also delete the move operator (also below a few times)

MFransen69 (changes requested):

MFransen69 (approved):

PR #1331, opened 2023-06-16
RDK-41134: Improve Thunder crash logging.

Dosakaya:
Created patch on master thunder.

1. Registered ExitDaemonHandler for SIGSEGV and SIGABRT, just before Dispatch.
2. Restored Google breakpad handler for SIGSEGV and SIGABRT after dumping data.
3. Changed Dumping File name from ThunderInternals.json to ThunderInternals_<time>.json

Note:
ThunderInternals_<time>.json file path is /opt/minidumps/ .

### DIFF
--- a/Source/WPEFramework/PluginHost.cpp
+++ b/Source/WPEFramework/PluginHost.cpp
@@ -18,6 +18,8 @@
  */
 
 #include "PluginServer.h"
+#include "PluginHost.h"
+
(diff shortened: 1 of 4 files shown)


woutermeek:
Please explain the problem which this solves. This approach is to resource consuming and violates architectural dependencies.

Please let us know what the problem is and we will help solve it.

PR #1337, opened 2023-06-23
Add a new deactivation reason "MemoryExceededOnSuspend"

varumugam123:
This is added to decide whether to discard a plugin (like WebKitBrowser) if it takes more memory on suspended state

### DIFF
--- a/Source/plugins/IShell.h
+++ b/Source/plugins/IShell.h
@@ -72,6 +72,7 @@ namespace PluginHost {
             AUTOMATIC,
             FAILURE,
             MEMORY_EXCEEDED,
+            MEMORY_EXCEEDED_ON_SUSPEND,
             STARTUP,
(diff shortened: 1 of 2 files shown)


pwielders (commented):
As the suspend is a volatile state, how can a race situation be prevented?
Also since you are coming from resumed and by design/architecture, the resumed state is the state in which you hold on to all your resources and all object you need to be able to be active, this state should always use more or best case, as much as the suspended situation.
What use-case are you trying to resolve here? Will you create a new monitor that monitors memory usage in suspend and in resumed state ?
Is there a Ticket number to read more info about this, seemingly, confusing error state ?

varumugam123:
This requirement is obsolete now. Closing the PR. Thanks @pwielders for sharing your insights!

PR #1340, opened 2023-06-28
[Docs] Add documentation for plugin execution mode

TeknoVenus:
Document running plugins in/out of process, including container and distributed mode


docs/assets/simple_execution_modes.drawio.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
woutermeek, line 1: Random thought: Maybe split the yellow  OoO plugins box into two boxes to avoid that people think it is a single process for all OoO plugins? I know it shows 2 WPEProcesses, but it might make it more clear.

docs/plugin/execution-modes.md
```diff
@@ -0,0 +1,117 @@
...
+      "root":{
+         "mode":"Local"
+      }
+   }
+}
+```
+
+## In Process Plugins
+
+This is the simplest execution mode, and the default if a mode is not set. An in-process plugin is loaded into and executed inside the main WPEFraemwork daemon.
```
pwielders: WPEFraemwork daemon -> WPEFramework Process (as the WPEFramework can run in the background as daemon or in the forground, without -b, and it still runs in the WPEFrameworkProcess)

docs/plugin/execution-modes.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Advantages**
+
+* Simple - there is no additional overhead or complexity from running additional processes
+* Highest performance - no additional RPC hops are required to communicate with the plugin. All calls to the plugin from the WPEFramework daemon are just local function calls.
+    * If an in-process plugin needs to communicate with another in-process plugin, Thunder will resolve any COM-RPC interface calls to local function calls. This removes the need to serialise/deserialise any data or cross any IPC boundary and results in the fastest performance possible
```
pwielders, line 41: local function calls, pure technically it is a VPTR dereference and than a local function call, so a "local virtual function call". Don't know if we want to go in so much detail... so whatever you think is oke

docs/plugin/execution-modes.md
```diff
@@ -0,0 +1,117 @@
...
+
+- [LXC ](https://linuxcontainers.org/)
+- [runc](https://github.com/opencontainers/runc)
+- [crun](https://github.com/containers/crun)
+- [Dobby](https://github.com/rdkcentral/Dobby) (RDK)
+- AWC (Liberty Global)
+
+A requirement for running a plugin in a container is a suitable container configuration must have been defined - Thunder cannot create container configurations dynamically on-the-fly for plugins. Thunder will look for container configurations in the following locations (with the below priority):
+
+1. `<volatile path>/<callsign>Container`
```
pwielders: <volatile path>/<callsign>Container -> <volatile path>/<callsign>/Container

docs/plugin/execution-modes.md
```diff
@@ -0,0 +1,117 @@
...
+
+* Maintenance - need to maintain a container config and correctly punch holes in the container to access resources (e.g. device nodes)
+* Startup time - plugin activation time may be increased due to the overhead of constructing the container environment
+
+## Distributed
+
+!!! danger
+	Distributed mode is considered experimental
+
+Distributed mode takes the out-of-process mode one step further by allowing plugins to run on an entirely different device than the main WPEFramework process. This device could even be running a different CPU and with a architecture. A COM-RPC channel is established over a TCP socket between the two devices to allow communication.
```
pwielders: "This device could even be running a different CPU and with a architecture" -> "This device could even be running a different CPU with a different architecture"

mkdocs.yml
```diff
@@ -36,6 +36,7 @@ nav:
     - Subsystems: plugin/subsystems.md
     - ProxyPool: plugin/proxypool.md
     - Private COM-RPC : plugin/private-comrpc.md
+    - Split Implementation: plugin/split-implementation.md
```
pwielders, line 39: Wanted to read this but seems it is till missing. In the preview I got a deadlink..

pwielders (changes requested):

docs/plugin/execution-modes.md
TeknoVenus: Ah good point, will replace daemon with process (and keep in mind for future pages!)

docs/plugin/execution-modes.md
TeknoVenus: Lets go with "local virtual function call" - I think that's enough detail, the main point is that we don't have to do any kind of IPC hop

mkdocs.yml
TeknoVenus: That's because I haven't written that page yet :) 

Plan to base it off this existing page in rdkcentral: https://wiki.rdkcentral.com/pages/viewpage.action?pageId=198262351.

docs/plugin/execution-modes.md
pwielders: agree !

pwielders (approved):

PR #1341, opened 2023-06-28
[Messaging] [Warning reporting] Making warning reporting run through the MessageControl plugin

VeithMetro:


VeithMetro:
@sebaszm I'm not entirely sure what you meant when you wanted to remove `enabled` from the warning reporting's side. As far as I can see, now it is only stored in `WarningReportingControl`, since it inherits from IControl and overrides both of its Enable methods - seems okay?

VeithMetro:
@sebaszm Do we want to go back to `EXTERNAL_HIDDEN`? I don't think it had anything to do with the bounds OOP issue - should I revert it, rebuild and check whether everything still works as intended?

Source/core/MessageStore.h
```diff
@@ -249,6 +278,7 @@ namespace Core {
             virtual bool Default(const Metadata& metadata) const = 0;
             virtual void Push(const Logging& log, const IEvent* message) = 0;
```
pwielders: Thinking out loud and knowing that we are going to get two more streams. I think that for the Push there is no need to explicitly say logging/Tracing or WarningReporting. I guess there is a baseclass for these and on the base class we have a "Length()" and "Serialize()". The implementation might be different but that does not matter (we push a reference) so would it be an opyion to bring this back to 1 method, called:
_virtual void Push(const Metadata& info, const IEvent* message); = 0_ ?

Source/messaging/DirectOutput.h
```diff
@@ -50,6 +50,7 @@ namespace WPEFramework {
             
             void Output(const Core::Messaging::IStore::Logging& log, const Core::Messaging::IEvent* message) const;
```
pwielders: Same as the push, hope we can use a baseclass for this..

sebaszm:
> @sebaszm Do we want to go back to `EXTERNAL_HIDDEN`? I don't think it had anything to do with the bounds OOP issue - should I revert it, rebuild and check whether everything still works as intended?

Yes, use EXTERNAL_HIDDEN.

Source/core/MessageStore.cpp
```diff
@@ -289,6 +335,30 @@ namespace Core {
             return (length);
         }
 
+        string IStore::WarningReporting::ToString(const abbreviate abbreviate) const
+        {
+            string result;
+            const Core::Time now(TimeStamp());
+
+            if (abbreviate == true) {
```
pwielders: Maybe better to compare with the enum value in stead of true/false. Would even expect the compiler to generate a warning for this :-)

Source/core/MessageStore.cpp
```diff
@@ -247,6 +267,32 @@ namespace Core {
             return (length);
         }
 
+        string IStore::Tracing::ToString(const abbreviate abbreviate) const
+        {
+            string result;
+            const Core::Time now(TimeStamp());
+
+            if (abbreviate == true) {
```
pwielders: Maybe better to compare with the enum value in stead of true/false. Would even expect the compiler to generate a warning for this :-)

Source/core/MessageStore.cpp
```diff
@@ -201,6 +201,26 @@ namespace Core {
             return (length);
         }
 
+        string MessageInfo::ToString(const abbreviate abbreviate) const
+        {
+            string result;
+            const Core::Time now(TimeStamp());
+            string time;
+
+            if (abbreviate == true) {
```
pwielders: Maybe better to compare with the enum value in stead of true/false. Would even expect the compiler to generate a warning for this :-)

pwielders (approved):

sebaszm (approved):

PR #1343, opened 2023-07-03
[REDIRECT] Add a template class to redirect stdout/stderr.

pwielders:
The template has been tested under windows (crappy OS) and should be oke for Linux. The template will be instantiated to redirect the STDOUT and the STDERR into a non-functional message stream to be routed to the MessageControl Plugin. From there the STDOUT/STDERR streams can be routed to any desired destination.


Source/messaging/ConsoleStreamRedirect.h
```diff
@@ -0,0 +1,498 @@
...
+					ASSERT(_index == Core::IResource::INVALID);
+					ASSERT(replacing != Core::IResource::INVALID);
+
+					if (replacing != Core::IResource::INVALID) {
+						Core::IResource::handle newDescriptor;
+
+						if (CreateOverlappedPipe(_handle, newDescriptor) != false) {
+							_flushall();
+							_copy = ::_dup(replacing);
+							if (::_dup2(newDescriptor, replacing) == -1) {
```
VeithMetro, line 235: Yet another school day, learned about `_dup`, `_dup2` and file descriptors in general

Source/messaging/ConsoleStreamRedirect.h
```diff
@@ -0,0 +1,498 @@
...
+					, _index(Core::IResource::INVALID)
+					, _copy(Core::IResource::INVALID)
+					, _handle(Core::IResource::INVALID) {
+				}
+				~ReaderImplementation() override {
+					Close();
+				}
+
+			public:
+
```
VeithMetro, line 361: Are these methods below not aliened with the `ReaderImplementation` class on purpose? :smile:

Source/messaging/ConsoleStreamRedirect.h
```diff
@@ -0,0 +1,498 @@
...
+			StandardError(const StandardError&);
+			StandardError& operator=(StandardError&&);
+			StandardError& operator=(const StandardError&);
+
+			StandardError() = default;
+			~StandardError() = default;
+
+		public:
+			void Output(const uint16_t length, const TCHAR buffer[]) {
+				// if (LocalLifetimeType<STREAMTYPE, &Core::System::MODULE_NAME, Core::Messaging::Metadata::type::TRACING>::IsEnabled() == true) {
```
VeithMetro, line 487: These both `Output()` methods are going to have something similar to the macros for Tracing, Logging, Warning Reporting inside, and they will end with a `MessageUnit::Push()` method?

VeithMetro (approved):
A question or two in the comments, but all in all looks very lean and mean, at least for my newbie eyes 😃

Source/messaging/ConsoleStreamRedirect.h
pwielders: Indeed, here we are going to "Push" the actual stderr/stdout to the buffers using the new values in the enum you already added. Does that make sense?

Source/messaging/ConsoleStreamRedirect.h
pwielders: Yes that is on purpose, here you can see that the Linux system has a much cleaner structure than the mess they created in Windows. 
The open and close are different on Windows and Linux hence why there are 2 implementations. In linux the system is consistent. Everything behaves as a file descriptor. Sockets, Files and SerialPorts can be treated in Linux as a filedescriptor with a-synchronous IO calls, that are all monitored by the the *Core*::ResourceMonitor.
On Windows it is a mess and every sub component of Windows has a slightly different API, hence why here (and in for example the SerialPort handling here: https://github.com/rdkcentral/Thunder/blob/master/Source/core/SerialPort.cpp#L82-L300) we use a bit of a different approach to handle.
I do think we could align it by extending the Core::IResource so that all a-synchronous IO on Windows is also done by the ResourceMonitor but as Windows is officially is not supported it is something we need to do if time permits.
And note that although the API used is posix compliant the dup -> (windows version) _dup, as dup2 (_dup2) are slightly deviating from the standard. I think we could have used (on windows) also the posix variant without the starting underscore but it would result in a warning..

Source/messaging/ConsoleStreamRedirect.h
VeithMetro: As long as we are checking if this message type `IsEnabled()`, and then sending a metadata object derived from `MessageInfo` plus the actual message as the parameters of `MessageUnit::Push()`, then it will work just the same as tracing, so it does make sense and looks solid :)

Source/messaging/ConsoleStreamRedirect.h
pwielders: Take care the official, correct Posix calls are without the underscore, this is windows crap ;-)

Source/messaging/ConsoleStreamRedirect.h
pwielders: Indeed the details should still be filled in, but this template should already redirect a filedescriptor (stderr or stdout) to the Output() with every line (ending by /n /n/r or /r/n) and not including the ending chars in an a-synchronous way.

VeithMetro (approved):
Once again, I do approve this amazing template

sebaszm (approved):

PR #1344, opened 2023-07-03
"SGMM393-547/RDKVREFPLT-129  : Including input security token value f…

kponmu736:
…ully to validate signature.   Reason for change: Previously not considered full input security token value to validate signature, first 32 bytes of input security token value only have considered to validate signature, If first 32 bytes are correct, ValidSignature API returned true value, It's giving chance to security breach.   Test Procedure: Enable Thunder Security, Get the current security token using WPEFrameworkSecurityUtility application, Pass different  security token in the Json request to validate secuityAgent.validate API, Unit testing done and working properly..   Risks: None.   Signed-off-by: kottalarathi_ponmuthu@comcast.com"


Source/websocket/JSONWebToken.cpp
```diff
@@ -167,12 +168,33 @@ namespace Web
...
+                memset (signatureOutput, 0, sizeof (signatureOutput));
+
+                decodedLength = Core::URL::Base64Decode(
+                            token.substr(pos + 1).c_str(),
+                            static_cast<uint16_t>(token.length() - pos - 1),
+                            reinterpret_cast<uint8_t*>(signatureOutput),
+                            static_cast<uint16_t>(signatureSource.length() * sizeof(TCHAR)),
+                            nullptr);
+		    
+                if ( decodedLength == sizeof(signature)) {
```
pwielders: Looking at this line, I think that the essence of this PR is to make sure that the signature decoded from Base64 is not longer than required, but I guess you could have depicted that upfront by the line:  (signatureSource.length() == ((((8 * sizeof(signature))+ 8)/6) ?  Why go through the hassle of converting the full Base64 string if you could have calculated the length of the Base64 representation of a SHA256 sum up front ? Thunder always strives to be Lean, mean and resource constrained ;-) I think that this has an impact of 1 additional if, the comparison in the if is a compiletime calculatable number, so neglectable performance impact...

pwielders (changes requested):
If it is indeed only a check to see if the length of  the token in Base64 is correct, it can be done much simpler and thisus requesting this to change to 1 if with an fixed number check..

Source/websocket/JSONWebToken.cpp
kponmu736: Hi,

Greetings ! Thank you very much for your code review comments !

Changed code as below, It's working fine.

	bool JSONWebToken::ValidSignature(const mode type, const string& token) const 
	{
        bool result = false;

        // Check if the Hash is correct..
        size_t pos = token.find_last_of('.');

        if (pos != string::npos) {
		
            if (type == JSONWebToken::mode::SHA256) {
					
                // Now calculate what we think it should be..
                Crypto::SHA256HMAC hash(_key);

                // Extract the signature and convert it to a binary string.
				uint8_t signature[Crypto::SHA256HMAC::Length];
								
				if ((token.length() - pos) != ((((8 * sizeof(signature))+ 8)/6)))
				{
					return false;
				}
				
                if (Core::URL::Base64Decode(token.substr(pos + 1).c_str(), static_cast<uint16_t>(token.length() - pos - 1), signature, sizeof(signature), nullptr) == sizeof(signature)) {

					hash.Input(reinterpret_cast<const uint8_t*>(token.substr(0, pos).c_str()), static_cast<uint16_t>(pos * sizeof(TCHAR)));
					result = (::memcmp(hash.Result(), signature, sizeof(signature)) == 0);
				}
            }
		}

		return (result);
    }


Newly added code is 

				if ((token.length() - pos) != ((((8 * sizeof(signature))+ 8)/6)))
				{
					return false;
				}
				
Here, Value of ((((8 * sizeof(signature))+ 8)/6)) is 44.

Consider this token "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.aHR0cDovL2xvY2FsaG9zdA.doZcrWnknbAEqIAAeyZ9D8_GJ1cT-XoOJkl-Clm39Rk",

length of "doZcrWnknbAEqIAAeyZ9D8_GJ1cT-XoOJkl-Clm39Rk" is 43 only.

Can i do -1 from ((((8 * sizeof(signature))+ 8)/6)) or Can i add . with "doZcrWnknbAEqIAAeyZ9D8_GJ1cT-XoOJkl-Clm39Rk" length.

Please suggest me.

pwielders: Sorry there is a rounding error in the example, it should be ((((8 * sizeof(signature))+ 5)/6)). This results in ((8 * 32) + 5) / 6 = 43. The +5 is to compensate for the fact that if due to the rounding we need 1 additional byte, it would result in ((5+1) = 6)/6 => 1 additional byte. This full byte is needed as the division will just drop off any remainder, but we need to align on full bytes..
Also for THunder we have some coding guidelines that do not allow for multiple returns in a method (Maintainability) so could we create the code like:

```
// Now calculate what we think it should be..
Crypto::SHA256HMAC hash(_key);

// Extract the signature and convert it to a binary string.
uint8_t signature[Crypto::SHA256HMAC::Length];
							
if ((token.length() - pos) == ((((8 * sizeof(signature)) + 5)/6)))
{
    if (Core::URL::Base64Decode(token.substr(pos + 1).c_str(), static_cast<uint16_t>(token.length() - pos - 1), signature, sizeof(signature), nullptr) == sizeof(signature)) {

        hash.Input(reinterpret_cast<const uint8_t*>(token.substr(0, pos).c_str()), static_cast<uint16_t>(pos * sizeof(TCHAR)));
        result = (::memcmp(hash.Result(), signature, sizeof(signature)) == 0);
    }
}
```
This way we only have 1 return statement and the default value of result was already set to false :-)

kponmu736: Hi,

Greetings ! Thank you very much for your review comments !

Addressed review comments, Added 

                if ((token.length() - pos - 1) == ((((8 * sizeof(signature)) + 5)/6)))
                {

                }
and removed all other newly added lines, Done unit testing, It's working fine.

Please review & merge it.       

Thank you very much once again !

pwielders (approved):

PR #1346, opened 2023-07-04
[Docs] Add documentation on plugin lifecycle

TeknoVenus:


pwielders (approved):
First read looks very good. Suggest to merge and get feedback from people less familiar with the framework.

PR #1349, opened 2023-07-07
minus max size of encoded data

HaseenaSainul:


Source/websocket/WebSocketLink.h
```diff
@@ -534,9 +534,9 @@ POP_WARNING()
 
                 if ((_state & WEBSOCKET) != 0) {
                     if (maxSendSize > 4) {
```
pwielders: I guess the maxSendSize (in comparison, if (maxSendSize > 8) should now also be increased to 8, otherwise we could get negative *and thus very huge numbers for the size in the following lines..

pwielders (approved):

PR #1350, opened 2023-07-07
[Docs] Messaging

VeithMetro:
Adding documentation for messaging that includes tracing, logging and warning reporting with an overview of the MessageControl plugin.


docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
+In the past, logging, tracing and warning reporting were completely separate issues and were handled individually. However, even though these message types have distinct characteristics, we recognized the advantages of consolidating them into a unified framework, which is now referred to as `Messaging`. We are convinced it is much better suited for the development, and with some recent fixes that will be described in the following sections, `Messaging` is fully operational. We strongly believe that it is a good time to shed some more light on that functionality, since it is simple to use and yet very effective.
```
TeknoVenus: - Lets be more specific here and specify Thunder versions where messaging can/can't be used (and note that warning reporting was only recently moving into the messaging system).
- What are the advantages of using the Messaging framework over just `printf()` statements? Why should we use it? We should make the pros/cons very clear in the introduction

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+Before we begin to describe what the `MessageControl` plugin does and where logging, tracing and warning reporting can be directed, it is probably a good idea to start from the ground up. In this case, what precisely are these message types, what are their differences in general and in Thunder, how exactly the messages are created, and what is going on with them before they end up at their final destination.
```
TeknoVenus: This doesn't really add very much information, just says what we're going to do. With clear subheadings/sections this is unnecessary. We can shorten/remove this paragraph

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+We can all agree that logging, tracing and warning reporting are important techniques often used in software development to gather information and provide insight into the behavior of an application. However, there are some fundamental differences between them.
+
```
TeknoVenus, line 6: Add a quick high-level summary of the differences between Tracing/Logging/WarningReporting here (i.e. logging is for production, trace is for developers etc)

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+```c++
+TRACE(Trace::Information, (_T("Not an A2DP audio sink device!")));
+
+TRACE_GLOBAL(Trace::Error, ("Is this not a descriptor to a DRM Node... =^..^= "));
+```
+
+In Thunder, there are two main macros that can be used to wrap messages that will be treated as traces. These macros, namely `TRACE` and `TRACE_GLOBAL`, can be found in `Thunder/Source/messaging/TraceControl.h`. Upon examining the code, we can notice that the only distinction between them is that `TRACE` includes the class name where it is used, whereas the global version refers to the function name instead of a class.
```
TeknoVenus, line 217: Do we have a recommendation on which macro plugin authors should be using?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+```c++
+TRACE_L1("Failed to load library: %s, error %s", filename, _error.c_str());
+```
+
+On a side note, if you browse the Thunder code, you may come across another macro called `TRACE_L1`. You might wonder how it differs from the previously mentioned macros. To understand its purpose, it is important to know that tracing is disabled in certain layers of Thunder. This is necessary because `Messaging` relies on both `COM` and `Core`, and we surely want to avoid circular dependencies. As a result, `TRACE_L1` is restricted to lower layers only. Within the framework, you can utilize it in `Core` and `Messaging`, but it should not be used in `COM`. Moreover, once you reach a certain point, refrain from using `TRACE_L1` altogether, and instead use the `TRACE` macro.
```
TeknoVenus: Move the TRACE_Lx information into its own section.

Think this is important, so add a warning box:
```
!!! warning
    Warning text goes here
```

to make it obvious that TRACE_Lx macros should only be used by the framework for low-level debug messages and not by plugins

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+                __LINE__,
+                WPEFramework::Core::ClassNameOnly(typeid(*this).name()).Text()
+            );
+            WPEFramework::Messaging::TextMessage __message__(__data__.Data());
+            WPEFramework::Messaging::MessageUnit::Instance().Push(__trace__, &__message__);
+        }
+    } while(false)
+```
+
+In the code fragment above, you can observe the internal structure of the macro. While more detailed explanations will be provided in subsequent paragraphs, let us cover the general process briefly. Firstly, we need to verify if the corresponding category is enabled. If it is, we proceed to create metadata for the message and send it alongside the message itself to the plugin, which will be discussed in greater detail in the plugin section. Before continuing, it is also worth noting that there is a way to print the trace messages directly on the console in case the `MessageControl` plugin is disabled. In such a situation, the `Output()` method of the `DirectOutput` class can be used, but this will be described further on in the following section.
```
TeknoVenus: DirectOutput deserves its own section, not mixed in with this since it applies to all messages not just tracing.

Add a new section dedicated to viewing the log output, which can cover the direct output and message control

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+SYSLOG(Logging::Startup, (_T("Failure in setting Key:Value:[%s]:[%s]\n"), index.Current().Key.Value().c_str(), index.Current().Value.Value().c_str()));
+```
+
+This relates to the main difference between the `TRACE` and `SYSLOG` macros in Thunder, which is that `SYSLOG` is present in any build, and `TRACE` is dropped in production, the same way as, for example, an `ASSERT` macro. It concludes the main distinction between these two message categories in Thunder: tracing should be used when we want to indicate a vital information during the development, and we should use logging to record any important data that could be useful in the future.
+
+```c++
+SYSLOG_GLOBAL(Logging::Fatal, (_T("Plugin config file [%s] could not be opened."), file.Name().c_str()));
+```
+
+Similarly as in the case of tracing, there is also a global version of the `SYSLOG` macro, and the difference is exactly the same as with tracing macros. Furthermore, the `DirectOutput` class can be configured using the `Mode()` method to send the messages to the system logger instead of simply printing them on the console.
```
TeknoVenus: Again, do we recommend a specific macro or is it just up to the developers?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+In the piece of code below we can see that the first noticeable difference between tracing and logging macros is that there is an assert which ensures that the macro parameter `CATEGORY` is an actual logging category, so there is no way to create a custom logging category, unlike in the case of tracing or warning reporting. Apart from this difference, everything looks very similar besides the omission of file, line and class name, since these are not particularly useful when it comes to logging.
```
TeknoVenus: The fact we can't create logging categories is a key bit of information - lets call that out in a note section

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+```c++
+REPORT_OUTOFBOUNDS_WARNING(WarningReporting::SinkStillHasReference, _referenceCount);
+```
+
+## Macros
+
+Within Thunder, there are five distinct macros dedicated to warning reporting, each with a slightly different use case scenario. You can find these macros inside `Source/core/WarningReportingControl.h`. The first two macros, `REPORT_WARNING` and `REPORT_WARNING_GLOBAL`, only require the `CATEGORY` parameter. These macros should be used when reporting a warning without without specific values within the category. While all categories in Thunder currently use actual values and compare them with the reporting bounds. While all categories in Thunder currently utilize actual values and compare them against reporting bounds, these macros have been implemented with future use cases in mind. If someone decides to add their own simple category without reporting values, they can effortlessly make use of these macros.
```
TeknoVenus: Each of the 5 macros should get their own heading with description of the macro and an (ideally self-contained) example.

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+            WPEFramework::Messaging::TextMessage __message__(__data__.Data());
+            WPEFramework::Messaging::MessageUnit::Instance().Push(__log__, &__message__);
+        }
+    } while(false)
+```
+
+In addition, you might be wondering why we have some `Messaging` components such as `Metadata` in `Core`, and why everything is not simply inside
+`Source/messaging`. From an architectural point of view, there is a reason for this, which revolves around the need for reporting capabilities within `Core`. For instance, we would like to measure how long it takes to lock and then unlock. To accomplish this, we need some of the messaging features to be accessible in `Core`, because otherwise we would get, as you may have guessed, circular dependencies.
+
+# Warning Reporting
```
TeknoVenus: Something unclear from reading this - all the examples of warning reporting are from inside the framework.

Since this guide is for plugin authors, do we expect warning reporting to be used inside plugin code too? If so, lets make that clear and give an example

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+    if (actualValue > _reportingBound.load(std::memory_order_relaxed)) {
+        report = CallAnalyze(moduleName, identifier, std::forward<Args>(args)...);
+    }
+    return report;
+}
+```
+
+In the code above, we can notice that the main job of this method is to check whether the `actualValue` reported by the macro exceeds the `_reportingBound` set by the user.  If the condition is met, then the `CallAnalyze()` method is called, which will invoke an `Analyze()` method specific to the warning reporting category. However, if the category does not have such a method, then `CallAnalyze()` simply returns `true`. In such fashion, we can easily check additional conditions necessary for a warning to trigger.
+
+## Warning Reporting proxy
```
TeknoVenus: This is very interesting, and good to document but less relevant to a plugin author so move to an "Internals" section

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+## Content of the message
+
+Before examining the workings of the plugin, it is crucial to provide a brief overview of the components that make up a message, as they directly impact the message processing. A message consists of three main parameters: module, category, and the actual content of the message. Understanding these parameters is essential for effective message handling.
+
+### Categories of each message type
+
+When it comes to tracing, the module parameter represents the name of the plugin responsible for sending the message, for example `Plugin_Cobalt`. However, when the `SYSLOG` macro is used, the module name will simply be `Syslog`.
```
TeknoVenus: Be obvious where this module name comes from (it's defined by the plugin in `Module.h` and each plugin has a unique module name)

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+
+Choosing the appropriate category is vital when working with macros. The category can be described as the class that either formats the message string or best represents the nature of the message. While it is possible to create custom categories, which will be explained in the following paragraph, it is generally recommended to utilize the predefined categories available in Thunder to avoid redundancy. Thunder already offers six default categories for tracing:
+
```
TeknoVenus, line 492: Overlap with the configuration documentation here

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+Each of them is created in `Thunder/Source/messaging/TraceCategories.h` using the `DEFINE_MESSAGING_CATEGORY` macro:
+
+```c++
+DEFINE_MESSAGING_CATEGORY(Messaging::BaseCategoryType<Core::Messaging::Metadata::type::TRACING>, Text);
+```
+
+As you can see in the code, these classes inherit from a templated class called `BaseCategoryType`. The association of these categories with tracing is established by passing the `TRACING` metadata type as a template parameter. Similarly, logging includes several categories, such as:
+
+-   Startup
```
TeknoVenus: Overlap with the configuration documentation here. Lets keep configuration in one place (don't mind which) to make things easier to keep up to date moving forward

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+- TooLongDecrypt
+- JobActiveForTooLong
+- TooLongPluginState
+- TooLongInvokeMessage
+
+### Creating a warning reporting category
+
+Unlike the tracing and logging categories, these warning reporting categories are not declared using macros. Instead, they are directly declared within the designated header files mentioned above.
+
+The code snippet below demonstrates the simplicity of creating a custom warning reporting category. If an additional `Analyze()` method is not necessary, it is sufficient to declare the `Serialize()` and `Deserialize()` methods to return `0`. However, attention must be given to the implementation of the `ToString()` method, along with the variables `DefaultWarningBound` and `DefaultReportBound`.
```
TeknoVenus: What's the difference between the warning and report bound? Do warnings/reports get handled differently internally?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,601 @@
...
+        visitor = (_T("It took suspiciously long to acquire a critical section"));
+        visitor += Core::Format(_T(", value %" PRId64 " [ms], max allowed %" PRId64 " [ms]"), actualValue, maxValue);
+    };
+
+    static constexpr uint32_t DefaultWarningBound = { 1000 };
+    static constexpr uint32_t DefaultReportBound = { 1000 };
+};
+```
+
+## From a macro to an output
```
TeknoVenus: Lets move this whole section to an "Internals" section to make it obvious this is just supplementary information that is not necessary required by a plugin author

TeknoVenus (changes requested):
Overall, this is a great start!

This isn't showing in the navigation menu - can you update the `mkdocs.yml` file so your document is correctly indexed? 

I've made some comments, but the majority of my feedback is related to the structure of the documentation. In my eyes, this documentation should serve two purposes:

* Explain to developers how best to use the features
* Document how things work internally

This document does a great job of explaining how messaging works underneath, which is very valuable, but plugin authors aren't necessarily going to care how things work under the hood, but they just want to know how to use them. 

I propose restructuring this document somewhat into (roughly) the below sections. The aim is to make the document easily skimmable by a busy developer, so make use of sub-headings to break things up as much as possible. There are also some sections that are missing today (e.g. how to configure messaging categories at runtime and how to define custom trace events from a plugin)

I also suggest a section at the end of the document called "Internals" where we document how messaging works internally. The rest of the document should focus on how to use it from the perspective of a developer.

* Introductory paragraph
* Difference between tracing/logging/warning reporting (this should just be a quick bit, a few bullets maybe - can expand on more in later sections)
* Viewing logs
  * MessageControl plugin
  * Direct Output
* How to adjust trace/logging/warningreporting levels
  * Configuration
  * Runtime
* Tracing
  * Plugin Tracing (TRACE/TRACE_GLOBAL macros)
  * Internal tracing (TRACE_LX)
  * Configuration
  * Defining custom trace categories for a plugin
* Logging
  * Usage
  * Configuration
* Warning Reporting
  * Usage
  * Configuration
  * Defining custom warning reporting categories
* Internals
  * How messaging works

---

As a general stylistic comment, please try to be as concise as possible and assume anyone reading this document is a very busy person! For example, instead of:

> Even though it is nice to learn some theory of how things work, it is probably best to learn in practice and from examples, so before moving to the next paragraph, let us have a quick look at a piece of code in which the warning reporting macro is actually utilized. The code fragment below can be found in `Source/core/WorkerPool.h`, and the idea behind this reporting is to automatically trigger a warning in case a job has been dispatched for way too long.

I suggest

> Below is an example of using the `REPORT_OUTOFBOUNDS_WARNING_EX` macro taken from `Source/core/WorkerPool.h`. Here a warning is triggered if a job has taken too long to complete (which could indicate a deadlock).

Some references for the writing style that works well for documentation

* Google's guide to writing good technical documentation: https://developers.google.com/tech-writing/one
* The plugin lifecycle documentation: https://rdkcentral.github.io/Thunder/plugin/lifecycle/

docs/plugin/messaging.md
VeithMetro: A valid point, we should underline that warning reporting can and potentially should be used inside the plugin code as well. I'll try to make that more clear and give an example of an actual use from the plugin code.

docs/plugin/messaging.md
VeithMetro: Sounds like a good idea. Let's have a configuration section in this document like you've suggested in your table of contents. I'll move the part of config documentation related to messaging to here, and also change it to match the current changes, since I can see that it is not entirely correct now.

docs/plugin/messaging.md
VeithMetro: Yes, `TRACE` should always be used if we have the `this` pointer available, so when tracing takes place inside a class, otherwise you have to use `TRACE_GLOBAL`. That's a good point though and I'll include that in the documentation, maybe even as a warning.

docs/plugin/messaging.md
VeithMetro: Once again a very valid point, because the `SYSLOG` macro should be used, and we will even mark `SYSLOG_GLOBAL` as deprecated, because that's no longer necessary since the separation of logging and tracing (logging does not have either a class or a function name anymore). I'll make changes to this part of the document.

docs/plugin/messaging.md
VeithMetro: We want the `MessageControl` plugin to handle reports and warnings *separately* - report means that we send the data to be stored, for example, in a file and then analyzed, while warning is more severe, so we want to instantly output it to a different location,  for instance, a console or a network stream. In the end, we are going to split it in the `MessageControl` plugin in such a way that it will be possible to filter out the warnings. I'll mention that in the documentation as well.

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
+In the past, prior to the R3 version of Thunder, logging, tracing and warning reporting were completely separate issues and were handled individually. However, even though these message types have distinct characteristics, we recognized the advantages of consolidating them into a unified framework, which is now referred to as `Messaging`. In the early versions of R3, we still had tracing enabled by default and messaging was in the early development. It is the R4 version of Thunder where `Messaging` is finally the default framework to handle all types of messages.
+
+
+# Advantages of using messaging
```
TeknoVenus: There's a quirk in the mkdocs theme we're using where the table of contents doesn't work if you use H1 (`#`) headings in the documentation.

All headings should be H2 (`##`) or smaller, so can you update to change this? If it's right, you'll see a Table of Contents appear on the right hand side of the page :)

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
...
+
+There are of course several advantages of using this framework to deliver the messages over simply using something like `printf`:
+
+* There are not only different types of messages, but also categories within these types, and each of them can be enabled/disabled at runtime
+* No serialization penalty between different processes
+* Always getting full lines one after another since all messages are timestamped
+* Possibility of redirecting the messages to another output like, e.g., a console, syslog, a file or a network stream, which is particularly useful in the case of containers, and this is something we are working towards
```
TeknoVenus: > and this is something we're working towards

Can you be clear which bit of this bullet we're working towards and which actually exists?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
...
+* In the near future, *standard out* and *standard error* will also be redirectable
+
+We are convinced that `Messaging` is much better suited for the development, and with some recent changes which will be described in the following sections, it is fully operational and enabled by default. We strongly believe that it is a good time to shed some more light on that functionality, since it is simple to use and yet very effective.
+
+
+# Differences between the message types
+
+Logging, tracing and warning reporting are important techniques often used in software development to gather information and provide insight into the behavior of an application. However, there are some fundamental differences between them within Thunder, in a nutshell:
+
+* Tracing is meant for the developers and is dropped in production
```
TeknoVenus, line 19: I think we should be explicit and say "dropped in production by default" as if I understand correctly it is still possible to enable Tracing in production if required?

Also what determines a production build?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
...
+    }
+}
+```
+
+
+# How to adjust messaging
+
+## Configuration
+
+The config file can be used to enable/disable the default messaging categories used for logging, tracing and warning reporting.
```
TeknoVenus: Be explicit and state that this is the Thunder config file in /etc/WPEFramework/config.json, not a plugin config file

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
...
+	```
+
+Warning Reporting enables various runtime checks for potentially erroneous conditions and can be enabled on a per-category basis. These are typically time-based - i.e. a warning will be reported if something exceeded an allowable time. Each category can also have its own configuration to tune the thresholds for triggering the warning.
+
+!!! note
+	Warning Reporting is only available if Thunder is compiled with the `WARNING_REPORTING` option.
+
+## Runtime
+
+It is also possible to use the `MessageControl` plugin to edit the configuration values at runtime. At the moment it is possible to enable/disable any category at runtime, either globally for logging and warning reporting, or individually per plugin for the tracing messages.
```
TeknoVenus: Can we show an example JSON-RPC request to change a messaging level at runtime?

docs/plugin/messaging.md
```diff
@@ -0,0 +1,886 @@
...
+
+By emphasizing the significance of these conditions, warning reporting enables proactive identification and resolution of potential problems, enhancing system reliability and user experience. It acts as an early detection mechanism, signaling the need for attention and potential action before the situation escalates into a critical failure or incident. Overall, warning reporting complements logging and tracing by focusing on identifying and communicating conditions that require immediate attention, facilitating effective troubleshooting and maintenance of software systems.
+
+```c++
+REPORT_OUTOFBOUNDS_WARNING(WarningReporting::SinkStillHasReference, _referenceCount);
```
TeknoVenus, line 348: Is this line meant to be here since we go through all the macros just below it?

TeknoVenus (commented):
Looking good! Only a few minor nitpicks then should be good to merge :)

docs/plugin/messaging.md
VeithMetro: So that's why the table of contents was not showing, it makes sense now. Thanks, I'll update it and remember that for the future docs 😃

docs/plugin/messaging.md
VeithMetro: In `Production`, so when building with the `Min_Size_Rel` flag, the `TRACE` and `TRACE_GLOBAL` macros are declared empty and there is no way to enable tracing. This version of Thunder is meant to be used only by operators who want the smallest footprint on memory possible.

In both `Debug` and `Release` versions, tracing is enabled. `Debug` is used by developers, so all macros are enabled and it has no code optimization at all, whereas `Release` is used by most of our operators and the QA team - asserts off and some code optimization.

I'll make sure to add this to the documentation, since you're right that it's worth to clarify that.

docs/plugin/messaging.md
VeithMetro: Good catch, it's indeed no longer necessary, as I expanded upon this example in the next section, so I'll remove this line.

docs/plugin/messaging.md
VeithMetro: Sounds like yet another great idea, I didn't want to include a screenshot with the ThunderUI, since it seemed to be a bit too basic, but a JSON-RPC call is a great way to showcase that. I'll add this to the documentation.

TeknoVenus: I have an example of documenting a JSON-RPC request here - personally quite like the arrow emojis to make it obvious which is the request and response :) https://rdkcentral.github.io/Thunder/plugin/config/#get-config

TeknoVenus (approved):
Looking good, thanks for your hard work on this!

pwielders (approved):

MFransen69 (approved):

PR #1351, opened 2023-07-07
Development/non happy day jsonrpc

pwielders:


Source/WPEFramework/PluginServer.h
```diff
@@ -3609,37 +3609,61 @@ POP_WARNING()
...
+
+                                // If we also do not have an id, we can not return a suitable JSON message!
+                                if (message->Recorded().IsSet() == false) {
+                                    response->ErrorCode = Web::STATUS_BAD_REQUEST;
+                                    response->Message = _T("JSON-RPC was incorrectly formatted, could not deduce the id");
+                                }
+                                else {
+                                    uint32_t id = message->Recorded().Value();
+                                    string error = message->Validity().Value().Message();
+                                    response->ErrorCode = Web::STATUS_NO_CONTENT;
```
TeknoVenus: 204 is not a suitable response code - as per HTTP spec a 204 response cannot contain a body

>  A 204 response is terminated by the end of the header section; it cannot contain content or trailers.

TeknoVenus (changes requested):
Much better, although there are still some JSON documents that trip it up.

Looking at the examples provided in the JSON-RPC spec I see many of them return ID as null if the JSON document cannot be parsed. I think this is because most JSON parsers don't stream the same way we do, so whilst we can deduce the ID even if subsequent data is malformed, most can't so just return a null ID. 

Do we want to/can we strictly follow the spec here? I like trying to return the ID if we can to make it easier for clients to correlate which message was malformed, but currently we return an empty body if we can't deduce the ID.

Other examples of malformed JSON:

Missing closing brace - still returns 200 OK
```json
{
	"jsonrpc": "2.0",
	"id": 1,
	"method": "Controller.1.status",
	"params": {}
```

Params contain junk - still returns 200 OK.

I guess the issue here is we're treating params as an opaque JSON blob, but I think we should still return an error as the whole document is invalid JSON
```json
{
	"jsonrpc": "2.0",
	"id": 1,
	"method": "Controller.1.status",
	"params": {
		ewfkjewfpowefkewopkefewfewfw
	}
}
```

Requests without an ID still hang. This example was taken from JSON-RPC spec. According to spec we should return 32600 with a null ID value in this scenario.
```json
{
	"jsonrpc": "2.0",
	"method": 1,
	"params": "bar"
}
```

Junk data for the method param will return 202 and an "Unknown Method" error. Whilst technaically correct, compliance to the spec should return a parse error. Not sure which I prefer!
```json
{
	"jsonrpc": "2.0",
	"id": 2,
	"method": junkjunkjuk,
	"params":{}
}
```
(Side note, the HTTP error code on this response is `202 Failure on JSONRPC: 4294934695`, the `4294934695` feels like an overflow?)

Source/WPEFramework/PluginServer.h
```diff
@@ -3609,37 +3609,61 @@ POP_WARNING()
...
+                                    else {
+                                        response = IFactories::Instance().Response();
+                                        response->Body(body);
+                                        if (body->Error.IsSet() == false) {
+                                            response->ErrorCode = Web::STATUS_OK;
+                                            response->Message = _T("JSONRPC executed succesfully");
+                                        }
+                                        else {
+                                            response->ErrorCode = Web::STATUS_ACCEPTED;
+                                            response->Message = _T("Failure on JSONRPC: ") + Core::NumberType<uint32_t>(body->Error.Code).Text();
```
TeknoVenus: Error code here can be negative if it's a JSON-RPC error so will overflow uint32_t

Source/websocket/WebSerializer.h
```diff
@@ -423,5 +430,40 @@ namespace Web {
...
+            Body(Body&&) = delete;
+            Body(const Body&) = delete;
+            Body& operator= (Body&&) = delete;
+            Body& operator= (const Body&) = delete;
+
+            Body() = default;
+            ~Body() override = default;
+
+        public:
+            void Clear() {
```
MFransen69: Not too important but I guess this should also be an override

MFransen69 (changes requested):

Source/WPEFramework/PluginServer.h
pwielders: Suggest any other code. I did not read the spec, and happy with any value, just think that 200 is definitely not correct as something is not good (it never reached the JSONRPC logic)

Source/websocket/WebSerializer.h
pwielders: Correct will adapt it..

Source/WPEFramework/PluginServer.h
pwielders: Correct, will adapt to int32_t

pwielders:
On the longer story, as it contains multiple remarks lets walk them down one-by-one:

Q:  "Do we want to/can we strictly follow the spec here? I like trying to return the ID if we can to make it easier for clients to correlate which message was malformed, but currently we return an empty body if we can't deduce the ID."
A: Definitely want to, if we can, return the ID included as it helps the other side finding the issue.
Also agree to return an id: null, and the error message if we can not, we just make it the clients problem to find the root cause :-)

Q:Missing closing brace - still returns 200 OK
A: Bug needs to be fixed, can you update one of the already existing Metrol-802/803 to track this ?

Q) I guess the issue here is we're treating params as an opaque JSON blob, but I think we should still return an error as the whole document is invalid JSON
A) Indeed since it is opaque the choice was made to not analyse it and leave it up to the "receiver" to properly parse. If we start validating opaque structures we have to do it twice. We are here discussing exceptional situations and spending twice the effort for opaque object to me, for an embedded system sounds like overkill. At the end, the callee will still get an error, just maybe not the expected result... So my suggestion here would be to leave this as is..
And as I reread I reconsidered :-) I think we should still return a parse error but now the parse error should be coming from the JSONRPC base class (in stead of the Serialization) so agree, will fix this (but not in the PluginServer :-) )

Q) Requests without an ID still hang. This example was taken from JSON-RPC spec. According to spec we should return 32600 with a null ID value in this scenario.
A) That is interesting as I read the spec differently and hence why you do not get an answer. The spec says for id (source: https://www.jsonrpc.org/specification#request_object) 
```
id
An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [[1]](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain fractional parts [[2]](https://www.jsonrpc.org/specification#id2)
```

Specifically this part: _"If it is not included it is assumed to be a notification"_ to me suggests that we should *not* send an answer (hence why the client is now incorrectly waiting)  since it is a notification?? But we can debate the expectation here.
Hmmm, as usual after answering all the questions I tend to reread the questions and answers. I can imagine that a response on a notification can be an empty message (I guess??) with a null for the id. However technically I see some severe issues with this:
1) Additional bytes for notifications (increased overhead)
2) Serialization, as you do not know which "notification" is acknowledged, you must wait before sending a next notification for the answer. Which makes the code probably quite complex. You need to stall that websocket... For what ? 

This means a lot of the mechanisms currently deployed in Thunder need refactoring. Currently we do not wait for notifications, they are considered fire-and-forget. This feature would just slow the whole communication down and increase communication overhead (a response must be send) for what? If the notification is not acknowledged for N seconds should we retry? The communication (at least for now) is over connection-oriented channels and thus should be guaranteed to deliver (unless the other side dies but that is detected :-) ). So I really would like to debate those examples, has a "paper-architect" written those specs ?

Q) Junk data for the method param will return 202 and an "Unknown Method" error. Whilst technically correct, compliance to the spec should return a parse error. Not sure which I prefer!
A) Agree, I have no preference here, as the method is parsed before the Invoke, we could make it a parse error without a penalty, in comparison to the opaque one (parameters) so up for a vote :-)

Q) Side note, the HTTP error code on this response is 202 Failure on JSONRPC: 4294934695, the 4294934695 feels like an overflow?
A) I guess a few lines later, you spotted the issue :-) Will fix.

Source/WPEFramework/PluginServer.h
pwielders: It automatically became Web::STATUS_BAD_REQUEST, since both with or without ID (passed back "id": null) it also return for the RESTFul call: Web::STATUS_BAD_REQUEST. Again did not read spec, so we can make it anything you like :-)

pwielders:
Did not test yet, left it for tomorrow :-)

TeknoVenus:
> A: Definitely want to, if we can, return the ID included as it helps the other side finding the issue.
> Also agree to return an id: null, and the error message if we can not, we just make it the clients problem to find the root cause :-)

Makes sense to me, lets return the ID where we can, but if not return null. 

>A: Bug needs to be fixed, can you update one of the already existing Metrol-802/803 to track this ?

Will do

>And as I reread I reconsidered :-) I think we should still return a parse error but now the parse error should be coming from the JSONRPC base class (in stead of the Serialization) so agree, will fix this (but not in the PluginServer :-) )

Agree, it should not be the plugins responsibility to deal with malformed JSON otherwise we're putting extra effort on plugin authors. The framework should take care of this for us.

>A) That is interesting as I read the spec differently and hence why you do not get an answer.

Looking a bit closer I think I see the issue

The example in section 7 shows the following:

```
--> {"jsonrpc": "2.0", "method": 1, "params": "bar"}
<-- {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null}
```

There are two issues in this example:

* Method is an int, not a string
* Missing ID

I believe the method being an int is what triggers the invalid request reply here, not the missing ID

As per section 4.1 in the spec, it states the following rules:

* A Notification is a Request object without an "id" member. 
* A Request object that is a Notification signifies the Client's lack of interest in the corresponding Response object, and as such no Response object needs to be returned to the client. 
* The Server MUST NOT reply to a  Notification, including those that are within a batch request.
* As such, the Client would not be aware of any errors (like e.g. "Invalid params","Internal error").

So I propose:

* If the JSON is un-parsable in any way, then return a parse error
  * For HTTP, return 4xx
* If the JSON is valid JSON, but not a valid JSON-RPC request/notification (e.g. missing the `jsonrpc` property, or method is an int not a string), then return invalid request
  * For HTTP, return 4xx error
* If the JSON-RPC request is valid but is a notification (e.g. missing the `id`)
  * For websockets, return nothing
  * For HTTP, return 204 and no body. We should never be in a scenario where the HTTP connection is never closed

There's some discussion in the JSON-RPC group, but it doesn't make things clearer to me!

* https://groups.google.com/g/json-rpc/c/a5tApN2Vn-c/m/oEUPkhl3AgAJ
* https://groups.google.com/g/json-rpc/c/2tQ9lRLRA1I/m/oiaocLY21F4J

Happy to discuss further on Skype! :)

>  A) Agree, I have no preference here, as the method is parsed before the  Invoke, we could make it a parse error without a penalty, in comparison  to the opaque one (parameters) so up for a vote :-)

My gut says we should go for parse error here - essentially any scenario where the JSON cannot be cleanly parsed we should return a parse error.

pwielders:
@TeknoVenus I think we covered all remarks and I did test them now all on Windows. All Errors now seem to be recognized, whether the JSONRPC code for the error is the expected one... not sure. Let me know if you still found an issue.
Oops one pending issue:
In case we send a notification over HTTP (Which is unexpected anyway :-) ) currently not sending back HTTP, return 204 and no body. We should never be in a scenario where the HTTP connection is never closed. 
Will pick that one up tomorrow!

pwielders:
@TeknoVenus all ready. Please review and try :-)

TeknoVenus (approved):
Looks good to me!

MFransen69 (approved):
LGTM :)

PR #1357, opened 2023-07-14
[Messaging] SYSLOG_GLOBAL is now deprecated

VeithMetro:


VeithMetro:
@pwielders first I tried with the preprocessor statements instead of the `DEPRECATED` macro, but it was possible to use the macro, I just had to do that correctly - another school day right before the weekend 😄

pwielders (approved):

PR #1360, opened 2023-07-18
WarningReporting: config option support with new config too

HaseenaSainul:


VeithMetro:
looks just like the generic cmake config, so all seems good

pwielders (approved):

PR #1363, opened 2023-07-18
[Docs] Error handling

TeknoVenus:
As discussed with @pwielders the `ICOMLink` example will currently deadlock if used with an out-of-process plugin due to the Dangling() notification being called on the communicator Close thread. This will be resolved separately.

### DIFF
--- a/docs/plugin/errors.md
+++ b/docs/plugin/errors.md
@@ -250,6 +250,8 @@ void TestPlugin::Deinitialize(PluginHost::IShell* service)
         _service->Unregister(&_notification);
 
         if (_testPlugin != nullptr) {
+            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+
             // This should release the last reference and destruct the object. If not,
             // there's something else holding on to it and we have a leak
             VARIABLE_IS_NOT_USED uint32_t result = _testPlugin->Release();
@@ -257,7 +259,6 @@ void TestPlugin::Deinitialize(PluginHost::IShell* service)
             _testPlugin = nullptr;
 
             // Shut down the out-of-process connection if still running
-            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
             if (connection != nullptr) {
                 connection->Terminate();
                 connection->Release();
@@ -420,6 +421,8 @@ void TestPlugin::Deinitialize(PluginHost::IShell* service)
         _service->Unregister(&_notification);
 
         if (_testPlugin != nullptr) {
+            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+
             // This should release the last reference and destruct the object. If not,
             // there's something else holding on to it and we have a leak
             VARIABLE_IS_NOT_USED uint32_t result = _testPlugin->Release();
@@ -427,7 +430,6 @@ void TestPlugin::Deinitialize(PluginHost::IShell* service)
             _testPlugin = nullptr;
 
             // Shut down the out-of-process connection if still running
-            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
             if (connection != nullptr) {
                 connection->Terminate();
                 connection->Release();
--- a/docs/plugin/errors.md
+++ b/docs/plugin/errors.md
@@ -215,7 +215,7 @@ private:
 ```
 
 
-```c++ linenums="1" title="TestPlugin.cpp" hl_lines="12 32 61-72"
+```c++ linenums="1" title="TestPlugin.cpp" hl_lines="12 32 62-73"
 const string TestPlugin::Initialize(PluginHost::IShell* service)
 {
     ASSERT(_service == nullptr);
@@ -386,7 +386,7 @@ private:
 };
 ```
 
-```cpp title="TestPlugin.cpp" linenums="1" hl_lines="12 32 61-65"
+```cpp title="TestPlugin.cpp" linenums="1" hl_lines="12 32 62-66"
 const string TestPlugin::Initialize(PluginHost::IShell* service)
 {
     ASSERT(_service == nullptr);
--- a/docs/plugin/errors.md
+++ b/docs/plugin/errors.md
@@ -1,0 +1,506 @@
+As with all code, it is important that Thunder plugins handle errors gracefully and consistently. 
+
+## Exceptions
+
+By default, Thunder is compiled with `-fno-exceptions` to disable exception support in the framework. This can be changed by enabling the `EXCEPTIONS_ENABLE` CMake option. As a result, plugins should **never** be designed to throw exceptions.
+
+If an exception does occur, the WPEFramework process will immediately shut down with an error to prevent any further issues and log the following message:
+
+```
+WPEFramework shutting down due to an uncaught exception.
+```
+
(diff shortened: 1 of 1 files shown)


pwielders:
@TeknoVenus I think the issues with the, not deadlock, Dangling reporting on the Communication Thread -> WorkerPool thread resolves the issue that you observed. This issue was also only visible if the TestPLugin was running out-of-process and indeed during the reporting of the dangling, a call out to the out-of-process part of the plugin was done and this the Communication thread was blocked by the COMRPC Invoke, waiting for messages (COMRPC) being send back and forth.
I think this pull: https://github.com/rdkcentral/Thunder/pull/1366 resolved it. Correct ?

pwielders (dismissed):

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+	The `Deactivated()` function might be called on a socket thread which we do not want to block. As a result, this function should return as quickly as possible and any work that needs doing (e.g. deactivating the plugin) must be done on a separate thread in the main worker pool.
+
+Example:
+
+```c++ linenums="1" title="TestPlugin.h" hl_lines="3-30 60"
+class TestPlugin : public PluginHost::IPlugin, public PluginHost::JSONRPC {
+private:
+    class Notification : public RPC::IRemoteConnection::INotification {
+    public:
+        Notification() = delete;
```
MFransen69: I guess and old example was taken :) Normally we do no longer delete the default c'tor if another one is present as it is then not added anyway

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+        explicit Notification(TestPlugin* parent)
+            : _parent(*parent)
+        {
+            ASSERT(parent != nullptr);
+        }
+
+        ~Notification() = default;
```
MFransen69: According to current coding guidelines (although I'm not sure we've already documented this :) ) we would put ~Notification() override = default; to better indicate it is a virtual d'tor

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+        Notification(const Notification&) = delete;
```
MFransen69, line 158: We certainly would also delete the assignment operator and nowadays we also delete the move operator and assignment operator (technically not needed but to make it more clear that if you would implement one or the other then it would be needed (but that last part we certainly have not documented and you would only see in very new code)

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+        INTERFACE_ENTRY(RPC::IRemoteConnection::INotification)
+        END_INTERFACE_MAP
+
+    private:
+        TestPlugin& _parent;
+    };
+
+public:
+    TestPlugin();
+    ~TestPlugin() override;
```
MFransen69: seen that the implementation is empty we most likely would make it = default

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+
+    if (!_testPlugin) {
+        // Error occurred, return non-empty string
+        result = "Failed to create ITestPlugin";
+    }
+
+    return result;
+}
+
+void TestPlugin::Deinitialize(PluginHost::IShell* service)
```
MFransen69, line 244: I guess you discussed it with Pierre and we only want to discuss the connection notification but for a complete Deinitialize example some code would be missing. Let me take that up with Pierre.

docs/plugin/errors.md
```diff
@@ -0,0 +1,506 @@
...
+| code             | message          | meaning                                                      |
+| ---------------- | ---------------- | ------------------------------------------------------------ |
+| -32700           | Parse error      | Invalid JSON was received by the server. An error occurred on the server while parsing the JSON text. |
+| -32600           | Invalid Request  | The JSON sent is not a valid Request object.                 |
+| -32601           | Method not found | The method does not exist / is not available.                |
+| -32602           | Invalid params   | Invalid method parameter(s).                                 |
+| -32603           | Internal error   | Internal JSON-RPC error.                                     |
+| -32000 to -32099 | Server error     | Reserved for implementation-defined server-errors.           |
+
+In the below example, an attempt is made in activate a non-existent plugin. The Controller plugin returns `ERROR_UNKNOWN_KEY`
```
MFransen69: Your the native Englishman here but shouldn't it be "an attempt is made to activate" or "in activating" (feel free to laugh if I'm incorrect :) )

MFransen69 (dismissed):
As all documentation is better than none at all feel free to ignore all my small/unimportant remarks (I guess the goal is also to not completely follow our latest idea of coding guidelines :) ), therefore I just approved it with some small comments. So just do what you think is best :)
(the only thing I want to check with Pierre is the Deinitialize example but if needed we will get back on that)

I discussed with Pierre in the mean time and he couldn't remember it was a deliberate decision to leave the connection termination for OOP plugins out of the Deinitialize example. So we propose to add that part as well so if people look the Deinitialize in the code example it is complete. See here for an example of the connection termination that should be added: 
https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/blob/b1c0c7a47b8e758058b820c528dadedcf19965d6/Packager/Packager.cpp#L85-L102

docs/plugin/errors.md
TeknoVenus: The main focus of the example is the connection notification but if the deinitialize() method isn't complete lets make is as accurate as possible. What's missing?

docs/plugin/errors.md
TeknoVenus: Good spot, that's a typo :) Will correct!

docs/plugin/errors.md
MFransen69: I added it above in the mean time but I did that by editing my comment which was not that smart I guess :)

Indeed we realize the example is on the connection notification but I talked to Pierre and we also think to better have the  example complete in case people start to copy from it :)

See here for an example of the connection termination for OoP plugins that should be added: 
https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/blob/b1c0c7a47b8e758058b820c528dadedcf19965d6/Packager/Packager.cpp#L85-L102

TeknoVenus: Great, thanks! Will update the example :)

TeknoVenus:
@MFransen69 Updated the example code to hopefully conform with coding guidelines better :)

docs/plugin/errors.md
```diff
@@ -0,0 +1,514 @@
...
+
+        if (_testPlugin != nullptr) {
+            // This should release the last reference and destruct the object. If not,
+            // there's something else holding on to it and we have a leak
+            VARIABLE_IS_NOT_USED uint32_t result = _testPlugin->Release();
+            ASSERT(result == Core::ERROR_DESTRUCTION_SUCCEEDED);
+            _testPlugin = nullptr;
+
+            // Shut down the out-of-process connection if still running
+            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
```
MFransen69: This needs a small correction, this line needs to move to above line 255. The release on the _testPlugin could also release the connection and then you would not be able to get it here anymore. If you do it before the release on _testPlugin it will always be available

MFransen69 (changes requested):

docs/plugin/errors.md
TeknoVenus: Corrected

MFransen69 (approved):

PR #1364, opened 2023-07-19
[Messaging] Fixing the way of deprecating the SYSLOG_GLOBAL macro

VeithMetro:


VeithMetro:
@pwielders while investigating a different issue, I did a full rebuild with rdkservices as well, and I got a compilation error regarding the `SYSLOG_GLOBAL` macro. Turns out we can't just use `DEPRECATED` - I did some more reading on this and looks like there is no way of using `__attribute__((deprecated))` on the macro, but it's suggested to use the `_Pragma` instead, since this was introduced to be used with macros

pwielders (approved):

PR #1367, opened 2023-07-19
[Docs] COM-RPC tags

TeknoVenus:
Move the tags documentation from ThunderTools repo into main documentation portal

Update to include latest tags


docs/plugin/interfaces/tags.md
```diff
@@ -0,0 +1,483 @@
...
+If a parameter is outside the valid range, then there are two possibilities:
+
+* If running a debug build, an ASSERT will be triggered if the value is outside the allowed range
+* If the stub generator is invoked with the `--secure` flag, then the range will be checked on all builds and an error (`ERROR_INVALID_RANGE`) will be returned if the value is outside the range
+
+##### Example
+
+* `@restrict:1..32` - Value must be between 1 and 32
+* `@restrict:256..1K` - Value must be between 256B and 1K in side
+* `@restrict:1M` - Value must be less than 1M
```
sebaszm: The range is inclusive. For example to have restriction for less than 1 MiB it should be `@restrict:(1M-1)`

docs/plugin/interfaces/tags.md
```diff
@@ -0,0 +1,483 @@
...
+In [IPerformance.h](https://github.com/rdkcentral/ThunderInterfaces/blob/5fa166bd17c6b910696c6113c5520141bcdea07b/interfaces/IPerformance.h#L32) it specifies, the maximum length for the buffer.
+
+
+### JSON-RPC Related Tags
+
+| Tag|Short Description|Deprecated|StubGen|JsonGen|Scope|
+|--|--|:--:|:--:|:--:|:--:|
+|[@json](#json)|Marks a class as JsonGenerator input | | No| Yes| Class|
+|[@json:omit](#json_omit)|Marks a method/property/notification to omit | | No| Yes|Method|
+|[@uncompliant:extended](#extended)|Generates the extended format instead of the short one| Yes | No| Yes|Class|
```
sebaszm: There is also `@uncompliant:collapsed` and `@compliant`. While compliant format is the default one it's possible to change the default with a switch to the generator, then the a `@compliant` tag may become useful.

docs/plugin/interfaces/tags.md
```diff
@@ -0,0 +1,483 @@
...
+All tags follow the same format of `@` followed by the tag name. They can be inserted into the interface using either inline `//` comments or `/* */` block comments. Tags can influence the generation of the COM-RPC ProxyStubs and the generated JSON-RPC interfaces. They can also be used to configure the documentation generation
+
+## Summary
+
+### General purpose tags
+
+| Tag|Short Description|Deprecated|StubGen|JsonGen|Scope|
+|--|--|:--:|:--:|:--:|:--:|
+|[@stubgen:skip](#stubgenskip)|Stop processing current file, now deprecated| Yes | Yes|Yes|File|
+|[@stop](#stop)|Same as @stubgen:skip, not deprecated, but also never advertised ||Yes| Yes|File|
```
sebaszm: That is an odd thing to say in a documentation (not advertised). We wish to get rid of `@stubgen:skip` in favour of `@omit`, hence the deprecation, but `@stop` is still there as a last resort when the generator is technically unable to parse the file, though in such case it is recommended that the complicated part is moved to a separate header file instead.

docs/plugin/interfaces/tags.md
```diff
@@ -0,0 +1,483 @@
...
+|[@stubgen:include](#stubgen_include)|Insert another C++ file ||Yes| Yes|File|
+|[@stubgen:stub](#stubgen_stub)|Emit empty function stub instead of full proxy implementation | | Yes| No| Method|
+|[@stub](#stub)|Same as `@stubgen:stub` | | Yes| No|Method|
+|[@insert](#insert)|Same as `@stubgen:include` | | Yes|Yes|File|
+|[@define](#define)|Defines a literal as a known identifier | | Yes |Yes|File|
+
+<hr/>
+
+#### @stubgen:skip
+!!! warning
```
sebaszm, line 24: Perhaps instead of this !!! exclamation style use the markup's blockquoute `>` ?

sebaszm (changes requested):

docs/plugin/interfaces/tags.md
TeknoVenus: Will correct

docs/plugin/interfaces/tags.md
TeknoVenus: What's the difference between extended and collapsed output? Could you share an example of how they differ?

docs/plugin/interfaces/tags.md
TeknoVenus: Was copy/pasted from the old copy of the documentation. Will clarify

docs/plugin/interfaces/tags.md
TeknoVenus: The `!!!` syntax generates admonitions (https://squidfunk.github.io/mkdocs-material/reference/admonitions/) when the documentation site is built, so is preferable to just standard markdown blockquotes for doing callouts

TeknoVenus:
@sebaszm Made some minor tweaks to address comments.

If you could clarify the difference between the collapsed/extended compliant tags that would be appreciated.

sebaszm:
> @sebaszm Made some minor tweaks to address comments.
> 
> If you could clarify the difference between the collapsed/extended compliant tags that would be appreciated.

Thanks!

The snag is in the way how calls with a single parameter are defined. The JSON-RPC compliant way is to enclose them in an object, like `params: { "name":  "abcd" }`. Historically there was also possible to collapse it, like `params: "abcd"`, since there is only one parameter it doesn't need a name to be accessed. So the _collapsed_ format means that methods, notifications and properties omit the enclosing object when there is only one parameter to pass. The _exteneded_ format is a mixture where the collapse is only used for properties, since they enforce one parameter anyway, but not for methods and notifications. And _compliant_ format is that all calls including properties use the enclosing object. _Collapsed_ and _extended_ are not recommended for new interfaces.

sebaszm (approved):

PR #1374, opened 2023-07-26
[Docs] Interface creation and design

TeknoVenus:
Overview of designing an interface, code-generation and worked example.

### DIFF
--- a/docs/plugin/interfaces/interfaces.md
+++ b/docs/plugin/interfaces/interfaces.md
@@ -30,9 +30,9 @@ During the build, code-generation tools automatically convert these interfaces i
 
 All COM-RPC interfaces inherit virtually from `IUnknown`. As with Microsoft COM, this contains 3 vital methods:
 
-* **QueryInterface** - allows clients to dynamically discover (at run time) whether or not an interface is supported. Given an interface ID, if the plugin supports that interface than a pointer to that interface will be returned, otherwise will return nullptr.
-* **AddRef** - increase the reference count on the object
-* **Release** - decrement the reference count on the object. When 0, it is safe to destroy the object
+* **QueryInterface()** - provides interface navigation. Allows clients to dynamically discover (at run time) whether or not an interface is supported. Given an interface ID, if the plugin supports that interface than a pointer to that interface will be returned, otherwise will return nullptr.
+* **AddRef()** - lifetime management. Increase the reference count on the object
+* **Release()** -  lifetime management. Decrement the reference count on the object. When 0, it is safe to destroy the object
 
 #### Interface Characteristics
 
@@ -44,13 +44,17 @@ All COM-RPC interfaces inherit virtually from `IUnknown`. As with Microsoft COM,
 #### Guidelines
 
 * Each COM-RPC interface must inherit **virtually** from `Core::IUnknown` and have a unique ID
+    * Virtual inheritance is important to prevent the [diamond problem](https://isocpp.org/wiki/faq/multiple-inheritance#mi-diamond) and ensure multiple interfaces are implemented on a single, reference counted object. Without this, each interface might have its own reference count and not be destroyed correctly.
+
 * Ensure API compatibility is maintained when updating interfaces to avoid breaking consumers
 * Methods should be pure virtual methods that can be overridden by the plugin that implements the interface
 * Methods should return **`Core::hresult`** which will store the error code from the method
     * If a method succeeds, it should return `Core::ERROR_NONE`
     * If a method fails, it should return a suitable error code that reflects the failure (e.g `Core::ERROR_READ_ERROR`)
     * If an error occurs over the COM-RPC transport the `COM_ERROR` bit will be set. This allows consumers to determine where the failure occurred
-* Ensure all enums have explicit data types set (e.g. `uint8_t`)
+* Ensure all enums have explicit data types set (if not set, the code generators will fall back to `uint32_t` as a safe default)
+    * If you know you will be communicating over different architectures, then you can define `INSTANCE_ID_BITS`  to specify a specific default integer width
+
 * C++ types such as `std::vector` and `std::map` are **not** compatible with COM-RPC
     * Only "plain-old data" (POD) types can be used (e.g. scalar values, interface pointers)
     * COM-RPC can auto-generate iterators for returning multiple results
@@ -59,7 +63,7 @@ All COM-RPC interfaces inherit virtually from `IUnknown`. As with Microsoft COM,
 
 #### Notifications & Sinks
 
-A COM-RPC interface not only allows for defining the methods exposed by a plugin, but can also be used to define notifications that plugins can rause and clients can subscribe to.
+A COM-RPC interface not only allows for defining the methods exposed by a plugin, but can also be used to define notifications that plugins can raise and clients can subscribe to.
 
 As with Microsoft COM, this is done by allowing clients to create implementations of notification interfaces as sinks, and register that sink with the plugin. When a notification occurs, the plugin will call the methods on the provided sink.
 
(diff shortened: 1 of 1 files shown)


docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+
+## Designing a good Interface
+A good interface defines a set of methods that a plugin can support, without dictating anything about the actual code that will implement the interface. It is possible many plugins could implement that same interface if they provide overlapping functionality (for example an `IBrowser` interface could be implemented by different web browser engines).
+
+The interface provides a clear boundary between the code that invokes methods on the plugin and the plugin code that implements the functionality. Typically, interface definitions are stored in a separate repository to the plugin implementations to reflect this boundary.
+
+Well designed interfaces should also be easy to read and self explanatory. Methods names should be descriptive and have doxygen-style comments that describe their inputs and outputs. All methods should return an error code to indicate whether they completed successfully, and any data that should be returned by the method should be stored in an output parameter.
```
pwielders: "All methods should return an error code" don't we want to mention her that is is prefered to be Core::hresult ? Or something like: "All methods should return an error code (core::hresult)", maybe with reasoning so that the COMRPC errors can be reported in the upper part or something like that ?

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+
+It is possible to build an interface from smaller sub-interfaces using composition. This is preferred to having a large monolithic interface, as is encourages reuse and modularity.
```
pwielders: I think typo :-) "as is encourages" =>  "as it encourages"

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+
+## Interface Definitions
+
+### COM-RPC Interfaces
+
+When designing a plugin interface, it is always best to start from the COM-RPC interface definition. A COM-RPC interface is defined in a C++ header file, and is a struct that inherits virtually from `Core::IUnknown`. This is very similar to interfaces in Microsoft COM. The plugin code will then provide an implementation of one or more COM-RPC interfaces.
+
+During the build, code-generation tools automatically convert these interfaces into ProxyStub classes that are used to resolve the interface at runtime and handle the serialisation/deserialisation of data over the wire.
```
pwielders: "During the build, code-generation tools automatically convert these interfaces into ProxyStub classes" convert feels a bit like it is a necessity??? It is more like (I think but leave it up to you) "During the build, code-generation tools generate, from these header files, code for marshaling (Proxies and Stubs) these calls over process boundaries" ? Cuase proxies and stubs are not a necessity, only if you want to run them in different processes..

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+When designing a plugin interface, it is always best to start from the COM-RPC interface definition. A COM-RPC interface is defined in a C++ header file, and is a struct that inherits virtually from `Core::IUnknown`. This is very similar to interfaces in Microsoft COM. The plugin code will then provide an implementation of one or more COM-RPC interfaces.
+
+During the build, code-generation tools automatically convert these interfaces into ProxyStub classes that are used to resolve the interface at runtime and handle the serialisation/deserialisation of data over the wire. 
+
+#### IUnknown
+
+All COM-RPC interfaces inherit virtually from `IUnknown`. As with Microsoft COM, this contains 3 vital methods:
+
+* **QueryInterface** - allows clients to dynamically discover (at run time) whether or not an interface is supported. Given an interface ID, if the plugin supports that interface than a pointer to that interface will be returned, otherwise will return nullptr.
+* **AddRef** - increase the reference count on the object
```
pwielders: Maybe mention that AddRef and Release are for Lifetime management of the object, and QueryInterface for Interface Navigation ? as generic more common terminology ?

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+#### Interface Characteristics
+
+* **Interfaces are not plugins**: A Thunder plugin can implement 0 or more COM-RPC interfaces, but an interface cannot be instantiated by itself because it does not define an implementation. It is possible for many Thunder plugins to implement the same interface
+* **COM-RPC clients interact with pointers to interfaces**: A client application communicating with a Thunder plugin over COM-RPC will receive nothing more than an opaque pointer through which it can access the interface methods. It cannot access any data from the plugin that implements the interface. This encapsulation ensures the client can only communicate with the plugin over an agreed interface
+* **Interfaces are immutable**: COM-RPC interfaces are not versioned, so if an interface needs changing it must only add new methods. It must never break existing methods
+* **Interfaces are strongly typed**: Each COM-RPC interface has a unique ID, used to identify it in the system. When a client requests a pointer to an interface, it does so using the ID of that interface. 
+
+#### Guidelines
+
+* Each COM-RPC interface must inherit **virtually** from `Core::IUnknown` and have a unique ID
```
pwielders, line 46: Highlight the reason behind the virtual? (making sure that if multiple interfaces are implemented on 1 object, there is only 1 object (instance) implementing the IUnknown and thus 1 instance that does the reference counting ? If you would not have this virtual, each interface might have its own reference count and thus if the first interface might be out of scope (reaching ref count 0), the whole object (including the other interface that might well still be in use) is deleted..

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+* Ensure API compatibility is maintained when updating interfaces to avoid breaking consumers
+* Methods should be pure virtual methods that can be overridden by the plugin that implements the interface
+* Methods should return `Core::hresult` which will store the error code from the method
+    * If the method succeeds, it should return `Core::ERROR_NONE`
+    * If an error occurs over the COM-RPC transport, the `COM_ERROR` bit will be set
+* Ensure all enums have explicit data types set (e.g. `uint8_t`)
+* C++ types such as `std::vector` and `std::map` are not compatible with COM-RPC
+    * Only "plain-old data" (POD) types can be used (e.g. scalar values, interface pointers)
+    * COM-RPC can auto-generate iterators for returning multiple results
+* Ensure integer widths are explicitly set (e.g. use `uint16_t` or `uint8_t` instead of just `int`) to prevent issues if crossing architecture boundaries
```
pwielders, line 61: Agree to the explicit definition, but I think if you do not mention this explixitly (for example on enums) it always falls back to a uint32_t regardless of the architecture as the generator tools do not know the architecture and thus made assumptions. Hence why the instanceid (type of Thunder) is used in the ProxyStub generation. It is assumed that if you build thunder and you need to use it over different architectures (32bit/64bits), the none conforming architecture (lowest denominator her, 32bit, but, during its build define: https://github.com/rdkcentral/Thunder/blob/98bda4436c51eacb076e4dc0ad87a68f0c04509c/Source/core/CMakeLists.txt#L197 -DINSTANCE_ID_BITS=64 so that although it is a 32 bits platform all crucial information in COMRPC, like the implementation instance id is set at 64 bits.

docs/plugin/interfaces/interfaces.md
```diff
@@ -0,0 +1,603 @@
...
+* Prefer asynchronous APIs for long running tasks (and use notifications to signal completion)
+
+#### Notifications & Sinks
+
+A COM-RPC interface not only allows for defining the methods exposed by a plugin, but can also be used to define notifications that plugins can rause and clients can subscribe to.
```
pwielders: rause ?

docs/plugin/interfaces/interfaces.md
TeknoVenus: I cover this later on in the "Guidelines" section, don't think it needs adding here as well

TeknoVenus:
@pwielders Any other changes needed for this page? Could possibly go into more detail in some areas, but I think it's enough for now.

pwielders (dismissed):

pwielders:
@TeknoVenus approved but I think it still contained some typos as I commented..

TeknoVenus:
Thanks, those comments didn't appear for me before! Will make corrections :)

pwielders (approved):

pwielders:
@TeknoVenus Thansk again for a good piece of documentation !

PR #1375, opened 2023-07-27
[Actions] Adding Warning Reporting to the Linux build

VeithMetro:
For now it's only a part of Linux build, as I don't think we have Warning Reporting project file in the main solution file for Windows. I could try to add it to the solution file, if it's something that's worth having, @pwielders?


pwielders:
@VeithMetro lets drop it on windows for now. Could you raise a JIRA ticket for it that if we have time to add it for Windows as well at a later time?

pwielders (approved):

PR #1376, opened 2023-07-27
processcontainers: changes to compile it again

HaseenaSainul:


Source/extensions/processcontainers/Messaging.h
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "Module.h"
+
+#include "../tracing/tracing.h"
```
pwielders: Do we not want to use messaging/messaging.h here ?

pwielders (commented):
Minor one, have a look at it if we need to change it ?

Source/extensions/processcontainers/Messaging.h
HaseenaSainul: done

pwielders (approved):

PR #1377, opened 2023-07-27
[Messaging] Necessary adjustments to allow using enum from MessageStore in the interfaces

VeithMetro:
@pwielders I also created a PR to ThunderInterfaces (https://github.com/rdkcentral/ThunderInterfaces/pull/266), but I'll have to rerun Actions there after this one is merged

### DIFF
--- a/Source/core/MessageStore.h
+++ b/Source/core/MessageStore.h
@@ -55,8 +55,6 @@ namespace Core {
         public:
             Metadata(const Metadata&) = default;
             Metadata& operator=(const Metadata&) = default;
-            Metadata(Metadata&&) = default;
-            Metadata& operator=(Metadata&&) = default;
 
             Metadata()
                 : _type(INVALID)
--- a/Source/core/MessageStore.h
+++ b/Source/core/MessageStore.h
@@ -52,9 +52,13 @@ namespace Core {
                 STANDARD_ERROR = 5
             };
 
+// @stop
+
         public:
             Metadata(const Metadata&) = default;
             Metadata& operator=(const Metadata&) = default;
+            Metadata(Metadata&&) = default;
+            Metadata& operator=(Metadata&&) = default;
 
             Metadata()
                 : _type(INVALID)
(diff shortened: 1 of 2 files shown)


Source/core/MessageStore.h
pwielders, line 1: Why did you remove the move constructors/operators?

pwielders (dismissed):

Source/core/MessageStore.h
VeithMetro: `ProxyStubGenerator: ERROR: IMessageControl.h: /home/veith/buildroot/output/host/arm-buildroot-linux-gnueabihf/sysroot/usr/include/WPEFramework/core/MessageStore.h(59): parse error: rvalue references are not supported: 'Metadata &&'
CMake Error at /home/veith/buildroot/output/host/share/cmake/Modules/FindProxyStubGenerator.cmake:95 (message):
  ProxyStubGenerator generator failed.`
Sebastian said to remove it for now, and he is going to add it to the generator in the future.

VeithMetro:
@pwielders looks like all of our Windows builds are crashing with such error:
`'SLEEPSLOT_POLLING_TIME': undeclared identifier`
And it is happening since this got merged: https://github.com/rdkcentral/Thunder/commit/a9e321ef1f5aeff320373ac0070fe36fc0449748#

Source/core/MessageStore.h
sebaszm: As long as it does not serve any purpose? Does it not?

VeithMetro: It does not, and it works as expected with move constructor/operator removed.
/edit: I've spoken with Sebastian and he suggested a proper way to test that, so I'll do that now to be extra sure.

VeithMetro:
@pwielders I tested if the move constructor/operator is used anywhere, and it is not, but after talking to @sebaszm he proposed that instead of removing the move semantics, we can stop the parser before them - seems to be working as indented and this way we can keep the move constructor/operator

sebaszm (approved):

PR #1378, opened 2023-07-27
[ONEM-31782] wait for Websocket in Open()

dawid-trendota-red:


Source/websocket/WebSocketLink.h
```diff
@@ -1022,9 +1022,41 @@ POP_WARNING()
...
+            }
+
+            uint32_t result = (((time == 0) || (IsWebSocket() == true)) ? Core::ERROR_NONE : Core::ERROR_TIMEDOUT);
+            Unlock();
+            return (result);
+        }
         uint32_t Open(const uint32_t waitTime)
         {
-            return (_channel.Open(waitTime));
+            auto start = std::chrono::high_resolution_clock::now();
```
pwielders: Agree that this is, given the waitTime, the best accurate determination of what waitTime is left, but it is a bit expensive (get times substracting 64 bits etc), thread switching with the SleepMs etc... Why not (and that might even improve performace) not do the open with a wait time of 0 (waitTime) and than only wait in the waitForLink(waitTime) ?

Source/websocket/WebSocketLink.h
```diff
@@ -1022,9 +1022,41 @@ POP_WARNING()
         {
             return (_channel.AbortUpgrade(status, reason));
         }
+        uint32_t WaitForLink(const uint32_t time) const
+        {
+            const int SLEEPSLOT_TIME = 100;
```
pwielders: good that you did a constant SLEEPSLOT_TIME = 100. Could we reuse the Socket::SLEEPSLOT_TIME. I internally raised a ticket to try to optimize this to move to a 10ms time slot. If we would, this would automatically happen here as well.

pwielders: Maybe add a const SLEEPSLOT_POLLING_TIME in core/Portability.h and use it in the SocketPort.h and use it here, so we change this SLEEPSLOT_POLLING_TIME in two places if we want to reduce the latency of the open/close of streams ?

pwielders (dismissed):

pwielders (approved):

PR #1379, opened 2023-07-27
Development/json number type

msieben:
Amend 1373


msieben:
Cancelling PR while investigating possible issues.

PR #1388, opened 2023-08-03
[Docs] Process Utilities

TeknoVenus:
Document how to find information about running processes and how to spawn child processes. 

Provide example of using the ResourceMonitor to monitor the stdout/err of a running process


pwielders (approved):
Can we have a discussion on this. It is currently aligned but not yet complete (see https://github.com/rdkcentral/Thunder/blob/master/Source/core/MessageStore.h#L51-L52) to route stdout and stderr (process stdin/stdout example) over the messaging functionality.
This avoids the mixup of outputs but might require (do not oversee it all yet) different handling, or disallow this catching of stdout/stderr by plugins..

PR #1389, opened 2023-08-03
Static lib: changes to make Core, WebSocket and Cryptalgo lib to build in static as well

HaseenaSainul:


Source/core/CMakeLists.txt
```diff
@@ -18,11 +18,7 @@
 set(TARGET ${NAMESPACE}Core)
 include(CompileSettings)
 
-find_package(Threads REQUIRED)
-find_package(ExecInfo)
-find_package(LIBRT)
-
-add_library(${TARGET} SHARED
+set(SOURCES
```
bramoosterhuis: please use target_sources instead of introducing a variable...  
https://cmake.org/cmake/help/v3.15/command/target_sources.html

Source/cryptalgo/CMakeLists.txt
```diff
@@ -17,14 +17,21 @@
 
 set(TARGET  ${NAMESPACE}Cryptalgo)
 
-add_library(${TARGET} SHARED
+set(SOURCES
```
bramoosterhuis: please use target_sources instead of introducing a variable...
https://cmake.org/cmake/help/v3.15/command/target_sources.html

Source/websocket/CMakeLists.txt
```diff
@@ -17,17 +17,14 @@
 
 set(TARGET ${NAMESPACE}WebSocket)
 
-add_library(${TARGET} SHARED
-        Module.cpp
+set(SOURCES
```
bramoosterhuis: please use target_sources instead of introducing a variable...
https://cmake.org/cmake/help/v3.15/command/target_sources.html

bramoosterhuis (changes requested):

Source/core/CMakeLists.txt
HaseenaSainul: updated

Source/cryptalgo/CMakeLists.txt
HaseenaSainul: updated

Source/websocket/CMakeLists.txt
HaseenaSainul: updated

bramoosterhuis (changes requested):
Can you revert back to the original ```add_library``` calls. Then add all optional source files with ```target_sources``` to the desired library. 

Thank you

HaseenaSainul:
> Can you revert back to the original `add_library` calls. Then add all optional source files with `target_sources` to the desired library.
> 
> Thank you

done

bramoosterhuis (approved):

PR #1390, opened 2023-08-04
[Messeging] Fixing issues with Controls on initialization and shutdown

VeithMetro:
1. Ensuring the proper initialization of statics in each transition unit on Thunder launch
2. Calling `Destroy()` on each announced `Control` of every message type to ensure a proper destruction of singletons on shutdown

@pwielders @sebaszm - my unique way to add you as reviewers until I have "great power"


sebaszm (changes requested):
Core::CriticalSection is recursive, so I this shouln't lock up.

sebaszm (dismissed):

Source/WPEFramework/CMakeLists.txt
```diff
@@ -69,7 +69,12 @@ target_link_libraries(${TARGET}
           ${NAMESPACE}COMProcess::${NAMESPACE}COMProcess  # For COM executable define
           Threads::Threads
         )
-        
+
+if(WARNING_REPORTING)
+    target_sources(${TARGET} PRIVATE WarningReportingCategories.cpp)
+    target_compile_definitions(${TARGET} PUBLIC __CORE_WARNING_REPORTING__)
```
pwielders: This looks a bit redundant. This flag should come from the WarningReporting CMake file and not from here as they should be completely independent.  
Also if the flag is set, I was expecting this to be the target_link_libraries(${TARGET}
        PRIVATE {NAMESPACE}WarningReporting::{NAMESPACE}WarningReporting)

@bramoosterhuis how can we make sure that the preprocessor statement (target_compile_definitions) is coming from the WarningReporting Cmake here: https://github.com/rdkcentral/Thunder/blob/master/Source/extensions/warningreporting/CMakeLists.txt#L32 ???

pwielders (changes requested):
I think/hope we still can decouple the preprocessor statement definition to the root, the WarningReproting librabry/cmake.

Source/WPEFramework/CMakeLists.txt
VeithMetro: You're right, this is not even a bit, but completely redundant, since we already have the same target compile definition in the cmake of core. This one can be removed 😄

pwielders (approved):

VeithMetro:
@pwielders I had a chat with @sebaszm, and he agrees on these changes (to unify the registration here with the one in ProxyStubs), so I think we can merge it

PR #1397, opened 2023-08-25
Development/static libraries

bramoosterhuis:


bramoosterhuis:
I tested it with
https://github.com/Metrological/ThunderStaticDemo

pwielders (changes requested):
@bramoosterhuis what triggered this issue,  to resolve it differently, was the move of the core dependencies (https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L79-L81) and in websocket (https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L78) to this higher layer and no longer being part of the "real" depending modules. I do not see these libraries being moved back to the core and/or websocket. Is that still an issue or did you just not do it yet..

.github/workflows/Build Archives Thunder on Linux.yml
```diff
@@ -0,0 +1,118 @@
...
+      with:
+        regex_pattern: '(?<=\[Options:).*(?=\])'
+        regex_flags: 'gim'
+        search_string: ${{github.event.pull_request.body}}
+
+    - name: Build Thunder with archives
+      run: |
+        cmake -G Ninja -S Thunder -B ${{matrix.build_type}}/build/Thunder \
+        -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wpedantic -Werror" \
+        -DBUILD_SHARED_LIBS=OFF \
```
VeithMetro: What is the point of having exactly the same Action, just with the `DBUILD_SHARED_LIBS=OFF`?
If we really want to build with both shared libs `ON` and `OFF`, we could add 3 additional build types, so that we have 6 and it would be in the same action, so like this:
```
strategy:
  matrix:
    build_type: [Debug, Release, MinSizeRel]
    shared_libs: [ON, OFF]
```

.github/workflows/Build Thunder on Linux.yml
```diff
@@ -100,6 +100,7 @@ jobs:
       run: |
         cmake -G Ninja -S Thunder -B ${{matrix.build_type}}/build/Thunder \
         -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wpedantic -Werror" \
+        -DBUILD_SHARED_LIBS=ON \
```
VeithMetro: So just to make sure - do we want actions in all of our other repos to build with Thunder with shared libs enabled?

.github/workflows/Test Thunder.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         cmake -G Ninja -S Thunder -B ${{matrix.build_type}}/build/Thunder \
         -DBINDING="127.0.0.1" \
         -DCMAKE_BUILD_TYPE=${{matrix.build_type}} \
+        -DBUILD_SHARED_LIBS=ON \
```
VeithMetro: Again, just to make sure - do we want to run the test only with shared libs enabled from now on?

VeithMetro (commented):
Looks solid to me, just a few questions in the comments

.github/workflows/Build Archives Thunder on Linux.yml
bramoosterhuis: To answer your first part: Me having no clue how git workflows work :smiling_face_with_tear:

VeithMetro: Everyday is a school day 😄

If we build with shared libs both on and off, then we have to account for it in Actions in other repos. If we want to build e.g. ThunderInterfaces based on both versions of Thunder (the ones with shared libs ON as well as OFF), then we have to make some adjustments to how artifacts are created and then unpacked (artifacts are the way to forward output files from one job to the next).

So, before doing that, we have to decide whether we want to build everything with and without shared libs, or not. Thoughts? @bramoosterhuis @pwielders

bramoosterhuis: IMHO our components don't need to be tested to be statically linked, maybe this demo repo: 
https://github.com/Metrological/ThunderStaticDemo

VeithMetro: I think so too, but on the other hand different jobs within the same action are done in paralel, so it is not like we are causing actions to take twice as long with building both versions. Maybe let's hear @pwielders opinion as well. 

On the side note, I don't have access to this demo repo 😄

bramoosterhuis: OK i fixed the rights of the repo

bramoosterhuis:
> @bramoosterhuis what triggered this issue, to resolve it differently, was the move of the core dependencies (https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L79-L81) and in websocket (https://github.com/rdkcentral/Thunder/blob/master/Source/CMakeLists.txt#L78) to this higher layer and no longer being part of the "real" depending modules. I do not see these libraries being moved back to the core and/or websocket. Is that still an issue or did you just not do it yet..

I totally forgot about this... fixed in https://github.com/rdkcentral/Thunder/pull/1397/commits/cee282e972c212fe018782192bbc567a6133d450

bramoosterhuis:
I switched off the ```tar``` and ```upload``` steps in the workflow for the static build since the rest of our components should use shared objects anyway.

VeithMetro (approved):

VeithMetro:
Since now there are 6 different Linux builds instead of 3, and they have different names (with the addition of shared libraries ON/OFF), it looks like we will need to contact the repo admins to change the required checks to remove:
```
Build Thunder on Linux / Build type - Debug
Build Thunder on Linux / Build type - MinSizeRel
Build Thunder on Linux / Build type - Release
```
And to add:
```
Build Thunder on Linux / Build type - Debug, shared libraries ON
Build Thunder on Linux / Build type - Debug, shared libraries OFF
Build Thunder on Linux / Build type - MinSizeRel, shared libraries ON
Build Thunder on Linux / Build type - MinSizeRel, shared libraries OFF
Build Thunder on Linux / Build type - Release, shared libraries ON
Build Thunder on Linux / Build type - Release, shared libraries OFF
```

VeithMetro:
@bramoosterhuis nice solution with the expression for `shared_lib` flag, now we don't need to make changes to required checks anymore 😄 

/edit: we need to adjust the artifacts in the test action, either make them only if `shared_lib` flag is ON, or we could also save each artifact with a different name (they are not used in different workflows like the ones from the `Build Thunder on...` actions)

bramoosterhuis:
> 

@VeithMetro You were too quick, I was still fiddling with it :sweat_smile:. Fixed in https://github.com/rdkcentral/Thunder/pull/1397/commits/51a85e6c19131f5806c3073e309a599801ce1eec

VeithMetro:
@bramoosterhuis there is still an issue with tests, and I think it's because of `-` in the name of a file in the `tar` command

bramoosterhuis:
> @bramoosterhuis there is still an issue with tests, and I think it's because of `-` in the name of a file in the `tar` command

@VeithMetro No it was because a test was relying on the fact that Core was always a shared object. So with the move to archives it could not load it anymore. I now fixed the test by supplying a independent test lib that always will be a shared object. See https://github.com/rdkcentral/Thunder/pull/1397/commits/1ba8ca97d7ba68b86daeebb51d8e075d8636d6d7

bramoosterhuis:
@HaseenaSainul: FYI; I refactored the static linking of Thunder. Using it is quite easy now. See https://github.com/Metrological/ThunderStaticDemo

VeithMetro (dismissed):

CMakeLists.txt
```diff
@@ -113,18 +112,18 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${NAMESPACE}.cmake"
...
 endif(CMAKE_VERSION VERSION_LESS 3.20.0 AND LEGACY_CONFIG_GENERATOR)
 
 install(DIRECTORY
-            "${CMAKE_SOURCE_DIR}/cmake/common"
-            "${CMAKE_SOURCE_DIR}/cmake/templates"
+            "${CMAKE_CURRENT_SOURCE_DIR}/cmake/common"
+            "${CMAKE_CURRENT_SOURCE_DIR}/cmake/templates"
         DESTINATION lib/cmake/${NAMESPACE})
 
 if (BUILD_LIB_AS_STATIC)
```
HaseenaSainul: This has to be changed based on flag name change to get modules also get installed

cmake/common/CmakeHelperFunctions.cmake
```diff
@@ -140,6 +140,8 @@ function(get_if_link_libraries libs dirs target)
             if ("${_library}" MATCHES "LINK_ONLY")
                 string(REGEX REPLACE "\\$<LINK_ONLY:" "" __fix ${_library})
                 string(REGEX REPLACE ">$" "" _library ${__fix})
+                message(AUTHOR_WARNING "${_library} is marked LINK_ONLY and is now skipped, if you experience link issues out of the blue; start here. :-)")
+                continue()
```
HaseenaSainul, line 144: yes this changes skipping the error I have seen without adding find_package in public scope, that's great :)

pwielders (dismissed):

HaseenaSainul (changes requested):
I am seeing LIBRT dependency error while linkng this with fireboat library. found same issue if I remove add_subdirectory(ThirdParty) in ThunderStaticDemo and use the external one.
In both case we have to get FindLIBRT.cmake. In earlier code, copying this also to usr/lib/cmake/WPEFramework. now this install is not working since the variable name is changed. already given comment against the line

bramoosterhuis:
> I am seeing LIBRT dependency error while linkng this with fireboat library. found same issue if I remove add_subdirectory(ThirdParty) in ThunderStaticDemo and use the external one. In both case we have to get FindLIBRT.cmake. In earlier code, copying this also to usr/lib/cmake/WPEFramework. now this install is not working since the variable name is changed. already given comment against the line

By removing  ```add_subdirectory(ThirdParty)``` in ThunderStaticDemo you deleting the cmake mechanism that injects all build requirements exposed by the library author. 

In other words, your on your own....

HaseenaSainul:
> > I am seeing LIBRT dependency error while linkng this with fireboat library. found same issue if I remove add_subdirectory(ThirdParty) in ThunderStaticDemo and use the external one. In both case we have to get FindLIBRT.cmake. In earlier code, copying this also to usr/lib/cmake/WPEFramework. now this install is not working since the variable name is changed. already given comment against the line
> 
> By removing `add_subdirectory(ThirdParty)` in ThunderStaticDemo you deleting the cmake mechanism that injects all build requirements exposed by the library author.
> 
> In other words, your on your own....

As per the our discussion it is fine @bramoosterhuis. just remove the redundant part then :)

HaseenaSainul (approved):

PR #1403, opened 2023-09-07
[Core] Update base64 and hex string to uint32_t

sebaszm:


pwielders (approved):
Was there a need for more than 64K string back and forth ? or is it used for consistency...

sebaszm:
> Was there a need for more than 64K string back and forth ? or is it used for consistency...

base64 is used for cookie jar and it can end up large, the hex string is upgraded as well for consistency with base64 (they perform a similiar task)

PR #1405, opened 2023-09-13
[Docs] Sockets

TeknoVenus:
Provide some examples of common socket programming tasks, and a background on the Resource Monitor object that should be used for monitoring those sockets.


pwielders:
Good documentation, what might be interesting to add is that domain sockets to be created can also regulate the right they will get by using the pipeline character in de defintion, see Core::NodeId. So addressed could be /tmp/communicator|0777 turns on all flag bits on the socket

pwielders (commented):
See comments :-) Maybe a bit of documentationaround the domain socket for NodeId to pass on the rights..

pwielders (approved):

PR #1409, opened 2023-09-19
SGMM393-547 : RDKServices: RDK service APIs responds to requests with…

vijith-tv:
… invalid security token

Reason for change: Including input security token value fully to validate signature.

Risks: Low


vijith-tv:
bringing in the changes to R2 branch
https://github.com/rdkcentral/Thunder/pull/1344

PR #1414, opened 2023-09-27
[WINDOWS] MAke sure the warning reporting is building under windows a…

pwielders:
…s well!

Take care there is a hard link between selecting warning resporting as a library and the build flags for core in this sommit, the buildflag __CORE_WARNING_REPORTING__ for core needs to be enabled if the warning reporting library is build (otherwise that build will fail!)

### DIFF
--- a/Source/core/WarningReportingControl.h
+++ b/Source/core/WarningReportingControl.h
@@ -19,10 +19,6 @@
 
 #pragma once
 
-#include <inttypes.h>
-
-#include "CallsignTLS.h"
-#include "IWarningReportingControl.h"
 #include "Module.h"
 #include "Optional.h"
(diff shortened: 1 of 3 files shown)


MFransen69 (approved):

VeithMetro (commented):
@pwielders does it actually work on Windows even without the EXTERNAL macro? Because I remember when I tested it on RPI without EXTERNAL, then I got many more controls than intended, which it was causing problems with both displaying the warnings properly and especially with gracefully shutting Thunder down

pwielders:
@VeithMetro at least it compiles :-) Did test it on windows but some other error occured on windows if you use the Warning Reporting, so it still requires more TLC for windows...

PR #1416, opened 2023-10-03
[Actions] [Warning Reporting] Adding Warning Reporting to be built in Actions for Windows

VeithMetro:


pwielders (dismissed):

VeithMetro:
After we disabled warning reporting from being build in core by default on Windows, this feature is on hold for now, since we need to figure out the best way to enabled warning reporting in core on Windows builds only when we want to build warning reporting extension

pwielders (approved):

PR #1425, opened 2023-10-10
[Actions] Changing Windows build to catch the potential fail in a single project

VeithMetro:


VeithMetro:
The last "fix" (using `_write()` instead of `write()` as it is deprecated on Windows and was causing a build error) produced a build error not in Thunder, but in the MessageControl plugin where `Format()` is used. In short, `write()` won't compile under Windows, and apparently `_write()` won't compile under Linux either, so I separated them with `ifdef` to compile on both Windows and Linux

pwielders (approved):

PR #1426, opened 2023-10-10
WebSocket: add masking bytes also while enabling MASKING_FRAME flag

HaseenaSainul:


Source/websocket/WebSocketLink.cpp
```diff
@@ -142,18 +136,22 @@ namespace Web {
...
                     _controlStatus &= (~REQUEST_PING);
-                    dataFrame[result++] = (_setFlags & MASKING_FRAME);
                 }
                 if (((_controlStatus & REQUEST_PONG) != 0) && ((result + 1) < maxSendSize)) {
                     dataFrame[result++] = FINISHING_FRAME | Protocol::PONG;
                     _controlStatus &= (~REQUEST_PONG);
-                    dataFrame[result++] = (_setFlags & MASKING_FRAME);
                 }
+
+                dataFrame[result++] = (_setFlags & MASKING_FRAME);
```
MFransen69: Let's bring this line back to where it was above. We need to set it unconditionally (see next comment), must for now be set to 0 when no masking used, and it will be easier when we later on (I created a Jira issue for it) need to handle the case that these frames can have application data.

Source/websocket/WebSocketLink.cpp
```diff
@@ -142,18 +136,22 @@ namespace Web {
...
+                // Now it seems only control message, hence append with masking keys
+                uint8_t maskKey[4];
```
MFransen69: The following lines may only be executed when the MASKING_FRAME flag is set (so add an if(_setFlags & MASKING_FRAME) != 0) around it.

Also line 135 (in the new version) needs to be changed 
(this one:  if (((_controlStatus & (REQUEST_CLOSE | REQUEST_PING | REQUEST_PONG)) != 0) && ((result + 1) < maxSendSize))

the "result + 1" was always wrong I think and needs to be result + 2 but it needs to be result + 6 when  MASKING_FRAME flag is set to make sure there is enough room for it (so easiest I think is probably to use ((_setFlags & MASKING_FRAME) == 0) ? : ;

MFransen69 (changes requested):

Source/websocket/WebSocketLink.cpp
HaseenaSainul: done

Source/websocket/WebSocketLink.cpp
HaseenaSainul: done

Source/websocket/WebSocketLink.cpp
```diff
@@ -138,7 +132,7 @@ namespace Web {
                 result += usedSize;
             }
 
-            if (((_controlStatus & (REQUEST_CLOSE | REQUEST_PING | REQUEST_PONG)) != 0) && ((result + 1) < maxSendSize)) {
+            if (((_controlStatus & (REQUEST_CLOSE | REQUEST_PING | REQUEST_PONG)) != 0) && (result + ((_setFlags & MASKING_FRAME) ? 6 : 2) < maxSendSize)) {
```
MFransen69: small remark but to make it more inline with our normal coding could you please make it ((_setFlags & MASKING_FRAME) != 0) we do not advocate the use of implicit conversion to boolean

MFransen69 (changes requested):

MFransen69 (approved):

PR #1428, opened 2023-10-12
Core/Thread: fix comparison of different signedness

bramoosterhuis:


pwielders (changes requested):
This is a dangerous fix. The input is 32 bits, so you are allowed to lets, say, pass in 128K + 4K as a value 132K -> 0x21000 => which due to this cast becomes: 0x1000 (16bits signed), it is comapred against PTHREAD_STACK_MIN, which evalueate to true (smaller than 8K minimum size) so the stack is set at 8K while we where expecting 132Kb.
So what you shoulddo to resolve the warning is "enlarge" the PTHREAD_STACK_MIN component of this equations,so:
stackSize < static_cast<uint32_t>(PTHREAD_STACK_MIN)
For these warnings, always cast/move to the largest sized in the equations...

pwielders (approved):

PR #1432, opened 2023-10-20
[docs]: Add 'Time' documentation.

msieben:
METROL-847


docs/utils/time.md
```diff
@@ -0,0 +1,60 @@
...
+
+Thunder models time with instances described by elements of the [Gregorian calendar](https://en.wikipedia.org/wiki/Gregorian_calendar) with its epoch defined by [Unix time](https://en.wikipedia.org/wiki/Unix_time). That is, it models time as (strictly) monotomic increasing integer values. The elements year, month, day, hour, second, and millisecond allow for unique descriptions of Gregorian calendar time instances.
+
+Set values can be individually retrieved. Time instances can be used in calculations and these instances can also be adjusted in millisecond decrements and increments.
+
+Values are depicted with respect to the [UTC timezone](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) unless otherwise specified. Alternatively, a derived convenience model incorporates local time and date offset by the timezone configured for the platform. The user conveniently can ignore any (underlying) offset.
+
+```c++
+// Local time instead of UTC
+Core::Time past(/*year*/ 2023, /*month* 10, /*day*/ 13, /*hours*/ 18, /*minutes*/ 02, /*seconds 0 ,/* milliseconds*/ 0, /*local time*/ true);
```
woutermeek: Missing / in "/*month*"

MFransen69: And missing */ in seconds :)

docs/utils/time.md
```diff
@@ -0,0 +1,60 @@
...
+
+// Internally represent the past with our local timezone instead of UTC
+TimeAsLocal local(past);
+```
+
+In the given example a time object is intialized to October 10 of the year 2023 at 18:02h and a second time object is initialized to the value of the system time. Note that the second object does not specify local time and hence it is silently set to reflect values for the UTC timezone. A final check is executed to test if past time is more than a second ago. For completeness a local time object is created from the object representing the past.
+
+## Date and time conversion options
+
+The external representation of time is a string formatted in a predefined pattern unless a single integer valued element is represented. Such patterns are intrinsically applied for the various conversions methods made available. Among such conversions  are the [RFC2311](https://www.rfc-editor.org/rfc/rfc2311) time and date specification, [RFC1036](https://www.rfc-editor.org/rfc/rfc1036) date specification, the [ISO8601]() date and time formats, and, [ANSI](https://webstore.ansi.org/standards/incits/ansiincits301997) date representation.
```
MFransen69: Perhaps good to explicitly  indicate we have both from and to conversions for these formats

MFransen69 (changes requested):

MFransen69 (approved):

PR #1433, opened 2023-10-20
Development/json types

msieben:


Source/core/JSON.h
```diff
@@ -4000,13 +4682,187 @@ namespace Core {
...
+
+                                    if ((typeStartPos + 1) < typeEndPos) {
+                                        // Determine type of contained element
+// FIXME: Inefficient
+                                        element = TypeEstimate(value.substr(typeStartPos + 1, typeEndPos - typeStartPos - 1), type);
+                                        delete element;
+
+// FIXME: Currently fallback to underlying String instead of actual type
+                                        switch(depth) {
+                                        case 1  : element = new ArrayType<String>;
```
pwielders: What issue is resolved by these allocations. Allocations definitely on these low levels of functionality, might cause fragmentation of memory. So far no issues have been reported that require this allocation so would like to understand the need of this construction. What is the use-case that gives erroneous output without this allocation construction?

pwielders (changes requested):
Would like to avoid allocation using new. So far it was never needed hence my quesion what are we fixing that requires this ?

Source/core/JSON.h
msieben: As discussed with @MFransen69 the code path is there not to resolve any reported issue although it shows fields in config.json have no corresponding IElement in the contained list. For debugging purposes and to simplify unit testing for now it has been quite helpful. The use of Add(..) prior Deserialize(..) and ignoring unmatched fields is preferred. A proposed (recommended) approach is to remove the code path from a release delivery and save it elsewhere for use with the aforementioned purposes.

msieben:
Closing the PR but keeping the related branch.

PR #1434, opened 2023-10-20
MessageControl: made dispatcher path as configurable

HaseenaSainul:


Source/WPEFramework/Config.h
```diff
@@ -357,6 +357,7 @@ namespace PluginHost {
 #endif
                 , ProxyStubPath()
                 , PostMortemPath(_T("/opt/minidumps"))
+                , MessageControlPath(_T("/tmp/MessageDispatcher"))
```
MFransen69: The configurable part of the path should be relative to the volatile path, not an absolute path

Source/WPEFramework/Config.h
```diff
@@ -400,6 +401,7 @@ namespace PluginHost {
                 Add(_T("volatilepath"), &VolatilePath);
                 Add(_T("proxystubpath"), &ProxyStubPath);
                 Add(_T("postmortempath"), &PostMortemPath);
+                Add(_T("messagecontrolpath"), &MessageControlPath);
```
MFransen69: Let's add this to the messaging section so that we group all messaging settings, then it is clear they all belong together.
(and then  you don't have to forward these settings seperately in the Open of tghe MessageControlUnit)
And as we are at this let's also move messaging port to that section (it is now in line 423), it was an oversight it was not moved before :) (and btw you don;t have to worry that it lead to backwards compatibility problems as messagingport is only applicable to windows)

Source/WPEFramework/GenericConfig.cmake
```diff
@@ -30,6 +30,7 @@ set(WEBSERVER_PATH "/boot/www" CACHE STRING "Root path for the HTTP server")
 set(WEBSERVER_PORT 8080 CACHE STRING "Port for the HTTP server")
 set(PROXYSTUB_PATH "${CMAKE_INSTALL_PREFIX}/lib/${NAMESPACE_LIB}/proxystubs" CACHE STRING "Proxy stub path")
 set(POSTMORTEM_PATH "/opt/minidumps" CACHE STRING "Core file path to do the postmortem of the crash")
+set(MESSAGECONTROL_PATH "/tmp/MessageDispatcher" CACHE STRING "MessageControl base path to create message files")
```
MFransen69: So this should also not include the volatile path (/tmp)

Source/WPEFramework/PluginHost.cpp
```diff
@@ -584,7 +584,7 @@ POP_WARNING()
             // Time to open up, the message buffer for this process and define it for the out-of-proccess systems
             // Define the environment variable for Messaging files, if it is not already set.
             uint32_t messagingErrorCode = Core::ERROR_GENERAL;
-            messagingErrorCode = Messaging::MessageUnit::Instance().Open(_config->VolatilePath(), _config->MessagingPort(), messagingSettings, _background, options.flushMode);
+            messagingErrorCode = Messaging::MessageUnit::Instance().Open(_config->MessageControlPath(), _config->MessagingPort(), messagingSettings, _background, options.flushMode);
```
MFransen69: So you would still pass on the Volatile path but the MessageControlPath and MessagingPort will be part of messageSettings so are not a seperate parameter

Source/messaging/MessageUnit.cpp
```diff
@@ -145,18 +145,29 @@ namespace WPEFramework {
         {
             uint32_t result = Core::ERROR_OPENING_FAILED;
 
-            string basePath = Core::Directory::Normalize(pathName) + _T("MessageDispatcher");
+            uint32_t permission = 0;
```
MFransen69: Code for getting the rights is already here: https://github.com/rdkcentral/Thunder/blob/3249f7cae8588a8d7022f03edfc9d9c0e412ed43/Source/core/NodeId.cpp#L373

Let's not have this mutliple times. I prose moving the rights code from NodeId.cpp to somewhere else (I think FileSystem might be a good place). Then here as well in NodeId we use the same code to extract the access rights)

Source/messaging/MessageUnit.cpp
```diff
@@ -145,18 +145,29 @@ namespace WPEFramework {
...
+            }
+
+            string basePath = Core::Directory::Normalize(pathName.substr(0, position));
             string identifier = _T("md");
 
             ASSERT(_dispatcher == nullptr);
 
             if (Core::File(basePath).IsDirectory() == true) {
                 //if directory exists remove it to clear data (eg. sockets) that can remain after previous run
                 Core::Directory(basePath.c_str()).Destroy();
```
MFransen69: I think we should no longer destroy the folder to empty it (suppose somebody created it from the outside perhaps with specific settings/rights) but delete all the md files from it. And then we should only create it when it doesn't exists and only set the rights when we create it.

MFransen69 (changes requested):

HaseenaSainul:
Thanks @MFransen69 for the comments. will check and update

Source/WPEFramework/Config.h
HaseenaSainul: moved both to messaging config

Source/WPEFramework/PluginHost.cpp
HaseenaSainul: updated

Source/messaging/MessageUnit.cpp
HaseenaSainul: moved to FileSytem.h

Source/messaging/MessageUnit.cpp
HaseenaSainul: current implementation will not destroy folder if it endue with '/'. During Directory construction call it get normalised with '/'.

Source/messaging/MessageUnit.cpp
```diff
@@ -139,32 +139,33 @@ namespace WPEFramework {
...
 
-            if (Core::File(basePath).IsDirectory() == true) {
+            if (Core::File(_settings.BasePath()).IsDirectory() == true) {
                 //if directory exists remove it to clear data (eg. sockets) that can remain after previous run
-                Core::Directory(basePath.c_str()).Destroy();
-            }
-
-            if (Core::Directory(basePath.c_str()).CreatePath() == false) {
-                TRACE_L1("Unable to create MessageDispatcher directory");
+                Core::Directory(_settings.BasePath().c_str()).Destroy();
```
MFransen69, line 154: Think we now still remove the whole folder to clear it correct or am I missing something? (and because of the else we do not recreate it again)

MFransen69 (changes requested):

MFransen69 (approved):

PR #1438, opened 2023-10-25
[DOC] Scheduler section

pczekaj99:
Scheduler section added inside Timers section


MFransen69:
Let's close this PR and create a new one that integrates the Scheduler documentation in the WorkerPool section with improved examples

PR #1447, opened 2023-11-08
[Core] Print that secure proxy/stub are used

pczekaj99:


pwielders:
The code is fine, just think we should try to move it to the WPEFramework module, if possible. Lets discuss it next week..

sebaszm (commented):
Perhaps I recall wrongly, but wasn't this also supposed to check if there is a mixture of flags between loaded proxystubs?

sebaszm (changes requested):
Needs to be moved as per Pierre's comment and made to verify flag coherency across loaded stubs.

PR #1448, opened 2023-11-09
JSON: fix the deserialization of JSON::ArrayType

arjun-tata:
In the context of the JSON::ArrayType type, a previous change was added as part of this PR -> https://github.com/rdkcentral/Thunder/pull/1112 for the serialization of empty array ie: to differentiate, between a set but empty and a null JSON array container. The resulting implementation, however, failed to deserialize properly, because when finding a empty JSON array (i.e. `someArray: []`), the resulting object would be equivalent to `someArray: null`, which is incorrect.

To fix this, this commit adds the setting of the `_state` data member in the `Deserialize` method when a matching pair of square brackets are correctly parsed.

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -3305,6 +3305,7 @@ namespace Core {
                         if (loaded < maxLength) {
                             switch (stream[loaded]) {
                             case ']':
+                                _state |= (modus::SET);
                                 offset = FIND_MARKER;
                                 loaded++;
                                 break;


pwielders (approved):

pwielders:
@arjun-tata good catch and thank you for the contribution. Thanks for adding this to the master and the detailed commit message, this made the review easy!
Do you need this on any specific release as well so we can cherry-pick it there ?

rjpdasilva:
> @arjun-tata good catch and thank you for the contribution. Thanks for adding this to the master and the detailed commit message, this made the review easy! Do you need this on any specific release as well so we can cherry-pick it there ?

Hi @pwielders, yes, it would be good for us to get this on `R2` as well, but for that, you'll need not only this change, but also the one from https://github.com/rdkcentral/Thunder/pull/1112 to go there as well, and first.

arjun-tata:
> > @arjun-tata good catch and thank you for the contribution. Thanks for adding this to the master and the detailed commit message, this made the review easy! Do you need this on any specific release as well so we can cherry-pick it there ?
> 
> Hi @pwielders, yes, it would be good for us to get this on `R2` as well, but for that, you'll need not only this change, but also the one from #1112 to go there as well, and first.

hello @pwielders , as mentioned could you please help to migrate these two changes to the `R2` branch as well, it will be helpful to us.

PR #1456, opened 2023-11-22
SecureSocketPort: Enable Server Name Indication (SNI) by default

rjpdasilva:
When establishing TLS connections, the WPEFramework's SecureSocketPort implementation, based on OpenSSL, is not enabling the SNI (Server Name Indication [1]) extension by default. Most server implementations do require SNI, i.e., that the client, when doing the TLS handshake, does provide information about which server is handshaking to, so that it can then provide a suitable certificate. This is particularly true for cloud based services. To comply with these, we should enable SNI by default. Enabling this, does not create any problems to servers that do not support or require SNI, so no harm done in enabling SNI by default.

Enabling the TLS SNI feature with OpenSSL can be done by calling the `SSL_set_tlsext_host_name()` ([2]) on the associated SSL context, with the host name we're trying to connect to, so that the host name is then set on the "server_name" extension in the "Client Hello" TLS message.

This way, in `SecureSocketPort::Handler::Initialize()`, which is where the SSL context is created and prepared, make the required call to `SSL_set_tlsext_host_name()` with the host name that is configured on the remote node associated with the `SecureSocketPort` instance.

For supporting this, extend the `Core::NodeId` class so that it "remembers" the original host name provided when creating nodes for representing IP connections to a remote host, which is the case of the TLS connections. For this, add a new `m_remoteHostName` member and an associated `RemoteHostName()` getter. There was already a `m_hostName` member and a `HostName()` getter which seem to suggest this feature was already present in `Core::NodeId`. However, further inspection and debugging shows that in fact this is not always the case and it is used for different purposes, hence the decision to use a new member/getter, to ensure the new needs and not breaking any current uses of `m_hostName`.

[1] https://en.wikipedia.org/wiki/Server_Name_Indication
[2] https://www.openssl.org/docs/man1.1.1/man3/SSL_get_servername_type.html

[RDKDEV-886](https://jira.rdkcentral.com/jira/browse/RDKDEV-886)

### DIFF
--- a/Source/core/NodeId.cpp
+++ b/Source/core/NodeId.cpp
@@ -224,6 +224,7 @@ namespace Core {
         const uint16_t nPortNumber,
         const enumType defaultType)
         : m_hostName()
+        , m_remoteHostName(strHostName)
     {
 
         m_structInfo.IPV4Socket.sin_family = TYPE_UNSPECIFIED;
@@ -241,6 +242,7 @@ namespace Core {
         const TCHAR strHostName[],
         const enumType defaultType)
         : m_hostName()
+        , m_remoteHostName(strHostName)
     {
 
         m_structInfo.IPV4Socket.sin_family = TYPE_UNSPECIFIED;
@@ -368,6 +370,7 @@ namespace Core {
         memcpy(&m_structInfo, &rInfo.m_structInfo, sizeof(m_structInfo));
 
         m_hostName = rInfo.m_hostName;
+        m_remoteHostName = rInfo.m_remoteHostName;
 
         // Give back our-selves.
         return (*this);
@@ -554,6 +557,12 @@ namespace Core {
         }
 
         return (m_hostName);
+    }
+
+    string
+    NodeId::RemoteHostName() const
+    {
+        return (m_remoteHostName);
     }
 
     void NodeId::HostName(const TCHAR strHostName[])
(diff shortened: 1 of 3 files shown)


rjpdasilva:
Hi, any progress on reviewing this PR?

pwielders (approved):

pwielders:
@rjpdasilva as this is R2, I think you need to trigger  @anand-ky  or @karuna2git to get this merged, and potentially all other PR's pending for R2. 
I think Comcast/RDK work with a patch files and will probably not merge it and probably create a patch for it. This is not recommended by us, but it is the way of working for RDK/Comcast!

rjpdasilva:
@anand-ky, @karuna2git,

Any chance to follow up on this please?

karuna2git:
Changes looks good & merging it. 

Could you please bring the change to master also.? So that future releases will automatically pick this up.

rjpdasilva:
Hi @karuna2git,

Thanks for merging.

Regarding:
> Could you please bring the change to master also.? So that future releases will automatically pick this up.

Last time I've checked, the `master` branch already deals properly with SNI, but it was implemented in a slightly different way and had some dependencies that could not be met in `R2`, hence I've implemented this simpler "version" of it without further dependencies. So, I don't think it is needed to port this for the `master` branch, since it's already implemented there as well.

karuna2git:
Got it.. Thanks 👍

PR #1458, opened 2023-11-27
[Messaging] Fixing warning reporting config (when messaging config comes from a separate file)

VeithMetro:
This restores the functionality not present for Warning Reporting, namely reading messaging config from a separate file, requested in this ticket: https://jira.rdkcentral.com/jira/browse/METROL-882

@pwielders at the moment, a file in the config has to be specified as just its name (not the whole path, like it has been described in our documentation: https://rdkcentral.github.io/Thunder/plugin/messaging/#configuration [scroll a bit down for `Tip`]). So to work properly, it has to be for example like this: `"messaging":"msg.json"`.

Do we want to change the documentation (mention that the file resides at `/etc/WPEFramework/`), or maybe we want users to specify the whole path, but this could potentially break some builds?


pwielders:
@VeithMetro it has to be a full path as /etc/wpeframework typically resides in ROM and this file is preferred to be in a wtritable area to symply change stuff at startup. Lets do a skype call.

pwielders (approved):

PR #1465, opened 2023-12-06
[Doc] Workerpool documentation

pczekaj99:


docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
 ## Conclusions
 
 To sum up, the main idea is not to reinvent the wheel. When creating plugins, developers should keep in mind they are working on a large system in an embedded environment. Because of that from an architectural point of view, a different set of rules applies than when working on developing PC or even mobile applications. We all have to be aware of limitations like the low amount of memory available, the difficulty of keeping a system with dozens of plugins scalable, or even the necessity to use abstractions to achieve portability. These are the main reasons why it is essential to use functionalities that are already given, instead of making things suboptimally on your own.
+
+## Scheduler
+### Overview
+`Workerpool` has the ability to schedule `jobs` using the `Scheduler` class. This can be very helpful when you do not want to start a `job` immediately but at a specific time. To learn more about how to create a `job` read [Workerpool](../threading/worker-pool) section of documentation.
```
VeithMetro: `To learn more about how to create a job read [Workerpool](../threading/worker-pool) section of documentation.` - this does not make much sense here (this is already WorkerPool's documentation 😄)

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+
+* `Schedule()`
+* `Reschedule()`
+* `Revoke()`
+
+Using these methods, we can schedule the performance of a task, change the date on which it will be performed or cancel it altogether. 
+### Scheduler methods
+
+#### Schedule
+`Schedule()` is used to plan when to perform the `job` we specified. Using this method is very simple. We specify the `job` and the date on which to start performing it in the appropriate format.
```
VeithMetro: 1. Let's not start each section with the name of the method if it is already in the section's title - change it to something like this method/functionality/mechanism etc.
2. Maybe instead of "We specify the `job` and the date on which to start performing it in the appropriate format", we could change it to something like: "We just need to specify the `job` with a point in time when it should be performed." - I think it sounds more clear, what do you think?

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+            }
+            else {
+                _threadPool.Submit(job, Core::infinite);
+            }
+        }
+```
+
+
+#### Reschedule
+`Reschedule()` method is used to change the execution date of an already scheduled `job`. It is used in the same way as the `Schedule()` method. It works as follows, first the scheduled `job` is attempted to be canceled and then the Schedule with the newly specified time is used.
```
VeithMetro: Maybe instead of "It is used in the same way as the `Schedule()` method", we could be more literal and say that it takes the same parameters, since it is probably what you meant 😃

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+            Schedule(time, job);
+            return (revoked);
+        }
+```
+
+
+#### Revoke
+`Revoke()` is used to cancel the execution of a `job`. We use it by specifying the `job` we want to cancel and the time after which it should be cancelled. We do not need to specify the time, if we leave the default value, the `job` will be canceled immediately. The `Reschedule` method returns a numeric value.
+
+ Error codes that can be returned:
```
VeithMetro: 1. I don't think writing down the error codes is necessary, as we already have a warning that "`ERROR_UNKNOWN_KEY` will be returned in case the job was not previously scheduled!", which seems enough
2. Are you sure this method can only return these two errors? Maybe it is a good idea to have a look at Revoke in ThreadPool to see what it returns, and in the warning section add something about other error codes which can be returned and what are the conditions for them

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+
+            bool revoked = (Revoke(job) != Core::ERROR_UNKNOWN_KEY);
+            Schedule(time, job);
+            return (revoked);
+        }
+```
+
+
+#### Revoke
+`Revoke()` is used to cancel the execution of a `job`. We use it by specifying the `job` we want to cancel and the time after which it should be cancelled. We do not need to specify the time, if we leave the default value, the `job` will be canceled immediately. The `Reschedule` method returns a numeric value.
```
VeithMetro: 1. Are you sure that in Revoke we pass "time after which it [job] should be cancelled"?
2. So if we leave the default value for the waitTime, then the `job` will be canceled immediately? By default we are passing `Core::infinite`, so it sounds like we will be waiting for as long as it takes to Revoke it.
3. Maybe add at the end that these numeric values are error codes just to be extra blunt

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+                    result = report;
+                }
+            }
+
+            return (result);
+        }
+```
+
+### Example
+`Scheduler` is used for example in rdkservices repository. Here is example use of `Schedule()` from there.
```
VeithMetro: 1. It's good you added an example, but are you sure that the Schedule which is used here is the same Schedule that you described above? 😄
2. I think it would be better to find some examples in https://github.com/rdkcentral/ThunderNanoServices, and then maybe try to write down a few sentences about it - we can all see that the Schedule was used here (maybe not the one from Scheduler, but anyway... 😜), but let's write about why it was used and to what end
3. Try to find at least one example (in the actual code not tests) for each of the methods which you described, not just Schedule

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,129 @@ functionality in Thunder, which will be further described in a different documen
...
+            {
+                FrontPanelBlinkInfo blinkInfo = m_blinkList.at(0);
+                setBlinkLed(blinkInfo);
+                if (m_isBlinking)
+                    blinkTimer.Schedule(Core::Time::Now().Add(blinkInfo.durationInMs), m_blinkTimer);
+            }
+        }
+```
+
+Below is test example of using it.
```
VeithMetro: I think it would be better to find actual examples in code where it is used, it must be used at least a few times in  https://github.com/rdkcentral/ThunderNanoServices, and then we you give an example from an actual code, you can write about why it was used here and how it is actually useful in practice and not just in theory.

VeithMetro (changes requested):
It is a good start, but we need to improve it a bit before merging it with WorkerPool 😄 
I left some remarks in the comments, try to work on them and then I'll review it again. Feel free to ask in case of any questions!

VeithMetro (changes requested):
Like we discussed on Skype, there are still some things to improve plus we have to give examples of `Reschedule()` from `JobType` class and describe it, as well as focus on `Schedule()` from the `Scheduler` class, since this is not used much and we should shed some light on that

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+
+## Scheduling jobs
+### Overview
+`Workerpool` has the ability to schedule `jobs`. This can be very helpful when you do not want to start a `job` immediately but at a specific time. `Workerpool` has several ways of scheduling `jobs`. The first is to use the `Workerpool` class methods. Another way is to use the `Scheduler` class. The most common case of scheduling tasks is to refer directly to a specific `job` and use its methods.
+
+To schedule a `job` you should use the `Scheduler` class or a direct reference to a specific `job`. The methods of the `Workerpool` class are described below mainly for illustrative purposes and to see how jobs are scheduled inside `Workerpool`.
```
VeithMetro: I think there is no need to mention the exact same thing twice - in the first paragraph you say that _"`Workerpool` has several ways of scheduling `jobs`. The first is to use the `Workerpool` class methods. Another way is to use the `Scheduler` class."_, and then you say _"To schedule a `job` you should use the `Scheduler` class or a direct reference to a specific `job`."_. Remove one of these fragments

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+            }
+            else {
+                _threadPool.Submit(job, Core::infinite);
+            }
+        }
+```
+
+
+#### Rescheduling job
+`Reschedule()` method is used to change the execution date of an already scheduled `job`. Method takes as arguments the new execution time of the specified `job` and the `job` whose execution time we want to change. It works as follows, first the scheduled `job` is attempted to be canceled and then the Schedule with the newly specified time is used.
```
VeithMetro: _"...and the `job` whose execution time we want to change."_ - job is no a person, so it should be "which" not "whose" 😄

docs/utils/threading/worker-pool.md
VeithMetro: > 1. I don't think writing down the error codes is necessary, as we already have a warning that "`ERROR_UNKNOWN_KEY` will be returned in case the job was not previously scheduled!", which seems enough
> 2. Are you sure this method can only return these two errors? Maybe it is a good idea to have a look at Revoke in ThreadPool to see what it returns, and in the warning section add something about other error codes which can be returned and what are the conditions for them

I think you missed this comment, as it looks like you did not make changes to this section

VeithMetro: And also the last bullet point from another one of my comments seems to be missed: https://github.com/rdkcentral/Thunder/pull/1465#discussion_r1422424693

docs/utils/threading/worker-pool.md
VeithMetro: I thought we agreed that we are going to remove the example from this test case, since this is just calling the methods for a sole reason to test them, not an actual use case. I think we should stick to including example from actual plugins like from https://github.com/rdkcentral/ThunderNanoServices, and the key is not to just paste the code fragments, but also describe them a bit (and here with a test case we cannot do that since there is nothing to describe)

VeithMetro: Here is a link to my comment from a previous review, check points 2 and 3: https://github.com/rdkcentral/Thunder/pull/1465#discussion_r1422457945

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+
+    // Revoke the job (cancel its execution)
+    _workerpool.Revoke(job);
+}
+```
+For more examples you can check `Thunder/Tests/unit/core/test_workerpool.cpp` file.
+
+
+### Scheduler class
+`Scheduler` is a small class designed to schedule jobs in the future. It has one public method `Schedule()`. It takes as arguments the `job` whose execution is to be scheduled in the future and the time after which the execution of the `job` should start.
```
VeithMetro: _"...the `job` whose..."_ - same story as above 😃

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+```cpp
+void Schedule(const Time& time, const ProxyType<IDispatch>& job) override {
+                _timer.Schedule(time, Timer(_pool, job));
+            }
+```
+The Scheduler uses the `Timer` class of the `Workerpool` class as a privileged class. How the `Timer` class is built and how it works can be seen in the `Timer.h` file.
```
VeithMetro: 1. I'm not sure I understand this properly - _"The Scheduler uses the `Timer` class of the `Workerpool` class as a privileged class."_ - maybe we could change it to _"The Scheduler uses a private `Workerpool` class called `Timer`."_ - does this carry the same meaning or have you meant something else?
2. _"How the `Timer` class is built and how it works can be seen in the `Timer.h` file."_ - it looks to me that in Timer.h we have a templated TimerType class, so we should be more exact here - mention that Timer (the private class of Workerpool) is used as a templated parameter of TimerType, which can be found in Timer.h etc.

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+
+### Job class scheduling methods
+Each `job` has two methods by which we can reschedule them for future execution and cancel them.
+ These methods are:
+
+* `Reschedule()`
+* `Revoke()`
+
+#### Rescheduling job
+`Reschedule()` allows the scheduled `job` execution time to be changed. The id argument takes the time by which we should postpone the execution of the job. The method returns `true` if the operation is successful.
```
VeithMetro: 1. What do you mean by _"The **id** argument"_?
2. I think that the argument is just a point in time, so we do not postpone the job **by** this time, but **until** this time. Example: the job is scheduled to be done today at 3pm, we can either postpone it by an hour, and then it will be done at 4pm today, but we can also postpone it until 4pm today - same outcome, but different ways of doing it

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+```cpp
+bool Reschedule(const Core::Time& time)
```
VeithMetro, line 198: If we include a piece of code, I'd write a few sentences of what is going on in this code

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+uint32_t Open(const uint32_t waitTime, const uint16_t connectionCheckTimer) {
+                _connectionCheckTimer = connectionCheckTimer * 1000;
+                if (connectionCheckTimer != 0) {
+                    _job.Reschedule(Core::Time::Now().Add(_connectionCheckTimer).Ticks());
+                }
+                return(BaseClass::Open(waitTime));
+            }
+```
+
+`Reschedule` method is used also in `ThunderNanoServices` repository. Here is few examples:
```
VeithMetro: _"Here is few examples:"_ - it think it should be: _"Here are a few examples:"_

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+                }
+
+                return (rescheduled && (ThreadPool::JobType<IMPLEMENTATION>::IsIdle() == false));
+            }
+```
+!!!warning
+    The `Job` must first be submitted to be rescheduled!
+
+#### Revoking job
+`Revoke()` is used to revoke the execution of `job`.
```
VeithMetro: `Revoke()` is already mentioned and talked about in the Workerpool documentation, so no point including it once again: https://github.com/rdkcentral/Thunder/blob/master/docs/utils/threading/worker-pool.md?plain=1#L87

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,217 @@ functionality in Thunder, which will be further described in a different documen
...
+// See if we need rescheduling
+        if (result != Core::infinite) {
+            Core::Time timestamp(Core::Time::Now());
+            timestamp.Add(result);
+            _job.Reschedule(timestamp);
+```
+
+`ProcessMonitor/ProcessMonitor.h`
+```cpp
+void ScheduleJob()
```
VeithMetro, line 254: 1. It is really nice that you include example from actual use cases, much better than including tests which only call the methods! But once again - if you include any code examples, please write a few sentences of what is going on in this code.
2. The main purpose to this documentation was to create a section for Scheduler. Maybe I am mistaken, but I don't think you included any use examples from Scheduler, only examples of methods from the templated class `JobType`, which is already talked about in the Workerpool documentation: https://github.com/rdkcentral/Thunder/blob/master/docs/utils/threading/worker-pool.md?plain=1#L51

VeithMetro (changes requested):
Good changes, but I still think there are a few things which need to be changes - look at my comments, please respond to each of them, and reach out to me in case of any questions

docs/utils/threading/worker-pool.md
pczekaj99: Fragment removed

docs/utils/threading/worker-pool.md
pczekaj99: Done

docs/utils/threading/worker-pool.md
pczekaj99: 1. It was typo
2. Done

docs/utils/threading/worker-pool.md
pczekaj99: Done

docs/utils/threading/worker-pool.md
pczekaj99: Section removed

docs/utils/threading/worker-pool.md
pczekaj99: 1. Descriptions added

docs/utils/threading/worker-pool.md
pczekaj99: Done

docs/utils/threading/worker-pool.md
pczekaj99: 1. I will change it as you proposed

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    public:
+    enum Status {
+         INITIATED,
+         CANCELED,
+         COMPLETED,
+    };
+
+    TestJob() = delete;
+    TestJob(const TestJob& copy) = delete;
```
MFransen69, line 11: I know it is technically not needed but in new code we also explicitly delete the move c'tor and move operator if the copy c'tor and assignment operator are deleted (for consistency and to make sure they are not forgotten when the copy and assignment are enabled; otherwise there is a automatic fallback to copying when the object is moved)

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    TestJob& operator=(const TestJob& RHS) = delete;
+    ~TestJob() override = default;
+    TestJob(const Status status, const uint32_t waitTime = 0,)
+        : _status(status)
+        , _waitTime(waitTime)
+    {
+    }
+
+public:
+    Status GetStatus()
```
MFransen69: Think this method could be const

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {
+        return _status;
+    }
+    void Cancelled()
```
MFransen69: Although naming is rather subjective I think this name is confusing. It now looks like a question, a getter for canceled state but it is actually triggering the cancellation if I'm right. So would a verb like "Cancel" not be a better fitting name?

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+         INITIATED,
+         CANCELED,
+         COMPLETED,
+    };
+
+    TestJob() = delete;
+    TestJob(const TestJob& copy) = delete;
+    TestJob& operator=(const TestJob& RHS) = delete;
+    ~TestJob() override = default;
+    TestJob(const Status status, const uint32_t waitTime = 0,)
```
MFransen69: Isn't it a bit unexpected that you can pass the initials state here? What if I pass the state completed? :)

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+public:
+    Status GetStatus()
+    {
+        return _status;
+    }
+    void Cancelled()
+    {
+        _status = (_status != COMPLETED) ? CANCELED : _status;
```
MFransen69: Reading the code in the Dispatch the job does not really have a purpose, right? So what is then the purpose of the sleep (and as we advocate to use sleeps as much as possible perhaps better to not add one in our code examples :) )
Also think about synchronization as Dispatch can run in parallel with this method which can lead to interesting results (e.g. if thread one runs just to (_status != COMPLETED) and the status is at that moment still INITIATED and there is a context switch leading to thread 2 being executed which executes the Dispatch leading to status being completed and then thread 1 continues setting the state to CANCELED  and that is what you are functionality wise to prevent in the Cancelled method)

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+    void Dispatch() override
+    {
+        _status = COMPLETED;
```
MFransen69: Think it would be better if this call actually tries do do something. Keeping a state might not be that useful as that is basically done in the workerpool internally. This will also make it easier to write some code that has correct thread synchronization. 
As an idea you could make it calculate 'n' numbers (input parameter) of the Fibonacci sequence (https://en.wikipedia.org/wiki/Fibonacci_sequence) or perhaps if you want it more thunder context like pretent it does some kind of decryption of some data

Tests/workerpool-scheduling/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // Now we are creating and submiting job
+    Core::ProxyType<Core::IDispatch> job_one = Core::ProxyType<Core::IDispatch>(Core::ProxyType<TestJob>(TestJob::Status INITIATED, 0));
+    workerpool.Submit(job);
+
+    // Print job status to check if it is initiated
+    std::cout << "Job status: " << job.GetStatus() << std::endl;
+
+    // Now run the job and wait for it to complete
+    workerpool.Run();
+    std::this_thread::sleep_for::(std::chrono::seconds(2));
```
MFransen69: As an idea use an event mechanism or c++ futures instead of waiting? And if we keep the waiting I propose we use the Core Time instead of std:: chrono as we want to advocate people using the Thunder abstractions

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+The methods of the `Workerpool` class are described below mainly for illustrative purposes and to see how jobs are scheduled inside `Workerpool`.
+
+### Workerpool class methods
+ To schedule a `job` using `Workerpool` class we use several methods:
+
+* `Schedule()`
+* `Reschedule()`
+* `Revoke()`
+
+Using these methods, we can schedule the performance of a task, change the date on which it will be performed or cancel it altogether.
```
MFransen69: Perhaps time would be better than date here

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+
+#### Scheduling job
+`Schedule()` is used to plan when to perform the `job` we specified. Using this method is very simple. We just need to specify the `job` with a point in time when it should be performed.
+!!! note
+    In case the given time would point to the past, the task will start executing immediately.
+
+```cpp
+void Schedule(const Core::Time& time, const Core::ProxyType<IDispatch>& job) override
```
MFransen69: Think it would be better to not show the workerpool code at all, but refer to the eample

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+```
+
+
+#### Rescheduling job
+`Reschedule()` method is used to change the execution date of an already scheduled `job`. Method takes as arguments the new execution time of the specified `job` and the `job` which execution time we want to change. It works as follows, first the scheduled `job` is attempted to be canceled and then the Schedule with the newly specified time is used.
+!!! note
+    `Reschedule()` returns true when the `job` cancellation succeeds and false when it fails, this can happen when the given task was not previously scheduled using `Schedule()`
+
+```cpp
+bool Reschedule(const Core::Time& time, const Core::ProxyType<IDispatch>& job) override
```
MFransen69: see above

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+                _threadPool.Submit(job, Core::infinite);
+            }
+        }
+```
+
+
+#### Rescheduling job
+`Reschedule()` method is used to change the execution date of an already scheduled `job`. Method takes as arguments the new execution time of the specified `job` and the `job` which execution time we want to change. It works as follows, first the scheduled `job` is attempted to be canceled and then the Schedule with the newly specified time is used.
+!!! note
+    `Reschedule()` returns true when the `job` cancellation succeeds and false when it fails, this can happen when the given task was not previously scheduled using `Schedule()`
```
MFransen69, line 176: Think a more logical explanation is when the job has already been executed

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+
+* `Reschedule()`
+* `Revoke()`
+
+#### Rescheduling job
+`Reschedule()` allows the scheduled `job` execution time to be changed. The argument takes the time value. We postpone the execution of the job until this time. The method returns `true` if the operation is successful.
+```cpp
+bool Reschedule(const Core::Time& time)
+            {
+                bool rescheduled = false;
```
MFransen69, line 200: Think again this is code from WorkerPool itself right? Better to not have it here at all?

docs/utils/threading/worker-pool.md
```diff
@@ -148,3 +148,309 @@ functionality in Thunder, which will be further described in a different documen
...
+                }
+            }
+
+            if (scheduleTime != 0) {
+                _job.Reschedule(scheduleTime);
+            }
+        }
+```
+
+Here is example how to use workerpool methods to schedule jobs, code can be compiled and is part of Thunder tests in `Tests` directory. Example contains two tests, first one shows how to create and submit single job, second test is more complex. At first we create three jobs, submit one and schedule the rest. After this we reschedule first job and run everything. Third job is revoked before starting it's execution. At last we print all job's status. Job one and two should be `COMPLETED`, job three should be `CANCELED`.
```
MFransen69: See review of the code above

MFransen69 (changes requested):

PR #1471, opened 2023-12-16
[FIXES] Addressed review comments from pull request and used new Gene…

pwielders:
…rator tool functionality.

### DIFF
--- a/Source/core/JSONRPC.h
+++ b/Source/core/JSONRPC.h
@@ -80,19 +80,19 @@ namespace Core {
                 void SetError(const uint32_t frameworkError)
                 {
                     switch (frameworkError) {
-                    case Core::ERROR_BAD_REQUEST:
+                    case Core::ERROR_INTERNAL_JSONRPC:
                         Code = -32603; // Internal Error
                         Text = _T("Unknown method.");
                         break;
-                    case Core::ERROR_INVALID_DESIGNATOR:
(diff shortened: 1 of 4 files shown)


Source/core/JSONRPC.h
```diff
@@ -108,36 +108,36 @@ namespace Core {
...
                         Text = _T("Call timed out.");
                         break;
-                    case Core::ERROR_PARSE_FAILURE:
+                    case Core::ERROR_PARSING_ENVELOPPE:
                         Code = -32700; // Parse error
                         Text = _T("Parsing of the parameters failed");
                         break;
                     case Core::ERROR_INVALID_RANGE:
-                        Code = Core::ERROR_INVALID_RANGE;
+                        Code = -33000 - Core::ERROR_INVALID_RANGE;
```
MFransen69: Perhaps make a const(expr) out of the 33000 with a meaningful name so it is more clear why we use it as a base for the application errors? If you look in the json rpc spec (as I was curious why it was 33000 anyway :) ) it looks like the codes were written with a signed 16 bits code in mind, but if 32 bit can be used indeed then this should be correct as well.

MFransen69 (changes requested):

MFransen69 (approved):

sebaszm (approved):

PR #1473, opened 2023-12-18
[CMake] Fixing an issue with propagating build type from Thunder to other modules

VeithMetro:


sebaszm (approved):

bramoosterhuis:
@VeithMetro Why is this added again?  I removed it on purpose a few weeks ago, because is should be provided and set by the build system.

https://github.com/rdkcentral/Thunder/commit/74eb7879fb08d5c2386585206c15f66fe36f63aa#diff-aa6caeba4876cc0465723ecd98e5fe0744a4e315ef30784528eaef8f472f5286L79

PR #1476, opened 2023-12-19
Revert "[CMake] Fixing an issue with propagating build type from Thunder to other modules"

bramoosterhuis:
Reverts rdkcentral/Thunder#1473


VeithMetro:
If this line has to be removed (again), then maybe let's first figure out why without it even if Thunder builds in Debug mode on Buildroot, everything else (Interfaces, Plugins, ClientLibs etc.) still always builds in Release

pwielders:
@bramoosterhuis if this is currently causing a build issue, we need to properly fix it and not just revert it. Can you introduce a proper solution than ?

PR #1477, opened 2023-12-19
Moving doc folder to WebPlatformForEmbedded/Documentation in order to…

volkan-aslan:
… avoid confusion by having 2 doc folders (doc and docs)


mkdocs.yml
```diff
@@ -67,6 +67,7 @@ nav:
         - COM-RPC: walkthrough/hello-world/client-comrpc.md
   - Misc:
     - Documentation: docs.md
+    - Reference: https://github.com/WebPlatformForEmbedded/Documentation/blob/master/Thunder/WPE-API-WPEFramework/WPE-API-WPEFramework.pdf
```
MFransen69, line 70: Hi Volkan, can we make it a reference md file and in that file add some text (e.g. something like "You can find references to some other related information here") and the refer to the pdf from there?
We might have more references in the future of course.

MFransen69 (changes requested):

PR #1481, opened 2023-12-27
Changed Autostart to startmode

HaseenaSainul:


sebaszm:
The field is still called startMode in IController. Perhaps it would be better if it's aligned to "startup" (or "startmode") everywhere?

pwielders:
@HaseenaSainul can we aignthe field to "startmode" ?

HaseenaSainul:
> @HaseenaSainul can we aignthe field to "startmode" ?

sure will do that

Source/plugins/IController.h
```diff
@@ -215,7 +215,6 @@ namespace Controller {
                     RESUMED /* @text resumed */
                 };
 
-                bool AutoStart /* @brief Determines if the plugin is to be started automatically along with the framework */;
```
pwielders, line 218: @HaseenaSainul / @sebaszm agree it was not partof the old-code and it is not part of the PR we are resolving here but to align with the disucssion that @HaseenaSainul and I had, it would be good to also (netxt to the "startmode" have the "resumed" flag inhere as these two values together define the default startup behavior of a Service. So maybe good to add this falg (boolean) here as well ?

pwielders (changes requested):
@HaseenaSainul thank you for also updating the documentation!

Source/plugins/IController.h
HaseenaSainul: done

pwielders (dismissed):

sebaszm:
"resume" or "resumed"? All others are "resume**d**" :)

HaseenaSainul:
> "resume" or "resumed"? All others are "resume**d**" :)

sorry, typo mistake. updated

sebaszm (approved):

PR #1507, opened 2024-01-19
do not invoke Terminate on notice of closed connection

bharath-raj-infosys:
do not invoke Terminate on notice of closed
 connection

### DIFF
--- a/Source/com/Communicator.h
+++ b/Source/com/Communicator.h
@@ -1104,8 +1104,6 @@ namespace RPC {
                         observer++;
                     }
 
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
As the comment tries to explain ;-), it is for non-happy day scenarios. If the other side does not "properly" close, it will be forcefully terminated by Thunder. So this is safety for cleaning up the system properly if the child process does not close  its own but the connection is lost. 
What is the use case of removing this safety net?

PR #1509, opened 2024-01-23
Awc contenerization

shelmy-shabu-infosys:
In Awc contenerization branch, 3 patches
1. 15869-awc proxy container
2. 21500-added notifyserviceready to awclistener
3. 33819-LISA not starting up properly


mhughesacn:
Hi Shelmy,
1. Most new code files should have a header.  I would expect Metrological Apache or LGI Apache (if you are working for LGI).
2. Some cmake files have an LGI Copyright but no actual license header. I would expect an Apache header.  (Look at the other files in the repo.)
3. The code matches are on autogenerated files: api.c, api.h.  These files do not need a header but where is the source file from which they were generated?  (Did this come up before in Thunder - please remind me if you know.)

Thank you.

Source/extensions/processcontainers/common/BaseRefCount.h
```diff
@@ -19,6 +19,10 @@
 
 #pragma once
 
+#include <cstdint>
+
+#include "core/Sync.h"
```
pwielders: Do not include single header files from a module but always the whole module, which always has the name `<modulename>/<modulename>.h` which in this case would be "core/core.h".
I think the `<cstdint>` is already included here: https://github.com/rdkcentral/Thunder/blob/master/Source/core/Portability.h#L362 which is part of the core/core.h (and indicrectly through "core/Sync.h") so I do not think it is needed and thus can be removed.

pwielders (changes requested):
Did not review the implementations/* as this is LGI contributed code and behind an interface from the perspective of Thunder. Assume that this code is reviewed within the LGI organisation.
Thunder related code files have been reviewed. Please see remarks

Source/extensions/processcontainers/common/BaseRefCount.h
shelmy-shabu-infosys: As you mentioned, I tried to remove "cstdint" and changed to "core/core.h" but it was throwing error. So I removed "cstdint" and didn't change "core/Sync.h", then it was fine. Is it possible to do like this... if it is fine, can I push the changes to the code..

shelmy-shabu-infosys: I removed "cstdint" from BaseRefCount.h

mhughesacn:
Hi Shelmy (& Pierre),
I understand that files under "implementations" are distinct, but from a repo point of view, ownership & license must be clear.  Please can you do it this way:

- Any new files that you add here have your header.

- Any modified files have an indication that you changed them - suggest a two liner on top:
Modifications are Copyright 2024 Metrological
Licensed under the Apache License, Version 2.0
(or the appropriate year and company name).  Note that the Apache license has a condition that changes be noted.

- Append a catch-all to top level NOTICE:
Files under the implementations directory are:
Copyright Liberty Global B.V.
Licensed under the Apache License, Version 2.0
unless otherwise noted.

Thank you.

shelmy-shabu-infosys:
> Hi Shelmy (& Pierre), I understand that files under "implementations" are distinct, but from a repo point of view, ownership & license must be clear. Please can you do it this way:
> 
> * Any new files that you add here have your header.
> * Any modified files have an indication that you changed them - suggest a two liner on top:
>   Modifications are Copyright 2024 Metrological
>   Licensed under the Apache License, Version 2.0
>   (or the appropriate year and company name).  Note that the Apache license has a condition that changes be noted.
> * Append a catch-all to top level NOTICE:
>   Files under the implementations directory are:
>   Copyright Liberty Global B.V.
>   Licensed under the Apache License, Version 2.0
>   unless otherwise noted.
> 
> Thank you.

Hi @mhughesacn,
I have added " Modifications are Copyright 2024 Liberty Global B.V.
Licensed under the Apache License, Version 2.0 "  these lines in the NOTICE and added a modification header in the BaseRefCount.h
While pushing those changes copyright and Blackduck scan failed.
Could you please review the changes..

Thank you

mhughesacn:
Thanks Shelmy:
Please use the wording I suggested for NOTICE, rather than just the two lines, so that it covers any files without headers.  (I am assuming that the statement is actually correct.)

```
Files under the implementations directory are:
Copyright Liberty Global B.V.
Licensed under the Apache License, Version 2.0
unless otherwise noted.
```

Then I can clear off the checks.

Thank you.

shelmy-shabu-infosys:
> Thanks Shelmy: Please use the wording I suggested for NOTICE, rather than just the two lines, so that it covers any files without headers. (I am assuming that the statement is actually correct.)
> 
> ```
> Files under the implementations directory are:
> Copyright Liberty Global B.V.
> Licensed under the Apache License, Version 2.0
> unless otherwise noted.
> ```
> 
> Then I can clear off the checks.
> 
> Thank you.

Hi @mhughesacn,

I added those four lines as you given, then also copyright and blackduck scan got failed.

Thank you

Source/extensions/processcontainers/common/BaseRefCount.h
pwielders: I think the error you got was a missing MODULE_NAME which is intentional as code (for the non-functional streams TRACING/LOGGING/WARNING REPORTING) need a source and the error complains it was not defined. What you could have done was #define MODULE_NAME Containers and than #include <core/core.h> by including the "core/Sync.h" the MODULE is now identified as "Core", which is technically incorrect as this is the Container module :-)   So could cou try again to include Core, but before including it set the right MODULE_NAME ? See: https://github.com/rdkcentral/Thunder/blob/master/Source/extensions/processcontainers/Module.h
Just learned that the Module file for this artifact already exists, I suggers including this Module file like: #include "../Module.h" i.s.o. "core/Sync.h"

shelmy-shabu-infosys: Hi @pwielders ,
I have added "../Module.h" i.s.o "core/Sync.h" in BaseRefCount.h

pwielders (dismissed):

pwielders:
@shelmy-shabu-infosys for me it is fine now, however it will not let me merge as the automatic build has not been triggered. Could you merge latest master into this PR. That will trigger the automatic build again and hopefully I van merge it than :-)

shelmy-shabu-infosys:
Hi @pwielders 
I have rebased the latest changes into this pull request, can you please review it..

Thank you..

pwielders (approved):

PR #1511, opened 2024-01-31
Add move constructor/operator for the class which support copy constructor/operator

HaseenaSainul:


Source/WPEFramework/Config.h
```diff
@@ -488,7 +593,7 @@ namespace PluginHost {
 
         public:
             InputInfo(const InputInfo&) = delete;
-            InputInfo& operator= (const InputInfo&) = delete;
+            InputInfo& operator=(const InputInfo&) = delete;
```
pwielders, line 601: If we delete the assignment operator for const <Object>&) we should also specifically delete the assignment operator  with the move parameter(<Object>&&).

Source/WPEFramework/Controller.h
```diff
@@ -202,7 +219,9 @@ namespace Plugin {
         };
 
     private:
+        Controller(Controller&&);
```
pwielders: This is, I guess, from the old days, when we did not allow std++11 since the delete keyword did not exist before std++11 we could not use it there. Since we now say that the minimum requirement for Thunder is std++11, we changed the rules and now copy/move constructurs/oassignment operators we do no want, we make them public with the = delete statement. We do not actively update all code to apply these rules, but if we touch the code, we do update it. So would expect these constructors to be moved to public with the = delete; behind it..

Source/WPEFramework/PostMortem.h
```diff
@@ -129,7 +152,7 @@ namespace PluginHost {
     public:
         PostMortemData(PostMortemData&&) = delete;
         PostMortemData(const PostMortemData&) = delete;
-        PostMortemData& operator= (const PostMortemData&) = delete;
+        PostMortemData& operator=(const PostMortemData&) = delete;
```
pwielders, line 156: I guess missing: 
`PostMortemData& operator=(PostMortemData&&) = delete;`

Source/WPEFramework/Controller.h
HaseenaSainul: done

Source/WPEFramework/Config.h
HaseenaSainul: As we discussed need to cross check

Source/WPEFramework/PostMortem.h
HaseenaSainul: As we discussed need to cross check

pwielders: Discussed it with Marcel and you're updates are correct. The idea is that indeed we do define/declare both the MoveConstructor *and* the MoveAssignment (as I expected) but we do not actively change *all* move contructors/move assignement operators, unless we touch the class. 
I misinterpreted theis change and "thought" you touched the class (whitespace changes) but that was probably due to some automatic formatting tool in your environment, So you did not "really" touch the calls and thus the MoveOperator was not mandatory to add...
So, no need to change this, consider the review remark as not made but if you touh a class alwasy add the Move Constructor and the Move Assignment, but I think you did that everywhere :-)

pwielders:
@HaseenaSainul sorry for the delay, but since we are now fully into R5, good to be merged but I guess we first need to fix the merge conflicts. Can you have a look at it and merge it with the latest master so we can merge it into master.

HaseenaSainul:
> @HaseenaSainul sorry for the delay, but since we are now fully into R5, good to be merged but I guess we first need to fix the merge conflicts. Can you have a look at it and merge it with the latest master so we can merge it into master.

@pwielders, I have rebased this PR with latest master

Source/WPEFramework/PostMortem.h
HaseenaSainul: done

Source/WPEFramework/Config.h
HaseenaSainul: done

pwielders (approved):

PR #1512, opened 2024-01-31
Services: update return value of AddRef

HaseenaSainul:


Source/core/Services.h
```diff
@@ -74,7 +74,7 @@ namespace Core {
         uint32_t AddRef() const
         {
             Core::InterlockedIncrement(_instanceCount);
-            return (Core::ERROR_COMPOSIT_OBJECT);
+            return (Core::ERROR_NONE);
```
pwielders: Here we are doing the reference count not for delete-ing. See the associated Release() but to enforce a warning if the Addref/Release is not balanced. The object here is a *real* composit, so the return of Core::ERROR_COMPOSIT_OBJECT is correct ans should stay. But the issue is now the Release. As in ThunderNanoServices you suggested it to also have the Release than return (Core::ERROR_COMPOSIT_OBJECT) we should also do it for these two AddRefs. So my request was to align the Release() here for ServiceAdministrator<> and SinkType<> to return ( Core::ERROR_COMPOSIT_OBJECT) i.s.o. Core::ERROR_NONE, to align with the way we do it now in ThunderNanoServices.
The fact that, in Release(), the check for no more reference *than* delete this, is missing makes it a COMPOSIT_OBJECT. Dleteion of theis object is than delegated to another object..

pwielders (approved):

Source/core/Services.h
HaseenaSainul: Updated

PR #1516, opened 2024-02-02
RDKDEV-941/ SGMM393-547: RDKServices: RDK service APIs responds to requests with invalid security token.

DineshkumarJP:
Reason for change: Including input security token value fully to validate signature.

Risks: Low

Signed-off-by: Dineshkumar P <dineshkumar.p@ltts.com>


pwielders (approved):
Awaiting for which release this is intended. It is already on master so good to go for R5!

pwielders:
Just got feedback this has to go into R4.2

PR #1517, opened 2024-02-02
Error update for JSONRPC requests 

HaseenaSainul:


Source/plugins/JSONRPC.h
```diff
@@ -437,10 +437,10 @@ namespace PluginHost {
             else if (realMethod == _T("exists")) {
                 result = Core::ERROR_NONE;
                 if (handler->Exists(parameters) == Core::ERROR_NONE) {
-                    response = _T("1");
```
pwielders, line 440: This will deviate from what they are expecting in R2 I think, So I assume this is now a breaking interface with R2 (and R4.2)? Right, if so we need to keep it as it was and indeed for R5 we can report a breaking interface change here, if we would like to break it. (Maybe something for the CCB)

pwielders (commented):
@HaseenaSainul I see you added the new states we introduced for master (R5) to more specifically designate JSONRPC protocol errors ny not using already exsiting error codes which is good. This is an internal enhancement.
Did you check that externally (what you see on the interface, the for example Unknown method (-32601) still return the JSONRPC specified error code ?
We need to make sure that externally these numbers stay the same..

Source/plugins/JSONRPC.h
HaseenaSainul: This change is added to make sure it sync with R4.2 version : https://github.com/rdkcentral/Thunder/blob/R4_2/Source/plugins/JSONRPC.h#L352-L356.

HaseenaSainul:
> @HaseenaSainul I see you added the new states we introduced for master (R5) to more specifically designate JSONRPC protocol errors ny not using already exsiting error codes which is good. This is an internal enhancement. Did you check that externally (what you see on the interface, the for example Unknown method (-32601) still return the JSONRPC specified error code ? We need to make sure that externally these numbers stay the same..

Yes, I have cross checked error case using displayInfo, deviceInfo and networkControl plugins with R4.2 as well as some other error like, sending wrong version, wrong callsign, wrong method. All are return same as R4.2. 
If I send wrong method name, it return Unknown method with exact error code in R4.2
curl --data-binary '{"jsonrpc": "4.2", "id": 1234567890, "method": "NetworkControl.1.dns1"}' -H 'content-type:text/plain;' http://192.168.1.5/jsonrpc
{"jsonrpc":"2.0","id":1234567890,"error":{"code":-32601,"message":"Unknown method."}}

pwielders (approved):

PR #1522, opened 2024-02-12
Development/cyclicbuffer

msieben:


Source/core/CyclicBuffer.cpp
```diff
@@ -178,7 +218,11 @@ namespace Core {
     void CyclicBuffer::AdminLock()
     {
 #ifdef __POSIX__
+#ifndef __DEBUG__
         pthread_mutex_lock(&(_administration->_mutex));
+#else
+        ASSERT(!pthread_mutex_lock(&(_administration->_mutex)));
```
pwielders: Lets not execute essential code within ASSERT statements. Just pick up the return value (always) and ASSERT on the result.. There are macros, to avoid WARNINGS on none used variables (if the ASSERT implementation is empty) 
`int result = pthread_mutex_lock(&(_administration->_mutex));
ASSERT(result == 0);
DEBUG_VARIABLE(result);`
should do the trick..
The reasoning behind this is that the scope (under what conditions should ASSERTS fire) is not set in stone to be "only" during DEBUG. Maybe RELEASE also allows ASSERTS to assert, so an ASSERT should not be associated with the __DEBUG__ preprocessor statements.

Source/core/CyclicBuffer.cpp
```diff
@@ -243,7 +287,11 @@ namespace Core {
     void CyclicBuffer::AdminUnlock()
     {
 #ifdef __POSIX__
+#ifndef __DEBUG__
         pthread_mutex_unlock(&(_administration->_mutex));
+#else
```
pwielders: See above, ASSERTS might be useful/configured in RELEASE as well. Just allways ASSERT if you think it is usefull regardless of the BUILD type..

pwielders (changes requested):
Overall explicit definition of the Posix Mutex sound like a sensible step.

msieben:
Done. Included similar changes to the condition variable.

msieben:
Integrated proposal and applied it is also to the condition variable.

Source/core/CyclicBuffer.cpp
```diff
@@ -120,8 +146,34 @@ namespace Core {
...
+
+                // default values
+                ret = pthread_mutexattr_init(&mutex_attr);
+                ASSERT(ret == 0); DEBUG_VARIABLE(ret);
+
+                // enable access for threads, also in different processes
+                ret = pthread_mutexattr_setpshared(&mutex_attr, PTHREAD_PROCESS_SHARED);
+                ASSERT(ret == 0); DEBUG_VARIABLE(ret);
+
+#ifndef __DEBUG__
```
pwielders: Reading up on the documentation, this flag triggers that a recursive lock (lock already taken by this thread) will return an error. In the context of the CyclicBuffer, I think the lock should indeed *not* be recursive so returning an error would be good. But why is this only applicable if we build this in non debug builds? Why would we not check this in Debug builds ?
And why is the functional code now in the ASSERT?

pwielders (commented):
Seems it is partially corrected ? Still some functional code in the ASSERT?

Source/core/CyclicBuffer.cpp
msieben: PTHREAD_MUTEX_ERRORCHECK is more expensive, and, as such, not the preferred type. PTHREAD_MUTEX_NORMAL usually suffices and is the default. Assert's are part of the debug(ging) process and additional error checking is done in this context.

msieben:
I have no objections separating the code from the assert. Can you answer the question: is error checking NOP in debug if asserts are NOP in debug? I had chosen for 'yes'.

Source/core/CyclicBuffer.cpp
```diff
@@ -56,8 +56,34 @@ namespace Core {
...
```
MFransen69: Shoudn't this be `#ifdef ___DEBUG___` ? (and as Pierre mentioned, see his arguments, not execute code in an ASSERT)

MFransen69 (changes requested):

Source/core/CyclicBuffer.cpp
msieben: Good catch. Indeed #ifdef and not #ifndef.

msieben: In retrospect, if the remark is about the fact that the type influences the set of errors returned from locking and unlocking, and, this assert may hide it at those code points, that is correct. Fair point.

pwielders (approved):

MFransen69 (approved):

PR #1526, opened 2024-02-22
[core/cyclicbuffer] : Cherry-picked from '907031e6b3d4055be0859fa8934…ea404ca2b42f9' 'development/cylicbuffer'.

msieben:
Triggers false positives with Thread use as multiple Worker execution might be done from the same underlying POSIX pthread.

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -572,9 +572,6 @@ namespace Core {
         uint32_t result = Core::ERROR_TIMEDOUT;
         uint32_t timeLeft = waitTime;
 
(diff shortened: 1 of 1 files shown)


pwielders (commented):
Why do we want to remove defensive coding? I guess if someone tries to go with the API outside of the required specs it is better to ASSERT than have him searching for bugs that are not there due to him incorrectly using the interface...

msieben:
This was discussed with MFransen69.

msieben:
@pwielders I am closing the PR. Be informed that neither the API (header, declarations) nor documentation do hint on single (Thunder type) Thread use per process as suggestively is intended. I recommend updating the documentation to make clear these ASSERTs are meant to be as part of that restriction.

PR #1527, opened 2024-02-22
[core/cyclicbuffer] : Cherry-picked from '723fc1139cfe6294235777a8506……e0976ad8037a0' 'development/cyclicbuffer'.

msieben:
Reading zero elements does not end the loop, but is also does not require any actual reading (I/O).

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -353,68 +353,68 @@ namespace Core {
         uint32_t head = 0;
         uint32_t offset = 0;
 
-        if (length) {
-            while (!foundData) {
-                oldTail = _administration->_tail;
-                head = _administration->_head;
-                offset = oldTail & _administration->_tailIndexMask;
+        ASSERT(length > 0);
 
-                result = Used(head, offset);
-                if (result == 0) {
-                    //no data, no need in trying reading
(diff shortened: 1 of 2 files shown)


Source/core/CyclicBuffer.cpp
```diff
@@ -353,66 +353,68 @@ namespace Core {
...
-            oldTail = _administration->_tail;
-            head = _administration->_head;
-            offset = oldTail & _administration->_tailIndexMask;
-
-            result = Used(head, offset);
-            if (result == 0) {
-                //no data, no need in trying reading
-                break;
-            }
+        if (length) {
```
MFransen69: Think this is the only actual change, correct? (github diff makes that hard to see)
We always explicitly compare with something in our code (even for a bool :) ) so I guess the above could be if(length > 0)?
And I guess this was added for the case somebody would call this with length 0?

MFransen69 (changes requested):

Source/core/CyclicBuffer.cpp
msieben: The indentation indeed hides the actual change.

MFransen69 (dismissed):

MFransen69 (approved):

PR #1528, opened 2024-02-23
Update test_cyclicbuffer.cpp

MFransen69:


MFransen69:
AS this might not be the only one, we will check more thoroughly

PR #1533, opened 2024-02-27
[LIBRARIES] Refernce counted Library COMRPC objects.

pwielders:
If COMRPC objects are created through the C interface, the object itself should hold on to a reference of the library it is created from this way, the library will not be unloaded as long as there are still COMRPC object living which reside in this SO. This PR makes sure that any instance instantiated through Core::ServiceType<COMRPCOBJECTIMPL> will also grab a refenrence onto the SO it is located in. On deletion (Last reference release of the COMRPC object) this reference on the SO will be dropped as well!

### DIFF
--- a/Source/core/Library.cpp
+++ b/Source/core/Library.cpp
@@ -51,33 +51,43 @@ namespace Core {
 
             // Looks like we need to add a ref count by opening it..
             handle = ::LoadLibrary(filename);
+            if (handle != nullptr) {
+                // Seems we have an dynamic library opened..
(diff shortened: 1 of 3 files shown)


MFransen69 (dismissed):

sramani-metro (commented):
Looks good to me.

MFransen69 (dismissed):

sramani-metro (approved):

MFransen69 (approved):

PR #1536, opened 2024-03-04
[Macos] First set of changes to support macos build

sramani-metro:
These changes may be common to all platforms


Source/core/NodeId.h
```diff
@@ -83,35 +88,56 @@ namespace Core {
...
             struct sockaddr_ll RawSocket;
+#endif 
 #endif
 #ifdef __CORE_BLUETOOTH_SUPPORT__
             struct sockaddr_hci BTSocket;
             struct bluetooth_extended L2Socket;
 #endif
 
         public:
+            address_family_t familyType() const
```
pwielders: To be inline with all the other methods, use Pascalcase, so FamilyType(). Make the member variable than familyType :-)

pwielders (changes requested):
Looks good but lets keep public method names PascalCase (FamilyType) otherwise we are sending a mixed signal about our casing.
Did you check (cause the PR is to big to really validate) the NetworkControlPlugin using dhcp if it still works? This plpugin relies heavily on the IPFrame stuff, so if it is tested you still get an IP address on the Pi through dhcp of the NewtorkControl) I think it is safe to merge... (after the PascalCase update offcourse ;-))

sramani-metro:
One notable side effect of this change in IPFrame.h is that windows build will start failing, because the definitions of struct ip  and other related data structures are unavailable in windows. We have been using the Linux specific structs namely, struct iphdr and have defined the same in [ThunderOnWindows/](https://github.com/WebPlatformForEmbedded/ThunderOnWindows/blob/master/include/net/ipheaders.h). What i have done now is switch to Posix compliant structs that are available in both Macos and Linux (struct ip and others). This allows us to use Posix Compliant structs and in future if Windows decides to add support for net headers, the Posix compliant ones would get more preference than the Linux ones. 

Other than this, i have changed the Pascal case for the method as requested in the comment and have tested this in a RPI to verify DHCP IP.

pwielders (dismissed):

pwielders (approved):

PR #1540, opened 2024-03-06
[DOC] Singletons section

pczekaj99:


docs/utils/singletons.md
```diff
@@ -0,0 +1,43 @@
+## Using and disposing singletons
```
MFransen69: Think it would be good to (very briefly) also describe the use of singletons in Thunder (what they are) and as section on the code in Core  we have to help write singletons (e.g. SingletonType, the role of Core::Singleton::Dispose() etc.) and perhaps some small code example. If you have question on this I think Sebastion will be able to provide some details)

MFransen69 (changes requested):

MFransen69 (approved):

PR #1546, opened 2024-03-13
[Config] By default no traces enabled

sebaszm:
string.split() has very peculiar behaviour... 
this ended up with module:"" if ENABLE_TRACING_MODULES was empty


VeithMetro (approved):

VeithMetro:
@bramoosterhuis and we were able to notice `module:""` because it caused traces from all modules to be enabled, flooding the log 😄

bramoosterhuis (approved):

PR #1548, opened 2024-03-14
[Messaging] Making both Metadata and Data message buffers size configurable

VeithMetro:


Source/messaging/MessageUnit.cpp
```diff
@@ -275,16 +275,17 @@ namespace WPEFramework {
             if (_settings.IsDirect() == true) {
                 _direct.Output(messageInfo, message);
             } else if (_dispatcher != nullptr) {
-                uint8_t serializationBuffer[DataSize];
-                uint16_t length = 0;
+                uint32_t dataSize = _settings.DataSize();
+                uint8_t* serializationBuffer = static_cast<uint8_t*>(::malloc(dataSize));
```
sebaszm: consider ALLOCA here instead of malloc

sebaszm (changes requested):

Source/messaging/MessageUnit.cpp
VeithMetro: Good idea, I'll change it to a static array with max size to both avoid mallocing this many times and to keep it consistent with Metadata like we discussed

sebaszm (approved):

PR #1555, opened 2024-04-01
Check  additional wpeframework Error code through Existing SetError() function instead of calling separately

tabbas651:


karuna2git (approved):

pwielders:
Agree that the move is a good move (it was already done on Matser) But is this really resolving an issue? and what is the issue? This would require a full QA cycle to add this to 4.4 is that worth the effort? Need t check if we have capacity for it to QA

tabbas651:
> Agree that the move is a good move (it was already done on Matser) But is this really resolving an issue? and what is the issue? This would require a full QA cycle to add this to 4.4 is that worth the effort? Need t check if we have capacity for it to QA

Agreed, While implementing these changes may not directly resolve any specific issues, it would contribute to enhancing code consistency, improving maintainability, and reducing redundancy which are beneficial in the long run.
If feasible, I recommend including this updates in the upcoming release.

PR #1557, opened 2024-04-03
Register/Unregister: ensure sink status before registering and unregistering

HaseenaSainul:


Source/WPEFramework/Controller.cpp
```diff
@@ -822,13 +822,18 @@ namespace Plugin {
...
-        ASSERT(std::find(_lifeTimeObservers.begin(), _lifeTimeObservers.end(), notification) == _lifeTimeObservers.end());
+        LifeTimeNotifiers::iterator index(std::find(_lifeTimeObservers.begin(), _lifeTimeObservers.end(), notification));
+        ASSERT(index == _lifeTimeObservers.end());
 
-        _lifeTimeObservers.push_back(notification);
-        notification->AddRef();
+        if (index == _lifeTimeObservers.end()) {
+            _lifeTimeObservers.push_back(notification);
+            notification->AddRef();
+        }
```
pwielders, line 838: In the return value, reflect that the notification sink was not really added. In debug it will ASSERT, in Release it will at least report it..

Source/WPEFramework/Controller.cpp
```diff
@@ -837,6 +842,8 @@ namespace Plugin {
 
     Core::hresult Controller::Unregister(Exchange::Controller::ILifeTime::INotification* notification)
     {
+        ASSERT(notification != nullptr);
+
```
pwielders, line 848: Also here lets report that the Sink was not unregistered as it did not exists.

Source/plugins/Types.h
```diff
@@ -59,44 +59,50 @@ namespace PluginHost {
...
-                _adminLock.Lock();
-                if (_state == state::LOADED) {
-
-                    INTERFACE* entry = _designated->QueryInterface<INTERFACE>();
-                    _designated->Release();
-                    _designated = entry;
-                    _state = state::RUNNING;
-                    if (entry != nullptr) {
-                        _parent.Activated(entry);
+                ASSERT(controller != nullptr);
```
pwielders, line 62: Here I think the addition of the ASSERT should be sufficient. Even in Release, let the application segfault. We are now leaving (igf this is called with a controller that is a nullptr, the object in a undefined state that might lead to unexpected crashes later on. I prefer the crash to happen now. So agree on the addition of the ASSERT but lets not try to work around it. Let the application crash. It is up to the callee to make sure the controller parameter is *not* null.

Source/plugins/Types.h
```diff
@@ -59,44 +59,50 @@ namespace PluginHost {
...
-                } else {
-                    _state = state::RUNNING;
+                    _adminLock.Unlock();
                 }
-                _adminLock.Unlock();
             }
             void Unregister(IShell* controller)
             {
-                _adminLock.Lock();
+                ASSERT(controller != nullptr);
```
pwielders, line 87: Same as above. it is a programming error if we pass in a nullptr for controller. Only ASSERT. In Release it will crash anyway, whihc is good as a programming mistake is made...

pwielders (changes requested):
I hope this is resolving a ticket to be implemented to make Thuner more robust. If this is to resolve an issue found anyhere in the software stack, we need to resolve that part. If running the software in Debug would still "stop" the system due to the ASSERTS, the issue must be resolved anyway.
On the Types.h, lets not try to work-a-round passing in nullptrs. Do love the ASSERT but in Release still would like the system to crash...The user (callee of the methods) should be aware and code for passing in non-nullptr values.

Source/WPEFramework/Controller.cpp
HaseenaSainul: done

Source/WPEFramework/Controller.cpp
HaseenaSainul: done

Source/plugins/Types.h
HaseenaSainul: done

Source/plugins/Types.h
HaseenaSainul: done

pwielders (approved):

PR #1559, opened 2024-04-04
[core/cyclicbuffer] : masked return value

msieben:


Source/core/CyclicBuffer.cpp
```diff
@@ -566,8 +568,8 @@ namespace Core {
         _administration->_reserved = actualLength;
         _administration->_reservedWritten = 0;
 
-        return actualLength;
-    }
+        return Core::ERROR_NONE;
+   }
```
VeithMetro, line 572: I'm pretty sure you have to return the `actualLength` and not just `Core::ERROR_NONE`, because it would break compatibility in some places, e.g. here: https://github.com/rdkcentral/Thunder/blob/52bce0c0a3147d0179d10e1460bfb4205b3a1b95/Source/messaging/MessageDispatcher.h#L184

VeithMetro (changes requested):

Source/core/CyclicBuffer.cpp
msieben: What would then be appropriate? Error values are non-negative if I am not mistaken:

https://github.com/rdkcentral/Thunder/blob/32bb12ba9794124f508074aed6bf17b9d4adaefb/Source/core/CyclicBuffer.cpp#L556
https://github.com/rdkcentral/Thunder/blob/32bb12ba9794124f508074aed6bf17b9d4adaefb/Source/core/CyclicBuffer.cpp#L549
https://github.com/rdkcentral/Thunder/blob/52bce0c0a3147d0179d10e1460bfb4205b3a1b95/Source/core/Portability.h#L861
https://github.com/rdkcentral/Thunder/blob/52bce0c0a3147d0179d10e1460bfb4205b3a1b95/Source/core/Portability.h#L872

VeithMetro: I'm not a big fan of two things here:

- We have two returns instead of just one
- If we have to return `actualLength`, and we do need to return it if we don't want to break combability or make other changes, then returning any error code in case something goes wrong could be misinterpreted by the one who uses this method

@sebaszm I assume we should just put 0 into `actualLength` if it is not possible to reserve the buffer for whatever reason, and do the rest of the code only if it is possible, so something like this?
```
uint32_t actualLength;

if (!noOtherReservation) {
            // Request not honored
            actualLength = 0;
}
else {
        uint32_t actualLength = length;
        if (length >= _administration->_size) {
            // Maximum write size is _administration->_size-1, to differentiate from empty situation.
            actualLength = _administration->_size - 1;
        }

        AssureFreeSpace(actualLength);
        ASSERT(actualLength <= Free());

        _administration->_reserved = actualLength;
        _administration->_reservedWritten = 0;
}

return actualLength;
```

VeithMetro (changes requested):

msieben:
Closing PR. More information at https://jira.rdkcentral.com/jira/browse/METROL-1000.

PR #1560, opened 2024-04-04
[core/cyclicbuffer] : Cherry pick from 'db117dcbe2ef7d01e80b8aacaff2a……db64c7da8ed'

msieben:
Assert on overflow in 'GetCompleteTail'.

An (unlikely) overflow  can occur if 'size' is set to its maximum value. Then, all bits in the 'mask' become set.


Source/core/CyclicBuffer.h
```diff
@@ -94,6 +96,8 @@ namespace Core {
 
             uint32_t GetCompleteTail(uint32_t offset) const
             {
+                ASSERT(_Parent._administration->_tailIndexMask < std::numeric_limits<uint32_t>::max());
```
VeithMetro: Maybe let's use static cast in order to get a maximum, since then we do not need to include `limits`.
So something like this instead:
```
static_cast<uint32_t>(~0)
```

VeithMetro (changes requested):
Agreed that this ASSERT could be here just to be sure we won't be diving by zero soon after.

Source/core/CyclicBuffer.h
msieben: Agreed.

VeithMetro (approved):

PR #1561, opened 2024-04-08
[JSONRPC] Add IConnectionServer to IShell

sebaszm:


sebaszm:
The changes in Core look scary, but they were trivial in fact, it's the diff that is mixed up. :)

Source/plugins/JSONRPC.h
```diff
@@ -34,11 +34,40 @@ namespace PluginHost {
...
+    private:
+        class Notification : public IShell::IJSONRPCLink::INotification {
+        public:
+            Notification(JSONRPC& parent)
+                : _parent(parent)
+            {
+            }
+            ~Notification() = default;
+
+            Notification(const Notification&) = delete;
```
pwielders, line 50: As this is new code, we should also delete the move constructor and the move assignment operator.

Source/WPEFramework/PluginServer.h
```diff
@@ -2814,8 +2814,6 @@ namespace PluginHost {
             {
                 _notificationLock.Lock();
 
-                ASSERT(std::find(_channelObservers.begin(), _channelObservers.end(), sink) == _channelObservers.end());
```
pwielders: Hopefully this is a temporary thing? Why do we want to allow a sink to be registered more than twice ?

sebaszm: Yes, hopefully.

There is a slight problem, e.g. reload UI page.: there is a new controller.1.register call that will call Subscribe which will Register the channel observer. Obviously in this scenario there is no controller.1.unregister called. The previous connection is closed shortly after, but late enough to trigger this assert. The observer will be released on behalf of the closed channel anyway thus there's no leak. However this is now prone to a dumb programmer. 

The simplest solution would be registering the observer upfront, not doing it lazily until there  is a first Subscribe call, but that smells of a resource waste.

Source/WPEFramework/PluginServer.cpp
```diff
@@ -290,8 +290,21 @@ namespace PluginHost {
...
         }
-        else {
+        else if (id == PluginHost::IShell::ICOMLink::ID) {
+            AddRef();
+            result = static_cast<PluginHost::IShell::ICOMLink*>(this);
+        }
+        else if (id == PluginHost::IShell::IConnectionServer::ID) {
+            _pluginHandling.Lock();
 
+            if (_jsonrpc != nullptr) {
```
pwielders: Do we want to limit this to only available if there is JSONRPC on the Plugin? It is now generic functionality. Might be a good start as the usecase if for JSONRPC but it might be usefull for more ?

Source/WPEFramework/PluginServer.h
```diff
@@ -2800,6 +2803,32 @@ namespace PluginHost {
             {
                 _processAdministrator.Unregister(sink);
             }
+            void Register(IShell::IConnectionServer::INotification* sink)
+            {
+                _notificationLock.Lock();
+
+                _channelObservers.push_back(sink);
+
+                ASSERT(sink != nullptr);
```
pwielders, line 2832: Right now we are not really interested in the Opened() connections but suppose it becomes a usefull use case, we really want to track opened and closed connections, than we need a way to get the initial set of opened connections at the time we register the notification. It think it would be good to iterate within the lock over the open connections and report them now to the newly installed (sink) sink?

Source/plugins/IShell.h
```diff
@@ -59,6 +61,22 @@ namespace PluginHost {
             virtual void* Instantiate(const RPC::Object& object, const uint32_t waitTime, uint32_t& connectionId) = 0;
         };
 
+        struct EXTERNAL IConnectionServer : virtual public Core::IUnknown {
```
pwielders, line 64: Love the generic approach !

Source/WPEFramework/PluginServer.cpp
sebaszm: Totally agree. Additionally you may still be interested in connections, but not necessarily implement provide JSONRPC yourself.

Source/WPEFramework/PluginServer.h
sebaszm: Definitely.

Source/WPEFramework/PluginServer.h
sebaszm: The assert was reinstated.  The comment above does not apply.

Source/plugins/IShell.h
```diff
@@ -255,66 +273,82 @@ namespace PluginHost {
         virtual Core::hresult Hibernate(const uint32_t timeout) = 0;
         virtual reason Reason() const = 0;
 
+        virtual void Register(IConnectionServer::INotification* sink) = 0;
```
pwielders, line 276: Do we still need these public? I can, I assume, QueryInterface<IConnectionServer>() and than Register/Unregister? Or am I missing something here? This would be a convenience method I guess but than I have to implement it on all IShells ?

Source/plugins/IShell.h
```diff
@@ -255,66 +273,82 @@ namespace PluginHost {
...
 
-        // Method to access, in the main space, a COM factory to instantiate objects out-of-process.
-        // This method will return a nullptr if it is NOT in the main process.
-        /* @stubgen:stub */
-        virtual ICOMLink* COMLink() = 0;
-
         inline void Register(RPC::IRemoteConnection::INotification* sink)
         {
-            ICOMLink* handler(COMLink());
+            ASSERT(sink != nullptr);
```
pwielders, line 286: Given the remark above ;-) Maybe we should also move this to a deditcated interface and than QueryInterface<IRemoteConnection>() and than Register ? What do you think? 
Hmmm these are real convenience methods... Cant we do that for the IConnectionServer as well than ?

pwielders (approved):
But left a bit of food-for-thought :-)

sebaszm:
> But left a bit of food-for-thought :-)

will follow up in another pull request!

PR #1577, opened 2024-04-18
Hibernate Abort improvements

adrianM27:
- fixed logging and return in case of late abortion
 - quicker for multi process plugins


pwielders (approved):
Thanks for the update! Looks very good !

PR #1578, opened 2024-04-18
Hibernate Abort improvements

adrianM27:
- fixed logging and return in case of late abortion
 - quicker for multi process plugins

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -707,11 +707,12 @@ namespace PluginHost
 
                 if (result == HIBERNATE_ERROR_NONE) {
                     result = HibernateChildren(parentPID, timeout);
-                    if(result != Core::ERROR_NONE && result != Core::ERROR_ABORTED) {
-                        //wakeup Parent process to revert Hibernation
-                        TRACE(Activity, (_T("Wakeup plugin [%s] process [%u] on Hibernate error [%d]"), Callsign().c_str(), parentPID, result));
-                        WakeupProcess(timeout, parentPID, _administrator.Configuration().HibernateLocator().c_str(), _T(""), &_hibernateStorage);
-                    }
+                }
+
+                if (result != Core::ERROR_NONE && result != Core::ERROR_ABORTED) {
+                    // try to wakeup Parent process to revert Hibernation and recover
+                    TRACE(Activity, (_T("Wakeup plugin [%s] process [%u] on Hibernate error [%d]"), Callsign().c_str(), parentPID, result));
+                    WakeupProcess(timeout, parentPID, _administrator.Configuration().HibernateLocator().c_str(), _T(""), &_hibernateStorage);
                 }
 
                 Lock();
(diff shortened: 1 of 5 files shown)


Source/WPEFramework/PluginServer.cpp
```diff
@@ -840,13 +841,19 @@ namespace PluginHost
 
         if (children.Count() > 0) {
 
+            // make sure to wakeup PIDs in opposite order to hibernation
+            // to quickly go over not hibernated PIDs and abort currently processed
+            std::list<ThreadId> childrenBackward;
```
pwielders: Do we really need to create a copy of the list? can you not children.Reset(false (or ->Count()) ? set the children interator to the end, and than while(children.Previous()) run through the list and WakeUp....?
And if that is not possible, at least use the std::vector<> which is smaller and allocate enough space upfront, using the reserve(children.Count()); This way the vector does not need to grow later..
See: https://baptiste-wicht.com/posts/2012/11/cpp-benchmark-vector-vs-list.html#:~:text=std%3A%3Avector%20is%20insanely,or%20inserting%20in%20the%20front

pwielders (commented):
small improvement maybe possible to avoid copying the list...

pwielders:
Remaining part looks like a good improvement!

pwielders:
Saw you addressed my review remarks for the master, so happy to take this to 4.4. Just need to discuss if this can go in during the next QA cycle!
Or do you want to align this to they way it is done in the master for me it is not  must, it is up to you!

pwielders (approved):

adrianM27:
Thank you Pierr for your review,  indeed I made it for master first to be sure unit tests pass.  I am good to wait with this merge, no rush here, lets follow QA cycle

adrianM27:
Hi Pierr,  since we are a bit in hurry to close some tickets, can we merge this change by end of this week?

pwielders:
@MFransen69 can we add this to the 4.4.2 upcoming release a new tag? Or should we take it up in the next QA cycle? I do think there are some minor risks in there that would opt for a QA cycle. So my sugegstion would be in the next QA cycle?
@adrianM27 we could still merge it (but it will not be part of the 4.4.2 tag), we have no QA tests for Hibernation so we will not test the majority of the patch anyway but some code is in shared functionality (locks that where moved) So I do think that QA on this has benefits. We can always merge it and you could use it b using the commit hash of the new tip of the 4.4 branch. Than it will be tagged with the new QA release if we have more issues we need to fix on 4.4..

adrianM27:
Hi @pwielders / @MFransen69  do we have any potential ETD for this PR?

MFransen69:
> Hi @pwielders / @MFransen69 do we have any potential ETD for this PR?

Hi @adrianM27 , I leave it up to @pwielders to merge this, that could be done now as we have 4.4.2. out of the door. If it is merged it will be part of 4.4.3, when that will be released we will have to discuss with Comcast.

adrianM27:
Hi, finally this PNR is not dependent to other deliverables in Comcast, we are fine to wait and follow your QA process

PR #1579, opened 2024-04-24
[Messaging] Never sending controls as one big chunk but always for a specific module

VeithMetro:
- There is a new method to get all available controls
- Removing `Type` from `MessageDataBufferType` since it is no longer templated
- Changing the size of metadata buffer to 4 kB which now should be plenty (we can probably shrink it even much more)


Source/messaging/MessageClient.cpp
```diff
@@ -138,16 +138,33 @@ namespace Messaging {
...
-    void MessageClient::Controls(Messaging::MessageUnit::Iterator& controls) const
+    void MessageClient::Modules(std::vector<string>& modules) const
+    {
+        _adminLock.Lock();
+
+        for (auto& client : _clients) {
+            client.second.Modules(modules);
+        }
+
+        std::sort(modules.begin(), modules.end());
```
pwielders: Why is the sorting and the filtering not part of the function client.second.Modules() in line 148 ?

Source/messaging/MessageUnit.h
```diff
@@ -786,6 +805,38 @@ namespace WPEFramework {
...
+                            uint16_t bufferSize = metaDataFrame->Response().Length();
+                            const uint8_t* buffer = metaDataFrame->Response().Value();
+
+                            Core::FrameType<0> frame(const_cast<uint8_t*>(buffer), bufferSize, bufferSize);
+                            Core::FrameType<0>::Reader reader(frame, 0);
+
+                            size_t iterator = reader.Number<size_t>();
+
+                            while(iterator > 0) {
+                                modules.push_back(reader.NullTerminatedText());
```
pwielders: Suggest to build some logic here. If the module is already in the vector<string> modules, do not add it, and sort it (put it in the right location. If you sort it you also now quicker if it is already in there etc.. (see also remark above).
This will avoid a large std::vector<> with a lot of duplicate modules that need to be sorted and removed later on..

pwielders (commented):
I thought there was also some header file that could maybe be moved from core to messaging? I do not see it here? Was that not possible at the end?

Source/messaging/MessageUnit.h
sebaszm: Agree with Pierre. I would not even sort it. Just made sure I'm not adding a double. The JS client will have all the time to sort the result if they really require it.

Source/messaging/MessageUnit.cpp
```diff
@@ -63,14 +68,59 @@ namespace WPEFramework {
...
                 uint16_t _offset;
-            } handler (buffer, length);
+            } handler(buffer, length, module);
 
             Core::Messaging::IControl::Iterate(handler);
 
             return (handler.Offset());
         }
+        
+        uint16_t MessageUnit::SerializeModules(uint8_t* buffer, const uint16_t length)
```
sebaszm: The signature is different, can we stick with just "Serialize"? :)

Source/messaging/MessageUnit.cpp
```diff
@@ -63,14 +68,59 @@ namespace WPEFramework {
...
+            private:
+                std::vector<string>& _modules;
+            } handler(modules);
+
+            Core::Messaging::IControl::Iterate(handler);
+
+            Core::FrameType<0> frame(buffer, length, length);
+            Core::FrameType<0>::Writer writer(frame, 0);
+
+            writer.Number<size_t>(modules.size());
```
sebaszm: I don't think there can be size_t modules...

Source/messaging/MessageUnit.h
```diff
@@ -736,25 +742,38 @@ namespace WPEFramework {
...
                 {
                     if (_channel.IsOpen() == true) {
 
                         // We got a connection to the spawned process side, get the list of traces from
                         // there and send our settings from here...
                         Core::ProxyType<MetadataFrame> metaDataFrame(Core::ProxyType<MetadataFrame>::Create());
 
-                        metaDataFrame->Parameters().Set(0, nullptr);
+                        uint8_t buffer[64];
+                        Core::FrameType<0> frame(buffer, 64, 64);
```
sebaszm: sizeof(buffer)

Source/messaging/MessageUnit.h
```diff
@@ -786,6 +805,38 @@ namespace WPEFramework {
                     }
                 }
 
+                void Modules(std::vector<string>& modules) const
+                {
+                    if (_channel.IsOpen() == true) {
+                        Core::ProxyType<MetadataFrame> metaDataFrame(Core::ProxyType<MetadataFrame>::Create());
+
+                        uint8_t buffer[1];
+                        Core::FrameType<0> frame(buffer, 1, 1);
```
sebaszm: sizeof(buffer)

Source/messaging/MessageUnit.h
```diff
@@ -819,16 +870,33 @@ namespace WPEFramework {
...
+                            if (reader.HasData()) {
+                                metadataFrameProtocol protocol = reader.Number<metadataFrameProtocol>();
+
+                                if (protocol == metadataFrameProtocol::UPDATE) {
+                                    Control newSettings;
+                                    newSettings.Deserialize(reader.Data(), reader.Length());
+                                    _parent.Update(newSettings, newSettings.Enabled());
+                                    message->Response().Set(0, nullptr);
+                                }
+                                else if (protocol == metadataFrameProtocol::CONTROLS && reader.HasData()) {
```
sebaszm: I think that's too defensive, assert for hasData would be enough

Source/messaging/MessageUnit.h
```diff
@@ -819,16 +870,33 @@ namespace WPEFramework {
...
-                            // What is coming in, is an update?
-                            if (message->Parameters().Length() > 0) {
-                                Control newSettings;
-                                newSettings.Deserialize(message->Parameters().Value(), message->Parameters().Length());
-                                _parent.Update(newSettings, newSettings.Enabled());
-                                message->Response().Set(0, nullptr);
+                            Core::FrameType<0> frame(const_cast<uint8_t*>(message->Parameters().Value()), message->Parameters().Length(), message->Parameters().Length());
+                            Core::FrameType<0>::Reader reader(frame, 0);
+
+                            if (reader.HasData()) {
```
sebaszm: do we need to check for hasData? This cannot be false unless there's a bug

Source/messaging/MessageUnit.h
```diff
@@ -37,11 +37,17 @@ namespace WPEFramework {
...
         class EXTERNAL MessageUnit : public Core::Messaging::IStore {
         public:
-            static constexpr uint16_t MetadataBufferSize = 8 * 1024;
+            static constexpr uint16_t MetadataBufferSize = 4 * 1024;
             static constexpr uint16_t TempMetadataBufferSize = 128;
             static constexpr uint16_t MaxDataBufferSize = 63 * 1024;
             static constexpr uint16_t TempDataBufferSize = 1024;
 
+            enum metadataFrameProtocol : uint8_t {
+                UPDATE      = 0,
```
sebaszm, line 46: Perhaps we could use the 0 as invalid content marker, as the frame reader will return 0es on error or frame truncation.

Source/messaging/MessageUnit.cpp
```diff
@@ -63,14 +68,59 @@ namespace WPEFramework {
...
+
+            std::vector<string>::iterator it;
```
sebaszm: const_iterator

sebaszm (changes requested):

Source/messaging/MessageClient.cpp
VeithMetro: The idea behind it was to only sort and remove duplicates once, since we need to do this after all clients have been already asked for their modules. This way we could avoid looking through the modules vector in each client.

But I do agree that from the memory perspective it is much more efficient not to add duplicates at all rather than removing them later on, and I suppose we care more about the used memory, so I'll change it and add a logic to Modules() in Client 😄 

> I thought there was also some header file that could maybe be moved from core to messaging? I do not see it here? Was that not possible at the end?

We discussed moving `IPCMessage.h` from core to messaging, but with Sebastian we decided to make this pull request solely about no longer sending controls as a big one chunk and the interface changes (so that we can surely have it ready before R5). After this one is merged, we will look into some cases of unnecessary string copying and also try to move this header file to messaging in another pull request.

Source/messaging/MessageUnit.h
VeithMetro: Like I replied above - I do agree and I will add a logic here, since this way it is much more memory efficient. And I also think we can get rid of sorting altogether - it was only needed to do the erase to remove duplicates. If we want the modules to be sorted in the UI, than we can do that in JavaScript

Source/messaging/MessageUnit.cpp
VeithMetro: Always happy to make changes to confuse we all know who (and probably also us in the future 😃 )

Source/messaging/MessageUnit.cpp
VeithMetro: Agree, let's make it uint8_t and add an ASSERT in Serialize if for some reason there are more than 255 modules (which would not fit into the buffer anyway..)

Source/messaging/MessageUnit.h
VeithMetro: Agree, there must be data if everything was correct, so an ASSERT fits much better

Source/messaging/MessageUnit.h
VeithMetro: True, this was only be false if there is a bug, so ASSERT is better than a check

Source/messaging/MessageUnit.h
VeithMetro: Like we discussed during our morning meeting, we no longer need this additional safety, since we have ASSERTs to check if the frame reader actually has data to read

Source/messaging/MessageUnit.cpp
VeithMetro: Agree, we are only reading from `modules` with this iterator, not writing , so `const_iterator` is better here

Source/messaging/MessageUnit.h
sebaszm: Correct.

sebaszm (approved):

PR #1585, opened 2024-05-04
Cleanup Resources when Socket Class destroyed

karuna2git:
When destructor of SocketClass identifies that the Socket is NOT Closed, it must call Close() to update the state n unregister resource monitors

### DIFF
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -419,7 +419,7 @@ namespace WPEFramework {
             ASSERT((m_Socket == INVALID_SOCKET) || (IsClosed()));
 
             if (m_Socket != INVALID_SOCKET) {
-                DestroySocket(m_Socket);
+                Closed();
             }
 
             ::free(m_SendBuffer);


pwielders (changes requested):
There is a reason why we propagate defensive coding. Defensive coding helps in finding bugs early without goign through a full cycle of testing (readup on the V model of software development, the earlier you find a bug the cheaper it is)
As you can see 3 lines above, we expect the socket to be INVALID at this point ir if it is not Invalid, it should *at least* be Closed! The reason for this is inheritance. This is a base class with pure methods. 
The Destructors in C++ class inheritance go from top to bottom (derived to base). This means that the destructor of the Derived class (the one that implements the pure virtuals of this class) has run *prior* to this destructor. This means that if we end up here and a virtual method would be called it is an erronous situation. That derived class is already destructed and is not expecting any calls anymore. 
If you would call Closed here it might still call any of the virtual methods due to the unregister, see: https://github.com/rdkcentral/Thunder/blob/e53452af823c44ade2ddfaf42e880e8fe1d403e9/Source/core/SocketPort.cpp#L1261
So this is masking and hiding a problem in the user layer and should *not* be added as this potentially creates segmentation faults.
As explained, also with a similar patch in this area, make sure the derived class (the class that implements the pure methods of this class) in its destrcutor calls: Close(core::infinite), to make sure the socket is removed from the Communication thread and it nolonger participates in communication *before* you call this destructor, hence why the ASSERT is here, there is a *good* reason for it.
So we *cannot* take this patch as it causes instability. If this solves an issue in your code, it means you forgot the Close(Core::infinite) in the destructor of the Derived class. Fix it there!

PR #1592, opened 2024-05-08
RDK-48815 : Crash handling

sethulakshmisk:
Reason for change: Improve Thunder Responsiveness [Crash handling in Thunder R4]
Test Procedure: Simulate crash in rdkservices and verify the call stack actvity logs are generated
Risks: NO
Priority: P1

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -28,6 +28,9 @@ Licensed under the Apache License, Version 2.0
 Copyright 2014 Fraunhofer FOKUS
 Licensed under the Apache License, Version 2.0
 
+Copyright 2024 Synamedia Ltd.
+Licensed under the Apache License, Version 2.0
+
 Copyright 2016-2017 TATA ELXSI
 Licensed under the Apache License, Version 2.0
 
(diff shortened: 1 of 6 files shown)


pwielders:
Below my analyses of the code. Please eleborate on what this patch tries to do if my analyses below is *not* correct. If my analyses below is correct, please close the PR as the root-cause should be solved in the RDKService (use grep to find the plugin that changes the signal handlers)

Interesting code.... Trying to understand what is going on here...
If I interpret the code correctly, you are restoring the Signal handlers to the Thunder signal handlers if they where changed each time before you have the worker pool thread execute new code.
From this I assume that there is code executed in RDKServices plugins that change the signal handlers. and this is reverting back that change. Would it not be easier to just remove that code from that plugin?
Now we are paying a CPU Resource usage penalty over and over (on all threads) since there is an issue in an RDK service that misbehaves. 
So I consider this a degradation of Framework responsiveness (we are wasting a lot of CPU cycles with this patch due to a misbehaving plugin) and consider this an invalid PR.

mhughesacn:
Thank you - downvotes cleared

PR #1597, opened 2024-05-13
Increased queueSize for WorkerPool to avoid full queue when over 30 

sborushevsky:
plugins are trying to activate.

### DIFF
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -123,7 +123,7 @@ namespace PluginHost {
             WorkerPoolImplementation& operator=(const WorkerPoolImplementation&) = delete;
 
             WorkerPoolImplementation(const uint32_t stackSize)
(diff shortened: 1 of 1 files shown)


anand-ky (approved):

MFransen69:
We agreed to cherry pick the R5 solution to R4.4

pwielders:
And it is an incorrect patch. The queue size should be proportional to the number of threads so that the system scales in case the number of threads are changed. This feature takes care of throttling the system load. For that it should not be fixed but a proportional part of the number of threads. 
We requested you to set the value to 50 to check if indeed you are overloading/stressing the system with requests/jobs. Just as a test. Learning that Comcast increases the ThreadPool count, we decided to make a proper fix (as it was proven that the system for ComCast was stressing the system) to make the total queu size proportional to the number of threads requested. Fix available in master and as the process states to be cherry-picked on request to other releases.

PR #1598, opened 2024-05-14
Renaming WPEFramework to Thunder

HaseenaSainul:


Source/core/Portability.h
```diff
@@ -684,7 +684,7 @@ typedef DWORD ThreadId;
 #define DEBUG_VARIABLE(x)
 #endif
```
MFransen69, line 686: Let's add for the time being add a "namespace WPEFramework = Thunder;", we are of course done now but this will people some time to gradually transition

MFransen69 (changes requested):

MFransen69 (approved):

PR #1605, opened 2024-05-20
[Core] Fixing each corner case for the CyclicBuffer overflow

VeithMetro:
This issue was first observed by @msieben while working on MessageDispatcher tests (https://jira.rdkcentral.com/jira/browse/METROL-970), and it looks like we have never encountered a situation when the CyclicBuffer was full since MessageControl was always able to PopData() quickly enough..

Besides, I have also updated tests on Mischa's branch to include all corner cases of the CyclicBuffer being overflown (https://github.com/rdkcentral/Thunder/commit/ff3c391037f275f60e7a79acdb8593ba99edb62a), and after the CyclicBuffer changes all of them pass correctly.

Here are the corner cases:

- Part of the offset ends up on the end of CyclicBuffer, and the other half ends up at its beginning
- The whole offset fits at the end of CyclicBuffer, and then the buffer itself ends up at the beginning of CyclicBuffer
- The whole offset fits at the end of CyclicBuffer, as well as a part of the buffer, but the other part ends up at the beginning of CyclicBuffer

### DIFF
--- a/Source/core/CyclicBuffer.cpp
+++ b/Source/core/CyclicBuffer.cpp
@@ -394,25 +394,25 @@ namespace Core {
                             foundData = false;
                         }
                     } else {
-                        uint32_t part1 = 0;
-                        uint32_t part2 = 0;
+                        uint32_t newOffset = 0;
 
                         if (_administration->_size < offset) {
-                            part2 = result - (offset - _administration->_size);
+                            const uint32_t skip = offset - _administration->_size;
+                            newOffset = skip + bufferLength;
+                            ::memcpy(buffer, _realBuffer + skip, bufferLength);
                         } else {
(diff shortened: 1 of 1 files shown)


pwielders:
Looks oke, but must admit, if I need to review it carefully I need to dive deep into it. Hope that you and @sebaszm discussed this in detail. I am relying on his verdict ;-)

VeithMetro:
> Looks oke, but must admit, if I need to review it carefully I need to dive deep into it. Hope that you and @sebaszm discussed this in detail. I am relying on his verdict ;-)

Indeed a deeper dive into it was necessary, but we did discuss it with @sebaszm quite extensively and checked each corner case we could think of.

sebaszm (approved):

PR #1609, opened 2024-05-23
Sync: Trace level for lock set to higher level than 1

HaseenaSainul:


VeithMetro (approved):
No more flooding of the log in debug, hurray 😄

pwielders (approved):

PR #1618, opened 2024-05-30
RDK-41134 Improve Thunder Responsiveness [Master]

sethulakshmisk:
Reason for change: Improve Thunder Responsiveness [Crash handling in Thunder Master]
Test Procedure: Simulate crash in rdkservices and verify the call stack actvity logs are generated
Risks: NO
Priority: P1

### DIFF
--- a/Source/WPEFramework/PluginServer.cpp
+++ b/Source/WPEFramework/PluginServer.cpp
@@ -19,6 +19,7 @@
 
 #include "PluginServer.h"
 #include "Controller.h"
+#include "PluginHost.h"
 
 #ifndef __WINDOWS__
 #include <syslog.h>
@@ -193,6 +194,8 @@ namespace PluginHost {
 
         WARNING_REPORTING_THREAD_SETCALLSIGN_GUARD(callsign.c_str());
     #endif
+      
+	    SetupCrashHandler();
 
     #ifdef __CORE_EXCEPTION_CATCHING__
 
--- a/Source/WPEFramework/PluginServer.h
+++ b/Source/WPEFramework/PluginServer.h
@@ -4393,6 +4393,12 @@ namespace PluginHost {
                 std::list<Core::callstack_info> stackList;
 
                 ::DumpCallStack((ThreadId)index.Current().Id.Value(), stackList);
+                for(const Core::callstack_info& entry : stackList)
+                {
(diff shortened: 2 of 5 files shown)


Source/WPEFramework/PluginServer.cpp
```diff
@@ -193,6 +194,8 @@ namespace PluginHost {
 
         WARNING_REPORTING_THREAD_SETCALLSIGN_GUARD(callsign.c_str());
     #endif
+      
+	    SetupCrashHandler();
```
pwielders, line 198: As already explained in https://github.com/rdkcentral/Thunder/pull/1592 there should be no need to check/install the crash handler each time a job is about to be run. Check the plugin for *not* modifying the signal handlers! 
This is to cover up an issue caused by a plugin. Fix the plugin.
It is also resource usage which should not be necessary. The handlers get installed at the beginning and that should be sufficient.
If that is not sufficient (if that is the issue) do it in the Initialize of the thread which only occures once, just before starting it.

Source/WPEFramework/PluginServer.h
```diff
@@ -4393,6 +4393,12 @@ namespace PluginHost {
                 std::list<Core::callstack_info> stackList;
 
                 ::DumpCallStack((ThreadId)index.Current().Id.Value(), stackList);
+                for(const Core::callstack_info& entry : stackList)
```
pwielders, line 4396: Why print the calls stack it is about to be dropped to disk, see line 4416.

pwielders:
Looked up, as I got in doubt with this second attempt to get this in, if signal handlers are on process lever or on thread level. The first hits I saw was that they where on process level as I thought as well. So this means that there is absolutely no use in reinstalling the signal handlers on a thread over and over again if someone (outside of Thunder) is constantly replacing the handlers. They are global, so the faulty (the ones you are trying to replace) can still trigger! 
You have to find the root-cause, the plugin changing the signal handlers and remove that code from the plugin!
The part where the signal being fired is specifically named is a useful enhancement of the code and we could take that into the product! PluginHost.cpp (325-338)

woutermeek:
@sethulakshmisk did you manage to find the root cause of the problem in any of your plugins? Do you understand what we mean or do you need additional information?

PR #1640, opened 2024-06-12
Thunder Build Error Fixed when ProcessContainer enabled

tabbas651:
* Fixed Thunder R4.4..1 Compilation error when enabled ProcessContiainer
* Correct the UnLock spelling
* Correct the messaging header file
* Updated the ProcessContainers::ProcessContainerization into Trace::Information


Source/extensions/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -56,16 +56,16 @@ namespace ProcessContainers {
             };
 
             Core::File configBundleFile(path + "/Container" + CONFIG_NAME);
-            TRACE(ProcessContainers::ProcessContainerization, (_T("searching %s container at %s, %s"), id.c_str(), configBundleFile.Name().c_str(), configBundleFile.PathName().c_str()));
```
MFransen69, line 59: Why did you change the trace category? Using the ProcessContainers category enables easily filtering on container messages. Might it not compile then let's fix that but not change the category
(after doing the rest of the review I think it might be related to the include below, please have a look there as well)

Source/extensions/processcontainers/implementations/DobbyImplementation/DobbyImplementation.h
```diff
@@ -24,7 +24,7 @@
 #include "processcontainers/common/BaseRefCount.h"
 #include "processcontainers/common/CGroupContainerInfo.h"
 
-#include "Messaging.h"
+#include "messaging/messaging.h"
```
MFransen69, line 27: I think it should include this file https://github.com/rdkcentral/Thunder/blob/b689c5941fc8a4b93b9fb08d9cc76be67c5cddf6/Source/extensions/processcontainers/Messaging.h

so the line is correct as is, it will then also include the ProcessContainers::ProcessContainerization catergory. In the  Messaging.h there is actually an error. Line 24 is incorrect. My proposal would be to remove line 24 there at all and include <messaging/messaging.h> in https://github.com/rdkcentral/Thunder/blob/b689c5941fc8a4b93b9fb08d9cc76be67c5cddf6/Source/extensions/processcontainers/Module.h like we do for core.h

MFransen69 (changes requested):

MFransen69:
We agreed this PR to be closed and the fixes to be incorporated into a fix for the remaining process containers issues

PR #1643, opened 2024-06-14
Support for autowakeup on IPC call to hibernated plugin

adrianM27:
We need a way to detect if someone is trying to call API of hibernated plugin via Thunder and eventually wakeup the plugin to continue API call execution, or return error.

### DIFF
--- a/Source/Thunder/CMakeLists.txt
+++ b/Source/Thunder/CMakeLists.txt
@@ -88,10 +88,15 @@ if (PROCESSCONTAINERS)
 endif(PROCESSCONTAINERS)
 
 if(HIBERNATESUPPORT)
+    option(HIBERNATESUPPORT_AUTOWAKEUP "Autowakeup on IPC call for Hibernate" OFF)
     target_link_libraries(${TARGET} PRIVATE
         ${NAMESPACE}Hibernate::${NAMESPACE}Hibernate)
     target_compile_definitions(${TARGET} PUBLIC
          HIBERNATE_SUPPORT_ENABLED=1)
+    if(HIBERNATESUPPORT_AUTOWAKEUP)
+         target_compile_definitions(${TARGET} PUBLIC
+               HIBERNATE_SUPPORT_AUTOWAKEUP_ENABLED=1)
+     endif()
 endif()
 
 set_target_properties(${TARGET} PROPERTIES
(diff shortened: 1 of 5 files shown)


adrianM27:
Hi,  We need a way to detect if someone is trying to call API of hibernated plugin via Thunder  and eventually wakeup the plugin to continue API call execution, or return error. Changes are subtle and will be great full for deep review.

pwielders:
Hi @adrianM27,

Is there some requirements document, or any other document, for this feature? Could you share it ?

This seems like a very hacky solution that messes up the architecture. I do not think that an IPCChannel should be aware of any hibernation. Also think there is no need to move it to such a low level. I think we are violating the D principle in SOLID. See: https://www.digitalocean.com/community/conceptual-articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle
What I take from the comments and the socket changes is that you want to "WakeUp" an Hibernated plugin in case a JSONRPC request for it comes in? 
I think, if needed, it could be done from the PluginServer (WPEFramework) alone, so only at the highest level where all the Hibernation functionality exists and no need to scatter a hibernation concept trough the whole code base.
The feature, I think, also opens up a whole box of Pandora issues, cause if we wake up how long should we wake up? Or is this a matter of waking up, servicing the JSONRPC request, and going back to Hibernation (I do not think I found that in the code). I guess the concept of hibernation is to free up Memory, so that in case of low resource devices the system could potentially startup/shutdown plugins quicker. By bringing the plugin in hibernation we free up memory to be used by another plugin. However, any JSONRPC call from anyone, may now bring the memory back into play by simply sending a request, which might very easily than lead to an OOM.
Than control of the plugins Hibernation mode is now back in the hands of anyone. There is no longer 1 captain orchestrating the plugins availability and thus making sure the memory usage is under control. Sending any requests, will change the state of a plugin. 
I think this is the beginning of much more issues. This is what was discussed at the beginning of the hibernation feature and why we think it would be better to use ZRam in combination with SWAP. Trust me, the kernel is better suited for this kind of features than what we are trying to do in userspace by programming this.

adrianM27:
Thanks Pierr,  I agree, changes are against current architecture .  I abandon the PR, since we found another way of doing it outside Thunder

PR #1653, opened 2024-06-19
Re-enable message exception

msieben:
Online Windows build test.


msieben:
Made it an actuel PR

msieben:
Request from @Pierre to investigate a restructuring of the MessageException code to avoid warnings in the Windows build system.

msieben:
All checks passed. Made it an actual PR.

Source/core/MessageException.h
```diff
@@ -21,23 +21,42 @@
...
+#include "Serialization.h"
 
 namespace Thunder {
 namespace Core {
 
-    class MessageException : public std::exception {
+    class  MessageException : public std::exception {
     private:
-        MessageException();
+        inline MessageException()
```
pwielders: According to the latest guidelines, we need to declare at least the move and the copy constructor. so please declare them = deleted, also think/hope that the default constructor is not needed, so set it as deleted as well and all public.
Also, since the second parameter in the constructor has a default, make sure that single parameter constructors are made explicit (see example I shared).

pwielders (changes requested):
Conforming to latests coding guidelines if legacy code is touched..

Source/core/MessageException.h
msieben: Now, you mention it. I remember vaguely such comment from one of the discussions, but not the documentation. It will be sorted.

Source/core/MessageException.h
```diff
@@ -21,23 +21,43 @@
...
-        MessageException(const string& message, bool inclSysMsg = false) throw();
-        ~MessageException() throw();
+        MessageException() = delete;
+        MessageException(const MessageException&) = delete;
+        MessageException(MessageException&&) = delete;
+        MessageException& operator=(const MessageException&) = delete;
+        MessageException& operator=(MessageException&&) = delete;
+
+PUSH_WARNING(DISABLE_WARNING_DEPRECATED_USE)
+        inline MessageException(const string& message, bool inclSysMsg) noexcept(true)
```
pwielders: This is a bit tricky, You are changing the interface (breaking it).The original interface could work with only 1 parameter (hence the explicit request on the constructor of this). Now it is mandatory that the constructor must have a boolean as the second parameter, which means that if there is code that is using this (might be in RDKServices) you break it with this change. Suggest to keep the signature of the constructor the same and make the constructor  explicit. Unless there is a good reason why you want to break it (technical issues ?)
P.S. You can argue if the explicit could cause a break as well, since you now need to do an explicit converion. Might also break the code but at least here the technical argument would be that it makes the code more pure, avoids implicit, unwanted conversions and object creations, the developper was not aware of..

pwielders (changes requested):
At least raise a technical reason for breaking the interface and if so we need to enlist ot for the next release so users know the interface is broken. If there is no valid reason, keep the interface as is..

Source/core/MessageException.h
msieben: Just missed the default value in the C&P.

pwielders (approved):

PR #1663, opened 2024-06-21
Dev/macos support

HaseenaSainul:


Source/core/FileObserver.h
pwielders, line 1: I think there is a lot of overlap between the Linux and the Apple implementation (I checked the observer, they are at least equal) suggest to only pre-processor the chunks that are different and not reimplement the whole class for Linux and Mac-OS....

pwielders (changes requested):
@HaseenaSainul can you check the windows build? Seems some changes run through to th e windows build and break it.

Source/core/FileObserver.h
HaseenaSainul: done

Source/core/SocketPort.cpp
```diff
@@ -238,7 +239,11 @@ namespace Thunder {
                         interfaceId = info->ipi_ifindex;
                         break;
                     }
+#ifdef __APPLE__
```
pwielders: To avoid a lot of single line preprocessor descriminitors, look at the top of this file and define the actual type for each platform, so for example at the beginning of this file do, for each platform:
#define __PACKAGE_TYPE__ IPV6_RECVPKTINFO
and
#define __PACKAGE_TYPE__ IPV6_PKTINFO
and than here, use the __PACKAGE_TYPE__ it will keep the code easier to read..

Source/core/SocketPort.cpp
```diff
@@ -770,7 +775,11 @@ namespace Thunder {
             }
 #endif
 
+#ifdef __APPLE__ 
+            if ((l_Result = ::socket(localNode.Type(), SocketMode(), localNode.Extension())) == INVALID_SOCKET) {
```
pwielders: If on apple SOCK_CLOEXEC does not exist, just define it in portability.h as #define SOCK_CLOEXEC  0 than check all code where for APPLE you where working around it and I guess (at least here) there i no need anymore to work around it.
If possible try to avoid #ifdef <PLATFORM> if possible as it makes code harder to read. If it must, you should but if it is not needed by just defining a simple value not available on the platform, just define it ;-) 
Or looking just a few more lines down, could it be that #define SOCK_CLOEXEC  O_CLOEXEC , only in the SocketPort.cpp file ? And if you do, do you than still need the setting of this flag a bit lower ?

Source/core/Thread.cpp
```diff
@@ -141,9 +151,25 @@ POP_WARNING()
 #endif
 
 #ifdef __POSIX__
-        void* Thread::StartThread(Thread* cClassPointer)
+    void* Thread::StartThread(Thread* cClassPointer)
```
pwielders, line 148: Would it not be an option to move the way Linux sets it as well here, just move the Linux here and you do not need a differentiation anymore between Linux and apple.. So also set the Process name for Linux from its own context..
Try to avoid conditional code blocks if possible without sacrificing functionality off-course.

pwielders (changes requested):

Source/core/CMakeLists.txt
```diff
@@ -244,6 +244,10 @@ if (NOT BUILD_SHARED_LIBS)
     )
 endif()
 
+if (APPLE)
+    add_definitions(-D__APPLE_USE_RFC_3542)
```
bramoosterhuis: I think you want to use [```target_compile_definitions```](https://cmake.org/cmake/help/v3.15/command/target_compile_definitions.html) and mark it PRIVATE if it should only be used by the core cpp files, otherwise mark it PUBLIC and so everyone sets this definition automatically that links to core.

CMakeLists.txt
```diff
@@ -132,7 +132,7 @@ if(APPLE)
     set(CMAKE_MACOSX_RPATH ON)
     set(CMAKE_SKIP_BUILD_RPATH FALSE)
     set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_LIBDIR}")
+    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
bramoosterhuis: I read [here](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#always-full-rpath) that if thunder uses the full rpath we need to add a check. I'm not that familiar with linking on MacOS,but maybe it wise to add it if Kitware is advising it.

```
# the RPATH to be used when installing, but only if it's not a system directory
list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
if("${isSystemDir}" STREQUAL "-1")
    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
endif("${isSystemDir}" STREQUAL "-1")
```

Source/core/NodeId.h
```diff
@@ -107,7 +107,7 @@ namespace Core {
 #ifdef __WINDOWS__
             address_family_t Family;
 #elif defined(__APPLE__)
-            struct __sockaddr_header saddr_hdr;
+            struct sockaddr saddr_hdr;
```
bramoosterhuis, line 106: Would it not be better to add a ```using address_family_t ``` for Apple [here](https://github.com/rdkcentral/Thunder/pull/1663/files#diff-230e659e1c108219454970e65208feab6710ccb40dbc41a34434a69c9b0c39dceR99-R103) since no code is actually initializing or using the struct other then returning the ```saddr_hdr.sa_family``` [here](https://github.com/rdkcentral/Thunder/pull/1663/files#diff-230e659e1c108219454970e65208feab6710ccb4dbc41a34434a69c9b0c39dceL132) from a const method

So we can simplify this code and also the code [here](https://github.com/rdkcentral/Thunder/pull/1663/files#diff-230e659e1c108219454970e65208feab6710ccb4dbc41a34434a69c9b0c39dceL131-L135).  

Maybe move the then platform depended ```address_family_t``` to portability?

Source/core/Thread.cpp
HaseenaSainul: done

Source/core/SocketPort.cpp
HaseenaSainul: Updated, but couldn't able to replace #define SOCK_CLOEXEC O_CLOEXEC , since other place we have to use SOCK_CLOEXEC as 0

Source/core/SocketPort.cpp
HaseenaSainul: done

Source/core/SocketPort.cpp
pwielders: @HaseenaSainul that remark sounds scary, how can it be that in certain situations it takes the value of O_CLOEXEC and in others it has the value 0 ?

HaseenaSainul: O_CLOEXEC is used for fcntl() not for socket() case, "int flags = fcntl(l_Result, F_GETFL, 0) | O_CLOEXEC;" So it has to be different, right ?

CMakeLists.txt
HaseenaSainul: done

Source/core/CMakeLists.txt
HaseenaSainul: done

Source/core/NodeId.h
HaseenaSainul: updated as per the discussion

Source/core/SocketPort.cpp
pwielders: Do not know :-) Everything in linux is a file also sockets. Lets just park this than we will look at it in a second round!

HaseenaSainul: Ok, sure

MFransen69 (approved):

pwielders (approved):

MFransen69:
As there was an issue with the github handles used all is moved into a new PR: https://github.com/rdkcentral/Thunder/pull/1680

PR #1666, opened 2024-06-24
core / Tests/unit/core] : improve 'Xgetopt'

msieben:
Modify 'test_xgetopt' accordingly


MFransen69 (dismissed):

msieben:
Closing PR, after feedback from @pwielders and @MFransen69, as the overhead in copying the arrays to avoid permutation of arguments, although typically preferred, is expected to have much impact on resources. Hence, permutation of parameters is unavoidable but acceptable.

PR #1674, opened 2024-06-26
RDK-50885 : Crash handling [Thunder Master]

sethulakshmisk:
Reason for change: Improve Thunder Responsiveness [Crash handling in Thunder Master]
Test Procedure: Simulate crash in rdkservices and verify the call stack activity logs are generated
Risks: NO
Priority: P1

### DIFF
--- a/Source/Thunder/PluginHost.cpp
+++ b/Source/Thunder/PluginHost.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "PluginServer.h"
+#include "core/Library.h"
 #include <fstream>
 
 #ifndef __WINDOWS__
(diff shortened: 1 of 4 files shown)


pwielders:
This seems yet another push of : https://github.com/rdkcentral/Thunder/pull/1618 See those pull requests for the concerns. 
Still looking for answers on what is the usecase? Who overwrites the signal handles why do you now find the rootcause, the plugin that rewrites the signal handlers.
Second remark, *no* need to create a PR per prelease. If applicable we will cherry-pick (feature of SCM to keep track on whats where) the commits to othere branches!! You are breqaking the SCM system this way!

woutermeek:
@sethulakshmisk could you please elaborate on the usecases question?

PR #1682, opened 2024-07-02
[Tests/unit/core] : DataElement improvements

msieben:


pwielders:
Question here is do we want to pay for misuse of the set interface? So if the offset is given bigger than the size should we actively set the buffer pointer to a nullptr. A condition we do check with an ASSERT.
We are building Thunder within the frame of "as small as it gets". If we can safe CPU cycles we do, course CPU and Memory resources are precious . So I think using the software outside of its intended use (passing in an offset > Size) is abuse of the interface, which we catch with the ASSERT and should not require any checking of the CPU cycles (so no checks for asserts and than pass in a nullptr, have it pass in an incorrect nullptr. Keep the code as lean and mean as possible.. Make sure that the testing process is correct and let them test with ASSERTS and if the ASSERT fires, raise a JIRA ticket and fix the incorrect usage of the interface.

msieben:
That is a verbose explanation without pinpointing certain / the code lines in question. Let me give some motivation. Some ASSERTs have a different purpose, eg, prevent numerical overflow even if the described condition holds, and, the buffer might be wrongly indexed, possibly introducing unintended effects, like for memcpy and memmove. I was told ASSERTS can become NOP by a setting in both Release and Debug mode. Hence the impact can be reduced and still be valuable. In contrast, I have no objections if the code is binned. In that case you are welcome to close the PR.

pwielders:
> That is a verbose explanation without pinpointing certain / the code lines in question. Let me give some motivation. Some ASSERTs have a different purpose, eg, prevent numerical overflow even if the described condition holds, and, the buffer might be wrongly indexed, possibly introducing unintended effects, like for memcpy and memmove. I was told ASSERTS can become NOP by a setting in both Release and Debug mode. Hence the impact can be reduced and still be valuable. In contrast, I have no objections if the code is binned. In that case you are welcome to close the PR.

For me the ASSERTS are good, defensive code is always good, but I do not prefer to have the if statements tna later on check that what is in the ASSERT again....

Source/core/DataElement.h
```diff
@@ -167,12 +184,22 @@ namespace Core {
         }
         inline DataElement(const ProxyType<DataStore>& buffer, const uint64_t offset, const uint64_t size = 0)
             : m_Storage(buffer)
-            , m_Buffer(&(buffer->Buffer())[offset])
+            , m_Buffer(buffer->Size() > offset ? &(buffer->Buffer())[offset] : nullptr)
```
pwielders: This is for example additional code where later on you will actually (using defensive coding) ASSERT on this. No need to than check it here as well. so This change is additional code for usage outside of the working area of this software AND this is validated later on using an ASSERT. So the ASSERT is great, which makes the additional code here unnecessary.

Source/core/DataElement.h
```diff
@@ -167,12 +184,22 @@ namespace Core {
...
             , m_Offset(offset)
-            , m_Size(buffer->Size() - offset)
+            , m_Size(buffer->Size() > offset ? buffer->Size() - offset : 0)
```
pwielders: Same here, validated by an assert. Do not change this.

Source/core/DataElement.h
```diff
@@ -206,30 +233,59 @@ namespace Core {
 
         inline DataElement(const DataElement& RHS, const uint64_t offset, const uint64_t size = 0)
             : m_Storage(RHS.m_Storage)
-            , m_Buffer(&(RHS.m_Buffer[offset]))
+            , m_Buffer(RHS.Size() > offset ? &(RHS.m_Buffer[offset]) : nullptr)
```
pwielders: Same here, validated by an assert. Do not change this.

Source/core/DataElement.h
```diff
@@ -206,30 +233,59 @@ namespace Core {
...
             if (size != 0) {
                 m_Size = size;
             }
-
-            ASSERT(offset + size <= RHS.m_Size);
         }
         inline DataElement(DataElement&& move, const uint64_t offset, const uint64_t size = 0)
             : m_Storage(std::move(move.m_Storage))
-            , m_Buffer(&(move.m_Buffer[offset]))
+            , m_Buffer(move.m_Size > offset ? &(move.m_Buffer[offset]) : nullptr)
```
pwielders: Same here, validated by an assert. Do not change this.

Source/core/DataElement.h
```diff
@@ -371,15 +438,19 @@ POP_WARNING()
 
             bool expanded = false;
 
-            // Make sure we are not shrinking beyond the size boundary
-            ASSERT(offset <= m_Size);
+            // Make sure we are not expanding beyond the size boundary
+            ASSERT(size <= (m_MaxSize - m_Size));
+
+            if ((expanded = Size(m_Size + size)) == true) {
```
pwielders, line 444: This is, I guess a failing testcase. Could you shortly refer to this failing test-case using a JIRA ticket or describe the failing testcase here.

pwielders (changes requested):

Source/core/DataElement.h
msieben: @Pierre, I appreciate the feedback and will make the appropriated changes, including filing the Jira ticket.

PR #1698, opened 2024-07-19
Development/ipcchannel dangling handlers testcase

msieben:
Testcase on request of @pwielders


msieben:
Also see https://github.com/rdkcentral/Thunder/pull/1697 which change is absent. Without that modification an ASSERT is triggered at https://github.com/rdkcentral/Thunder/blob/development/ipcchannel_dangling_handlers_testcase/Source/core/IPCConnector.h#L511 from https://github.com/rdkcentral/Thunder/blob/development/ipcchannel_dangling_handlers_testcase/Tests/unit/core/test_ipc.cpp#L873. The possible reason is the registration at https://github.com/rdkcentral/Thunder/blob/development/ipcchannel_dangling_handlers/Source/core/IPCChannel.h#L502 which is not unregistered by https://github.com/rdkcentral/Thunder/blob/development/ipcchannel_dangling_handlers_testcase/Source/core/IPCChannel.h#L327.

MFransen69 (dismissed):

msieben:
@pwielders and @MFransen69 re-ordering code as mentioned in https://github.com/rdkcentral/Thunder/pull/1697 avoids seen issues.

msieben:
Never intended to be merged.

PR #1705, opened 2024-07-26
[COM] Restore channel accessor used by proxystub security

sebaszm:


pwielders (approved):
I guess you need it somewhere in the generated code, but might be good to discuss the dynamics here. The channel might potentially be Released after/during the time you have/get it...

PR #1708, opened 2024-08-01
Call ServiceAdministrator AddRef/Release from Service cons/dest functions to support proper exit for thunder plugin

HaseenaSainul:


pwielders (approved):
Good catch !

PR #1713, opened 2024-08-05
JSON: add error logs for integer overflow case

HaseenaSainul:


VeithMetro (dismissed):

Source/core/JSON.h
```diff
@@ -764,6 +768,12 @@ namespace Core {
                         _set |= ERROR;
                         completed = true;
                     }
+
```
pwielders, line 783: I think you could make this much simpler. Remove the whole current calculation and keep the previous than on line 772 change if (previous != current) to if (previous > _value) and I guess you are done?

pwielders (changes requested):

Source/core/JSON.h
HaseenaSainul: updated the logic as we discussed

pwielders (approved):

PR #1721, opened 2024-08-12
[docs] : add 'guidelines' in 'plugins/interfaces'

msieben:


woutermeek:
I've received some initial feedback which I'll add at the appropriate lines

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+* **Simpler API**: A clean interface that hides complexity makes the API simpler and more intuitive to use. Users of the interface do not need to understand the complexities behind the scenes, leading to a better developer experience.
+
+To achieve these benefits, interfaces should be carefully designed to define clear and minimal contracts for interaction, focusing on what needs to be done rather than how it is done.
+
+Comply to the Single Responsibility Principle (SPR) {#ComplyToSingleResponsibilityPrinciple}
+---------------------------------------------------
+
+The Single Responsibility Principle (SRP) is a fundamental principle in software development that states that an interface should have only one reason to change, which should also adhere to the idea of having a single, well-defined responsibility. Here are some reasons why interfaces should comply with the Single Responsibility Principle:
```
woutermeek: Could this be reworded to make it easier to understand.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+When designing an interface make sure not only to focus on the static behaviour, meaning what methods and notifications are necessary to make sure the client can use it for its purpose, but also on the dynamic behaviour, meaning how should the interface behave when the methods are called at certain times in a certain order and when certain notifications are sent in between (also see the "Less is more" paragraph above on advise on how to reduce the complexity of the dynamic behaviour). This of course also taking into account that the embedded domain we are operating in with Thunder allows for calls being done from multiple applications and/or threads so any call can happen in parallel with any other or any notification. 
+
+Think about data that needs to be changed and/or read atomically. For example it may look logical in some case to have setters/getters for individual items but if the overall consistency of some items combined needs to be guarded it may make more sense to combine these items in one getter and setter, this prevents one application writing ValueA and then ValueB, while another one reads both A and B in between the two writes and is reading an inconsistent set.
```
woutermeek, line 289: Would it help to add diagrams?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+In any case make sure to next to the static behaviour of the interface also document the expected dynamic behaviour so that both the implementer of the interface as well as the user know what to expect. 
+
+Also for notifications one should think very carefully how they behave dynamically and what data the notifications do carry. For example a naive idea might be to always sent a notification on a data change event including the new value of the data. Although this may work perfectly if the data does not change much it will have a big influence on the expected behaviour if the data changes often and the consistency of the data read must be guaranteed.
```
woutermeek, line 298: Would it help to add diagrams here as well?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+5. **Interoperability**: Interfaces facilitate interoperability between different components or even different systems. By adhering to a common interface, different parts of a system or different systems altogether can communicate effectively.
+6. **Testing**: Interfaces help in unit testing by allowing developers to create mock objects that implement interfaces. This enables easier testing of individual components in isolation, leading to more robust and reliable code.
+
+Overall, interface-based development promotes a modular, flexible, and maintainable codebase that is easier to work with and extend over time.
+
+To exploit these advantages to the fullest, Thunder uses Interfaces definitions between plugins and internally. The interface definitions created for Thunder should adhere to certain rules/recommendations and guidelines. This document describes these Rules/Recommendations and guidelines for designing a good interface and the syntax to be used to describe the interface.
+
+Rule, Recommendation or guideline {#RuleRecommendationGuideline}
```
lucienkennedylamb: "Guideline"

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+2. **Recommendation**:
+    * A recommendation is a suggestion or advice on the best course of action, but it is not mandatory.
+    * Recommendations are based on best practices, standards, or guidelines, but individuals or organizations have the freedom to choose whether or not to follow them.
+    * While recommendations are meant to guide decision-making, they allow for flexibility and individual judgment in applying them.
+3. **Guideline**:
+    * A guideline is a set of general principles or rules that provide guidance or direction on how to achieve a particular objective.
+    * Guidelines are more flexible than rules, as they offer suggestions on the most effective way to accomplish a task or goal.
+    * They serve as a framework for good practice and can be adapted or customized based on specific circumstances or requirements.
+
+In summary, rules are strict, mandatory, and non-negotiable directives, recommendations are suggestions or advice that are not obligatory, and guidelines offer flexible principles for achieving a specific objective.
```
lucienkennedylamb, line 67: The worded distinction between recommendation and guideline isn't very clear.
Especially because it is mentioned that recommendations can be a guidelines.
The concrete examples later provide better insight, but wording could be improved here.
In the sections below, the order is different: Rule, Guideline, Recommendation.
Would be great to be consistent in both places.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+=====
+
+Interfaces should not leak implementation details {#InterfacesShouldNotLeakImplementationDetails}
+-------------------------------------------------
+
+Interfaces should be designed to provide a clear and concise abstraction of functionality without exposing the underlying implementation details. Here are some reasons why interfaces should not leak implementation details:
+
+* **Encapsulation**: By hiding implementation details, interfaces ensure that the internal workings of a system or component are not exposed to the outside world. This promotes encapsulation, which is a core principle of object-oriented design.
+* **Flexibility and Maintainability**: When interfaces do not expose implementation details, the underlying code can be changed, optimized, or refactored without affecting the code that depends on the interface. This makes the system more adaptable and easier to maintain.
+* **Abstraction**: Interfaces should provide a high-level abstraction that defines what operations can be performed without detailing how they are performed. This allows developers to focus on the "what" rather than the "how," making the system easier to understand and use.
```
lucienkennedylamb: 'developers' in this sense are interface users, not interface implementers.
Probably worth making that clear to avoid confusion.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+* **Reduced Coupling**: If implementation details are hidden, components that depend on the interface are less tightly coupled to specific implementations. This helps in reducing dependencies and makes it easier to swap out one implementation for another.
+* **Security**: Hiding implementation details can also enhance security by preventing unauthorized access to sensitive information or potential manipulation of internal processes.
+* **Interchangeability**: When interfaces do not leak implementation details, it becomes easier to interchange different implementations of the same interface. This is particularly useful in scenarios where multiple implementations might be needed, such as testing with mock objects or switching between different service providers.
+* **Simpler API**: A clean interface that hides complexity makes the API simpler and more intuitive to use. Users of the interface do not need to understand the complexities behind the scenes, leading to a better developer experience.
+
+To achieve these benefits, interfaces should be carefully designed to define clear and minimal contracts for interaction, focusing on what needs to be done rather than how it is done.
```
lucienkennedylamb, line 85: I'm wondering if this list of rules could be turned into a set of Interface Design Goals as a checklist for interface developers to easily run through?

The guidelines have all the information needed for developers, but it is quite passively wordy.
Sometimes a checklist which asks questions of the reader can be helpful to engage them.
e.g. Simpler API - Is there a way to cut the number of methods or parameters in the interface to reduce its API surface, making it easier to implement and test?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+    * Recommendations are based on best practices, standards, or guidelines, but individuals or organizations have the freedom to choose whether or not to follow them.
+    * While recommendations are meant to guide decision-making, they allow for flexibility and individual judgment in applying them.
+3. **Guideline**:
+    * A guideline is a set of general principles or rules that provide guidance or direction on how to achieve a particular objective.
+    * Guidelines are more flexible than rules, as they offer suggestions on the most effective way to accomplish a task or goal.
+    * They serve as a framework for good practice and can be adapted or customized based on specific circumstances or requirements.
+
+In summary, rules are strict, mandatory, and non-negotiable directives, recommendations are suggestions or advice that are not obligatory, and guidelines offer flexible principles for achieving a specific objective.
+
+Rules {#Rules}
```
lucienkennedylamb, line 69: Do we also want rules on documentation style/quality, method/variable naming convention and mandatory test authoring?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+The Single Responsibility Principle (SRP) is a fundamental principle in software development that states that an interface should have only one reason to change, which should also adhere to the idea of having a single, well-defined responsibility. Here are some reasons why interfaces should comply with the Single Responsibility Principle:
+
+* **Clear and focused functionality**: Interfaces that adhere to the SRP are easier to understand and work with because they have a clear and focused purpose. Developers can quickly grasp what the interface is meant to do and how it should be implemented.
+* **Encourages better design**: By adhering to the SRP, interfaces are more likely to be well-designed and follow best practices in software development. Separating different responsibilities into distinct interfaces helps to create a more modular and maintainable codebase.
+* **Easier to maintain and extend**: When an interface has a single responsibility, changes to that responsibility are less likely to impact other parts of the codebase. This makes it easier to maintain and extend the code without causing unintended consequences or introducing bugs.
```
lucienkennedylamb: Not sure about "changes to that responsibility"?
Changes to or extensions of the interface maybe?
I get the idea here - the blast radius of changes is smaller because the interface responsibility is more fine grained that monolith interfaces.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+* **Promotes code reusability**: Interfaces that adhere to the SRP are more likely to be reusable in different contexts. Since each interface represents a single responsibility, it can be easily reused in various parts of the codebase without introducing unnecessary complexity.
+* **Improved testability**: Interfaces with a single responsibility are often easier to test because their behavior is well-defined and focused. This makes it simpler to write unit tests for classes that implement the interface and ensures that each component of the software can be tested effectively in isolation.
+* **Enhanced flexibility**: Interfaces that follow the SRP are more flexible and adaptable to changes in requirements. If a new responsibility needs to be added or an existing one modified, it can be done more easily without affecting other parts of the codebase.
```
lucienkennedylamb: "If a new responsibility needs to be added" - does this mean by rescoping an existing interface or adding a new interface which is more inline with SRP?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+In conclusion, interfaces that comply with the Single Responsibility Principle are easier to understand, maintain, and extend. By keeping interfaces focused on a single responsibility, developers can create more modular, reusable, and robust software systems.
+
+Interfaces published cannot change, only extend {#InterfacesPublishedCannotChangeOnlyExtend}
+-----------------------------------------------
+
+When it comes to interfaces, it's important to extend them rather than change them directly to maintain compatibility with existing code. By extending an interface, you can add new functionality without breaking the code that already uses the interface. This allows for a more flexible and scalable design, as different classes can implement the new interface without affecting the existing codebase. Additionally, extending interfaces promotes the principle of "open-closed" in software design, where classes are open for extension but closed for modification. This helps in creating more maintainable and modular code.
```
lucienkennedylamb, line 106: Also helps with maintaining test suites, which only need extending in the same way.

Is it worth providing a treatment on deprecated methods?  Never nice, but sometimes inevitable that a new method replaces an old one - but important for backwards compat to keep the old in play.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+ 
+Guidelines {#Guidelines}
+==========
+
+Interface methods that are implementation invariants should be marked const {#InterfaceMethodsThatAreImplementationInvariantShouldBeMarkedConst} 
+---------------------------------------------------------------------------
+
+If a method on an interface is expected to \*not\* functionally change anything in the implementation of the interface, mark the method as const. This allows for passing const interfaces in case the interface needs to be passed to users that should not "control" the implementation.
```
lucienkennedylamb, line 114: Any section which talks about how to write code, like this const method guideline, would benefit from a code example.
Is there a section on const params?  Will read on...

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+The Observer pattern, is a behavioral design pattern where there is a one-to-many dependency between the parent interface and the INotification (nested child interface) so that when one parent interface implementation changes state, all its dependents INotification child interface implementations are notified and updated automatically.
+
+In software development, this pattern is commonly used to establish communication between different parts of a system without them being directly coupled. The Publisher (or Subject) maintains a list of Subscribers (or Observers) interested in being notified of changes, and when an event occurs or state changes, the Publisher notifies all Subscribers by invoking specific methods on them. This allows for a loosely coupled architecture where changes in one component trigger updates in other components without them having explicit knowledge of each other.
+
+As there is a one-to-many dependency the parent interface requires methods to add and remove the INotification. It is a guideline to call these ```core::hresult Register(INotifcation*)``` and ```core::hresult Unregister(const INotification*);```
```
lucienkennedylamb: Typo INotifcation

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+    };
+    Virtual core::hresult Process(ICallback* sink) = 0;
+    Virtual core::hresult Abort() = 0;
+};
+```
+
+Recommendations {#Recommendations} 
+===============
+
+State complete interfaces {#StatetCompleteInterfaces}
```
lucienkennedylamb: Typo #StatetCompleteInterfaces

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+Return ```core::hresult``` on methods {#ReturnCorehresultOnMethods}
+---------------------------------
+As most interfaces are designed to work over process boundaries, it is recommended to cater for issues during the passing of the process boundary, e.g. the other process might have crashed. This means that method calls on interfaces that seem triial but do change the state of an implementation, like ```SetTime()```, may fail due to cross process communication. If the return value of a method is defined as core::hresult, the Thunder COMRPC framework will signal issues with the communication by returning a negative value. The whole positive range from 0-2147483647 (0-0x7FFFFFFF) is free for the implementer to return values.
```
lucienkennedylamb: Typo triial 

Only using positive range for coded return values seems like a new rule.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+Recommendations {#Recommendations} 
+===============
+
+State complete interfaces {#StatetCompleteInterfaces}
+-------------------------
+If a method on the interface can change an implementation into a different state, make sure the interface also has methods to revert the entered state. A good example is an interface that can ```core::hresult Start() = 0``` the executions of a job. As the execution of the job has an expected lifespan which is greater that the calls lifespan, make sure there is a method on the interface as well to bring back the interface to the initial, not started state. E.g. an ```core::hresult::Abort() = 0``` method. 
+
+Return ```core::hresult``` on methods {#ReturnCorehresultOnMethods}
```
lucienkennedylamb, line 174: It it worth making this a rule?
Or can we provide examples of when this may not be needed?

Should this apply to INotification and ICallback methods too?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+    Virtual core::hresult Abort() = 0;
+};
+```
+
+Recommendations {#Recommendations} 
+===============
+
+State complete interfaces {#StatetCompleteInterfaces}
+-------------------------
+If a method on the interface can change an implementation into a different state, make sure the interface also has methods to revert the entered state. A good example is an interface that can ```core::hresult Start() = 0``` the executions of a job. As the execution of the job has an expected lifespan which is greater that the calls lifespan, make sure there is a method on the interface as well to bring back the interface to the initial, not started state. E.g. an ```core::hresult::Abort() = 0``` method.
```
lucienkennedylamb: Typo - extra :: in ```core::hresult::Abort() = 0```

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+Return ```core::hresult``` on methods {#ReturnCorehresultOnMethods}
+---------------------------------
+As most interfaces are designed to work over process boundaries, it is recommended to cater for issues during the passing of the process boundary, e.g. the other process might have crashed. This means that method calls on interfaces that seem triial but do change the state of an implementation, like ```SetTime()```, may fail due to cross process communication. If the return value of a method is defined as core::hresult, the Thunder COMRPC framework will signal issues with the communication by returning a negative value. The whole positive range from 0-2147483647 (0-0x7FFFFFFF) is free for the implementer to return values.
+
+Specifically define the length of the enum to be used {#SpecificallyDefineTheLenthgOfTheEnumToBeUsed}
```
lucienkennedylamb: Typo #SpecificallyDefineThe**_Lenthg_**OfTheEnumToBeUsed

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+-----------------------------------------------------
+To allow for the code generator to optimize for the communication frame, it is recommended to always define the length of an enum if it is used as a parameter. If the enum length is not defined, it is assumed to be 32 bits. If an enum only has less than 256 values, it means 3 bytes are wasted on the line.
```
lucienkennedylamb, line 180: Maybe guidance on future expansion here is needed.  Careful not to limit possible future use.
Changing the enum bit width is typically not extensible later, but adding enum values into an existing bit width should be fine.

If defining the enum length is changed to a rule, we force the developer into making an active choice?

Code example would be good here.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+------------------------------------------------
+
+Describe/define the interface as verbose as possible. This is the contract between two parties, it is better to clearly state what the intention of a parameter is than to leave the intent floating. This means that if a parameter is expected ***not*** to be changed by the implementer define it as const, even if from a C/C++ language point of view it does not add value.
+
+``` c++
+core::hresult Logging(const string& cat, bool& enabled /* @out */) const = 0;
+core::hresult Logging(const string& cat, const bool enabled) = 0;
+```
+
+Although the const bool in the second method has no added value, it is a recommendation to use it, just to indicate clearly that the intend is that the implementation should not and can not change the parameter.
```
lucienkennedylamb: intend -> intent

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+-----------------------------------------------------
+To allow for the code generator to optimize for the communication frame, it is recommended to always define the length of an enum if it is used as a parameter. If the enum length is not defined, it is assumed to be 32 bits. If an enum only has less than 256 values, it means 3 bytes are wasted on the line.
+
+Be verbose even if there is functionally no need {#BeVerboseEvenIfThereIsFunctionallyNoNeed}
+------------------------------------------------
+
+Describe/define the interface as verbose as possible. This is the contract between two parties, it is better to clearly state what the intention of a parameter is than to leave the intent floating. This means that if a parameter is expected ***not*** to be changed by the implementer define it as const, even if from a C/C++ language point of view it does not add value.
+
+``` c++
+core::hresult Logging(const string& cat, bool& enabled /* @out */) const = 0;
```
lucienkennedylamb, line 202: Is there a section which describes use of in/out annotations?
Will read on....

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+The following keywords are mandatory in a COMRPC interface definition, which is an interface that can be used over process boundaries. The interface definition complies with the C/C++ standard rules for header file definitions and thus these interfaces can (and will) be used directly in C/C++ code.
+
+``` c++
+struct EXTERNAL IObject : public virtual Core::IUnknown {
+    enum { ID = ID_OBJECT };
+};
+```
+
+A ***struct*** in the ***C programming language*** (and many derivatives) is a ***composite data type*** (or ***record***) declaration that defines a physically grouped list of variables under one name in a block of memory, allowing the different variables to be accessed via a single ***pointer*** or by the struct declared name which returns the same address. The struct data type can contain other data types so is used for mixed-data-type records such as a hard-drive directory entry (file length, name, extension, physical address, etc.), or other mixed-type records (name, address, telephone, balance, etc.).
```
lucienkennedylamb: Good info on struct in C, but no comparison to its use here in C++, which acts like a class with default public access.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+Less is more {#LessIsMore}
+------------
+
+When designing an interface spend some time thinking on how the interface can be as easy to use, understand and can be consistent without the need of adding a lot of documentation and the need of explaining all the cornercases that can happen when using the interface in a certain way.
+
+As an interface is in principle immutable it is better to spend more time on this before releasing the interface than later on trying to improve the interface or writing a lot of documentation explaining on how to use it in certain situations. 
+
+When designing the interface it is good to use the "less it more" principle when adding methods to it to make sure it can do what should be possible with it. 
+Less methods will make it easier to understand the interface, less documentation will be needed, it is easier to keep it consistent and there is less need to think or document what will happen if you use the methods in an unexpected order.
```
lucienkennedylamb: Less testing too!

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+So this is how this was specified in the ISteamer interface [here](https://github.com/rdkcentral/ThunderInterfaces/blob/a229ea291aa52dc99e9c27839938f4f2af4a6190/interfaces/IStream.h#L105C32-L105C53)
+
+
+Take both Static and Dynamic behaviour into account {#StaticAndDynamic}
+------------------------
+
+When designing an interface make sure not only to focus on the static behaviour, meaning what methods and notifications are necessary to make sure the client can use it for its purpose, but also on the dynamic behaviour, meaning how should the interface behave when the methods are called at certain times in a certain order and when certain notifications are sent in between (also see the "Less is more" paragraph above on advise on how to reduce the complexity of the dynamic behaviour). This of course also taking into account that the embedded domain we are operating in with Thunder allows for calls being done from multiple applications and/or threads so any call can happen in parallel with any other or any notification.
```
lucienkennedylamb, line 287: Use of the @pre for prerequisite conditions before calling a method could be useful here, but couldn't see it in the tag list.
It's part of doxygen though.
At least, any dependent behavior should be called out in the API documentation.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+------------------------------------------------
+
+Sometimes when designing an interface it can be expected that the implementation of a certain action can be quite expensive and take some time. 
+
+In such a case (we usually use an expected duration of a few hundred milliseconds as the limit) or when the duration of an action is non deterministic and worse case would take longer than a few hundred milliseconds, it is advised to make the interface asynchronous. The interface will then only pass the data needed to start the action to the component and add a callback to the interface to make it possible for the implementation to return the result asynchronously. The component will handle the action in the background (do not use threads for this but use the Thunder workerpool) and call the provided callback when the action is completed to signal this to the initiator of the action. Not only will this prevent the call to fail because of JSON-RPC or COM-RPC timeouts but it will also make sure the framework is not blocked internally and also the client can continue without being blocked waiting for the result (or needed to reserve a thread for this).
```
lucienkennedylamb, line 378: I would word the limit on method execution times more strongly.
e.g. Any call could take more than 300ms must use an asynchronous pattern.

Probably worth making developers aware of the JSON-RPC and COM-RPC timeout values.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+For example an interface that will require user interaction for its action (like asking for the pin to be entered) is a perfect example of one that should be made asynchronous as it is of course completely nondeterministic when the call will return as the successful entering of the PIN could take minutes/hours potentially.
+
+Also note, as was pointed out in the "State complete interfaces" paragraph of this page, that there should also be a means to stop the initiated action before it has completed. This is for example needed when the component that started the action needs to shut down and with this it can signal that the action can be stopped as it is no longer needed and the provided callback is no longer valid and should not longer be called.
+
+
```
lucienkennedylamb, line 385: Do we need guidance on choice of JSON over COM when exposing an interface?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+The ID’s are split up in ranges. 
+
+|begin|end|Owner|ID storage|
+|--|-|-|-|
+|0x00000000|0x00000080|Thunder internal|```Thunder/Source/plugin/ids.h```|
+|0x00000080|0x80000000|Public Plugin interfaces|```ThunderInterfaces/interfaces/ids.h```|
+|0x80000000|0xA0000000|Custom Plugin interfaces|```<Custom>```|
+|0xA0000000|0xFFFFFFFF|QA Interfaces|```ThunderInterfaces/interfaces/ids.h```|
+
+Tags/annotations available for optimizing the generated code {#TagsAnnotationsAvailableForOptimizingTheGeneratedCode}
```
anand-ky, line 238: Can we make @text:keep as default for JSON generation. If not can we pls recommend to use @text:keep

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+Describe/define the interface as verbose as possible. This is the contract between two parties, it is better to clearly state what the intention of a parameter is than to leave the intent floating. This means that if a parameter is expected ***not*** to be changed by the implementer define it as const, even if from a C/C++ language point of view it does not add value.
+
+``` c++
+core::hresult Logging(const string& cat, bool& enabled /* @out */) const = 0;
+core::hresult Logging(const string& cat, const bool enabled) = 0;
+```
+
+Although the const bool in the second method has no added value, it is a recommendation to use it, just to indicate clearly that the intend is that the implementation should not and can not change the parameter.
+
+!!! note "TODO"
```
anand-ky: I see there is a Todo for properties. Just adding that we should provide guidelines on when to use properties vs specific getter/setter. I see some info under @property tag. Maybe include that in the guidelines here?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+|--|-|-|-|
+|0x00000000|0x00000080|Thunder internal|```Thunder/Source/plugin/ids.h```|
+|0x00000080|0x80000000|Public Plugin interfaces|```ThunderInterfaces/interfaces/ids.h```|
+|0x80000000|0xA0000000|Custom Plugin interfaces|```<Custom>```|
+|0xA0000000|0xFFFFFFFF|QA Interfaces|```ThunderInterfaces/interfaces/ids.h```|
+
+Tags/annotations available for optimizing the generated code {#TagsAnnotationsAvailableForOptimizingTheGeneratedCode}
+============================================================
+ 
+Tags/annotations that can be used to optimize the generated code, influence the generated JSON-RPC interface from the interface (if desired at all), or document the interface can be found here:
```
anand-ky, line 241: Could we add a section for doxygen commands that should be mandatory (\brief, \param, \details ) to ensure interfaces are properly documented

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+For example an interface that will require user interaction for its action (like asking for the pin to be entered) is a perfect example of one that should be made asynchronous as it is of course completely nondeterministic when the call will return as the successful entering of the PIN could take minutes/hours potentially.
```
anand-ky, line 380: Can we add example of Async interface.

docs/plugin/interfaces/guidelines.md
MFransen69: Hi @anand-ky we cannot make it the default as that would break backwards compatibility for a lot of existing interfaces. 
We think the mandatory rules for these kind of items (e.g. what is mandatory to document, casing used, naming etc.) should be in a separate document that list them for RDK-E, let's discuss this in the next interfaces meeting.

docs/plugin/interfaces/guidelines.md
MFransen69: Hi @anand-ky see my comment above, we propose to have this in a separate document as it may not always be the same.

docs/plugin/interfaces/guidelines.md
ramasat: The last line sounds a bit incomplete.

docs/plugin/interfaces/guidelines.md
ramasat: Strange rule and works only for COMRPC.

Also regarding the negative error codes, perhaps, there can be a reserved range (within the negative numbers) for framework specific error code, not the entire negative number range.

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+The ```Core::IUnknown``` must always be the base of the interface to allow for interface navigation (```QueryInterface```) and life-time management of the interface (reference counting). As typical implementations of an interface might implement multiple interface by one implementation it is mandatory to all route back to a single lifetime management/Navigational base class (diamond structure) of the core::IUnknown implementation. This is realized by making sure that all interface inherit ```virtual``` from the ```Core::IUnknown```. 
+
+As small-as-it-gets is what Thunder stands for, so identifying an interface by a name (or using a GUID, as Microsoft did) is a bit of overkill. Thunder uses an ID which should be unique. The ID to be assigned to the interface is mandatory and it is defines as ```enum { ID = <NUMBER> };``` the number should be unique and thus all ID’s should be registered in ```ids.h``` to guarantee uniqueness.
+
+The ID’s are split up in ranges. 
+
+|begin|end|Owner|ID storage|
+|--|-|-|-|
+|0x00000000|0x00000080|Thunder internal|```Thunder/Source/plugin/ids.h```|
+|0x00000080|0x80000000|Public Plugin interfaces|```ThunderInterfaces/interfaces/ids.h```|
```
ramasat, line 232: What exactly are public plugins and custom plugins?

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
...
+
+    core::hresult Play() const = 0;
+    core::hresult Stop() = 0;
+    core::hresult Pause() const = 0;
+    core::hresult Rewind(const uint32_t speed) = 0;
+    core::hresult Forward(const uint32_t speed) = 0;
+};
+```
+At first this might seem logical but how would you expect the implementation to respond, if you would do Pause(), before Play()? Or Play() after Rewind() without stopping first? You would need to think about all the different options and specify how the interface should respond in that situation. 
+But if you think of it, what all the different methods do is specify a certain playback speed. So if you have one method with which you could set the player speed (and allow a negative number) it could do all the above while at the same time be consistent, easy to understand and no need to explain what will happen in certain situation as just the last set speed is the one that is current.
```
ramasat, line 280: While I fully agree and endorse the "less is more" principle, often it is contextual on "how much is less". Clarity of the interface gets reduced after certain threshold on "less". In my opinion, we might want to make this sounds like a "follow the balanced approach".

docs/plugin/interfaces/guidelines.md
```diff
@@ -0,0 +1,383 @@
+Interface Definition
+====================
+
```
japsen73, line 3: Adding a few generic comment.
First I really like what is already here.

- I think we might be missing some guidelines for how we recommend Interfaces are tagged (e.g @text:keep, @alt, ..) to properly support JsonRPC. 
- Also, we might aligning on using or camel case vs Pascal case for methods, at the same time having it consistent across JsonRPC and Com-RPC
- A section with respect to versioning of interfaces may make sense. When an interface is extended, how is the version updated.
- Recommend we use enum class for types in interfaces

docs/plugin/interfaces/guidelines.md
MFransen69: Hi @lucienkennedylamb see if I understand your remark correctly: you mean when as a 'customer of the interface' you should use one over the other? Or when as an interface provider decide if to provide both or only COM-RPC?

lucienkennedylamb: I was alluding to the interface developer - choosing to provide both or only COM-RPC, so I'm wondering if we should provide guidance on how to make that decision.
But actually you're right about the 'customer' or client of an interface too - If they have both options available to them should they always go with COM-RPC over JSON-RPC which provides stronger typing?

MFransen69: Thanks for clearing this up, I'll add some text to describe at least the options and when you should consider what. 
On your question for the customer of the interface our advise would be to use COM-RPC for the native world, it is strong typed and binary (so will reduce resource usage). JSON-RPC is targeted at the webworld where that is much easier to use.

woutermeek (approved):

PR #1723, opened 2024-08-13
[ProcessContainers] Runtime configurable containers

sebaszm:
Container runtime is selectable via enum string "lxc", "runc", "crun", "dobby", "awc" in root/configuration/containertype. LXC is the default.

/etc/Thunder/plugins/Messenger.json
```
{
 "locator":"libThunderMessenger.so",
 "classname":"Messenger",
 "startmode":"Deactivated",
 "configuration":{
  "root":{
   "mode":"Container",
   "configuration": {
     "containertype": "lxc"
   }
  }
 }
}
```

Each container runtime has separate configuration that is passed to its Initialize(config) method. Initialize() and Deinitialize() replace system-wide Logging() call. Container system configuration has been moved from COM to the runtime implementation (it was specific to LXC). Each runtime can define its own configuration, if any.

/etc/Thunder/config.json
```
{
 "version":"5.0.0",
  ...
 "processcontainers": {
   "lxc": {
     "logging":"100KB",
     "loggingdir":"/tmp"
   }
 },
```

The runtime implementations are registering themselves in ContainerAdminstrator via static initialization.

The previously singleton "administrators" of each implementation are turned into producers that create IContainer concrete objects for the ContainerAdminstrator (i.e. the a factory). The BaseAdministrator common codeis removed as the functionality has been taken over by the factory. The ContainerAdminstrator holds the live containers in a ProxyList.

This solution requires southbound interface changes. LXC and runC implementations are adapted along.


Source/extensions/processcontainers/IProcessContainers.h
```diff
@@ -0,0 +1,127 @@
...
+        virtual ~IContainerProducer() = default;
+
+        // Initializes the producer
+        virtual uint32_t Initialize(const string& /* configuration */) = 0;
+
+        // Shuts down the producer
+        virtual void Deinitialize()= 0;
+
+        // Creates a new container
+        virtual Core::ProxyType<IContainer> Container(const string& id, IStringIterator& searchPaths, const string& logPath, const string& configuration) = 0;
```
pwielders, line 123: We are working everywhere with Interface pointers everywhere, why are we now using a proxyType around the interface (at the end we can put it in a ProxyTypeContainer, but for the interface it makes it more depending on Thunder features) Can we not simply retunr a IContainer here ?

Source/extensions/processcontainers/ContainerAdministrator.h
```diff
@@ -0,0 +1,113 @@
...
+        void Deinitialize();
+
+    public:
+        template<typename T, typename... Args>
+        Core::ProxyType<T> Create(Args&&... args) {
+            return (_containers.Instance<T>(std::forward<Args>(args)...));
+        }
+
+    public:
+        Core::ProxyType<IContainer> Container(const IContainer::containertype type, const string& id,
```
pwielders, line 69: Why is the explicit public Container call, here on 69, not also emplacing the container in the _containers list ? It looks like it is done by calling (from the real implementation line 64 ?)

pwielders: Never mind figured it out. It is all due to the automatic lifetime management from the ProxyListType<>... If that the administrator holds the last reference it drops the container. Correct ?

Source/extensions/processcontainers/ContainerProducer.h
```diff
@@ -0,0 +1,51 @@
...
+
+        ContainerProducerRegistrationType()
+        {
+           ContainerAdministrator::Instance().Announce<PRODUCER>(CONTAINERTYPE);
+        }
+
+        ~ContainerProducerRegistrationType()
+        {
+            ContainerAdministrator::Instance().Revoke(CONTAINERTYPE);
+        }
```
pwielders, line 46: Why not have in this class, the PRODUCER an aggregated member and than announce (CONTAINERTYPE, IContainerProducer& _Producer); Saves you a list with autopointers and new in the administrator and lifetime is guaranteed aligned with the Announce and Revoke. 
It is my allergy to visible "new" 's and std::unique_ptr that triggered this deep dive ;-) and potential simple fix to avoid them..

pwielders (commented):
Since you are on a deserved holiday, oke to merge and will pick the comment up in a next cycle ?

pwielders (approved):
wil pick up the remarks in the next round. Sebas is on holiday now..

PR #1724, opened 2024-08-14
Classname demangling on Proxy interface for linux

nxtum:


sebaszm (dismissed):

Source/Thunder/Controller.cpp
```diff
@@ -1227,13 +1227,25 @@ namespace Plugin {
...
 
+            int status = 0;
             for (const ProxyStub::UnknownProxy* proxy : collection) {
                 IMetadata::Data::Proxy data;
                 data.Instance = proxy->Implementation();
                 data.Interface = proxy->InterfaceId();
                 data.Count = proxy->ReferenceCount();
-                data.Name = proxy->Name();
 
+                #ifdef __LINUX__
```
pwielders: This sounds/looks like code duplication ;-) Time for a new method in Portability like string Demangle(const TCHAR name[]) , cause this code is also here: https://github.com/rdkcentral/Thunder/blob/master/Source/core/Portability.cpp#L227-L244, Tha do not call it here bt there where the proxy->Name() is set, s every one using the ProxyStub::UnknownProxy type can benefit form this work...

Source/com/IUnknown.h
```diff
@@ -148,7 +148,7 @@ namespace ProxyStub {
             , _parent(parent)
             , _channel(channel)
             , _remoteReferences(1)
-            , _name(&(name[7])) /* It always starts with "struct ", remove it...*/
```
pwielders, line 151: This, I think, came for the windows platform (all had "struct " in front of it and I had to drop it). Suggest to "call" the new method you create ;-) from above _string Demangle(const TCHAR name[])_ in portability.h/portability.cpp  where for windows, it is demangled but you stripe of the struct keyword, and for Linux you really demangle? Than the Proxy will give the right name and you do not need to adapt the controller. This way everyone will profit from a proper name of the interface ;-)
ALso have a look at the method ClassNameOnly...

pwielders:
I did a bit of seraching the repo, I think it is already abstracted awya (demangling) see:     https://github.com/rdkcentral/Thunder/blob/56f9e4aafcc1a8f43658061594d9cba84655607c/Source/core/Trace.cpp#L31-L105 Might be usefull to move this to portability and use it from the tree locations (1 where it comes from, 2 the CallStack usage shown in the first link I gave above, 3 the location in Source/com/IUnknown where you assign it to the name (not from location 7 but from location 0). Than there is no need to change anything in Contoller (I hope ;-) )

nxtum:
> I did a bit of seraching the repo, I think it is already abstracted awya (demangling) see:
> 
> https://github.com/rdkcentral/Thunder/blob/56f9e4aafcc1a8f43658061594d9cba84655607c/Source/core/Trace.cpp#L31-L105
> Might be usefull to move this to portability and use it from the tree locations (1 where it comes from, 2 the CallStack usage shown in the first link I gave above, 3 the location in Source/com/IUnknown where you assign it to the name (not from location 7 but from location 0). Than there is no need to change anything in Contoller (I hope ;-) )

I could include Demangling class in Portability.h and have a static self instance of itself for example:
static Demangling& demangleInstance() {static Demangling demangleClassNames; return demangleClassNames;} in the class

and basically call this if i want to use the demangle functions?
I havent dived deeply into it but i think something like this could work, is that what you had in mind with your last comment?

Source/com/IUnknown.h
```diff
@@ -148,7 +148,7 @@ namespace ProxyStub {
...
+            , _name(Demangle(name).c_str())
```
pwielders: Now you creatd a very scary situation.. The temporary string is probably out of scope if you use it. Suggest to move the Demangle(name) -> 405: inline const char* Name() const -> inline string name() const, so we do not pay the penay of a whole string object in each proxy but do get a safe string by the time we need it ;-)

Source/core/Portability.cpp
```diff
@@ -291,18 +293,26 @@ void SleepUs(const unsigned int time) {
     while ( ((result = ::nanosleep(&elapse, &elapse)) != 0) && (errno == EINTR) ) /* Intentionally left empty */;
 }
 
+string Demangle(const TCHAR name[]) {
+    return (name);
```
pwielders: Now both implementations I find do not call the genric Dmangled function, I guess it should be called in one of them or ?????

Source/core/Portability.cpp
```diff
@@ -291,18 +293,26 @@ void SleepUs(const unsigned int time) {
...
 
 #if defined(__WINDOWS__)
 
-void SleepUs(const uint32_t time) {
+    void
+    SleepUs(const uint32_t time) {
     std::this_thread::sleep_for(std::chrono::microseconds(time));
 }
 
+string Demangle(const TCHAR name[]) {
```
pwielders: With comment above, also he we do not call any demangling? Is that expected ?

Source/core/Portability.h
```diff
@@ -787,6 +794,23 @@ namespace Core {
...
+        ~Demangling() = default;
+        Demangling(const Demangling&) = delete;
+        Demangling& operator=(const Demangling&) = delete;
+
+    public:
+        TextFragment Demangled(const char name[]);
+        TextFragment ClassName(const char name[]);
+        TextFragment ClassNameOnly(const char name[]);
+    };
+
```
pwielders: As ClassName/ClassNameOnly and Demangled are now methods on a class with *no* memebers, why not just create methods in a namespace Demangling ? Something like:
`
namespace Thunder { namespace Core { namespace Demangling {
    TextFragment Demangled(const char name[]);
    TextFragment ClassName(const char name[]);
    TextFragment ClassNameOnly(const char name[]);
}}}
`

This makes the usage simpler and simpler...

pwielders (dismissed):

pwielders (dismissed):
Looks good. Sorry you had to do the cleanup but again some technical debt has been resolved!

Source/core/Portability.cpp
```diff
@@ -410,5 +395,67 @@ namespace Core {
     {
         toString(dst, format, ap);
     }
-}
-}
+
+    TextFragment Demangled(const char name[]) {
+
+        char allocationName[512];
```
sebaszm: I understand we're trying to avoid an extra malloc here, but let's remember this will crash if the demangled name happens to be larger than 511 chars (which is not super hard with C++, but hopefully not common in Thunder) as __cxa_demangle() will attempt a realloc in such case.

sebaszm (dismissed):
LGTM.

Source/core/Portability.cpp
pwielders: @sebaszm well spotted! Every day is a schoolday! I read up on the API specification of the "abi::__cxa_demangle" and indeed it will/can realloc if the buffer is too small. So I guess this is a real issue. Good to tackle it!

pwielders (approved):

PR #1727, opened 2024-08-16
Documentation for new code generator tags, august2024

nxtum:


docs/plugin/interfaces/interfaces.md
```diff
@@ -76,6 +76,165 @@ By default, when the `@json` tag is added, all methods in the COM-RPC interface
...
 In older Thunder versions (<R4), JSON-RPC interfaces were defined using separate JSON schema files. These would then need to be manually wired up in the plugin. By using the code-generators, we can eliminate this step, making it much faster and easier to write plugins. It is no longer recommended to create JSON schema files for JSON-RPC interfaces.
 
+#### Overview
+
+* It is now possible to specify a separate JSON-RPC interface for handling connection issues and correct session management, and this will bring more options when the JSON-RPC interface deviates from the COM-RPC interface.
+
+* It is supported to send JSON-RPC over COM-RPC using the 'IDispatch' interface without the need for serialization.
+    * This is particularly useful for out-of-process (OOP) and remote plugins.
+
+* The JSON-RPC generator now supports the usage of nested "plain-old data"(POD) types (such as plain C structs) in the C++ interface.
```
MFransen69, line 83: Also nested PODs are supported

docs/plugin/interfaces/interfaces.md
```diff
@@ -76,6 +76,165 @@ By default, when the `@json` tag is added, all methods in the COM-RPC interface
 
 In older Thunder versions (<R4), JSON-RPC interfaces were defined using separate JSON schema files. These would then need to be manually wired up in the plugin. By using the code-generators, we can eliminate this step, making it much faster and easier to write plugins. It is no longer recommended to create JSON schema files for JSON-RPC interfaces.
 
+#### Overview
+
+* It is now possible to specify a separate JSON-RPC interface for handling connection issues and correct session management, and this will bring more options when the JSON-RPC interface deviates from the COM-RPC interface.
+
+* It is supported to send JSON-RPC over COM-RPC using the 'IDispatch' interface without the need for serialization.
```
MFransen69: ?

docs/plugin/interfaces/interfaces.md
```diff
@@ -76,6 +76,165 @@ By default, when the `@json` tag is added, all methods in the COM-RPC interface
...
+    * This is particularly useful for out-of-process (OOP) and remote plugins.
+
+* The JSON-RPC generator now supports the usage of nested "plain-old data"(POD) types (such as plain C structs) in the C++ interface.
+    * The JSON-RPC generator can now parse these structs and their usage in methods and generate a JSON-RPC interface for such cases.
+
+* `Core::hresult` is required as a return type value of all JSON-RPC enabled interfaces.
```
MFransen69: ?

docs/plugin/interfaces/interfaces.md
```diff
@@ -76,6 +76,165 @@ By default, when the `@json` tag is added, all methods in the COM-RPC interface
 
 In older Thunder versions (<R4), JSON-RPC interfaces were defined using separate JSON schema files. These would then need to be manually wired up in the plugin. By using the code-generators, we can eliminate this step, making it much faster and easier to write plugins. It is no longer recommended to create JSON schema files for JSON-RPC interfaces.
 
+#### Overview
```
MFransen69, line 79: Perhaps we should also document that fixed size arrays are now supported (e.g. array[10], it is used in the new examples below) in certain cases. 
We haven't yet documented although it was already possible but guess we should add it now that you can also use variable lenght arrays (so also for json rpc) : https://github.com/rdkcentral/ThunderInterfaces/blob/a229ea291aa52dc99e9c27839938f4f2af4a6190/interfaces/IDisplayInfo.h#L101

docs/plugin/interfaces/tags.md
```diff
@@ -106,6 +106,8 @@ Ddefines a literal as a known identifier (equivalent of `#define` in C++ code)
 |[@interface](#interface)| Specifies a parameter holding interface ID value for void* interface passing |  | Yes | No |Method paramter|
 |[@length](#length)|Specifies the expression to evaluate length of an array parameter (can be other parameter name, or constant, or math expression)|  | No | Yes | Method Parameter|
 |[@maxlength](#maxlength)|Specifies a maximum buffer length value |  | No | Yes |Method parameter|
+|[@default](#default)|Provides a default value for an unset optional type |  | Yes | Yes |Method parameter|
+|[@encode:base64](#encode:base64)|Encodes C-Style arrays as JSON-RPC base64 arrays |  | Yes | Yes |Method parameter|
```
MFransen69: To be in line with the above documentation think the base64 should not be here but just document #encode influences the generated encoding and list the options (so also bitmask, see Sebastians presentation ) where the details are documented (so more like @text is documented). And isn't it only for json rpc so should it not be in the table below? And previously in the variable size arrays I'm sure it generated a string, so perhaps that is still the case?

docs/plugin/interfaces/interfaces.md
```diff
@@ -76,6 +76,165 @@ By default, when the `@json` tag is added, all methods in the COM-RPC interface
...
+* `Core::Time` is now supported in the IDL header files.
+    * It is presented as an [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) formatted string in JSON-RPC.
+
+* `Core::OptionalType<T>` allows a member to be optional (this superseded @optional), and must be used if an attribute is expected to be optional on JSON-RPC. In COM-RPC the OptionalType can be used to see if a value was set, and in JSON-RPC it is then allowed to omit this parameter.
+    * A @default tag can be used to provide a value, in the case T is not set. See more [here](../tags/#default).
+    * Note: Currently, OptionalType does not work with C-Style arrays.
+
+* JSON-RPC supports the usage of bitmask enums (a combination of enum values can be set in the same enum parameter at the same time).
+    * This is mapped as an array of values in the JSON-RPC interface.
+    * See more information about `/* bitmask */` [here](../tags/#bitmask).
```
MFransen69: Shouldn't we refer to @encode:bitmask (and document thaT @bitmask is deprecated, guess we need to ask sebastian)

MFransen69 (changes requested):

MFransen69 (approved):

PR #1730, opened 2024-08-21
[Windows][Documentation] Fixing config paths, updating the build Thunder on Windows documentation

VeithMetro:
It was initially reported by Tym when he made a clean Windows build using the [instruction from our documentation](https://rdkcentral.github.io/Thunder/introduction/build_windows/), that running the Windows build ends in an ASSERT from MessageDispatcher. The ASSERT fires due to incorrect path in the config file, which is an easy "fix".

Besides, I thought it was a good opportunity to slightly update the build instruction alongside this change, since there were some small tweaks I had to make so that it works from the scratch. The whole Windows build documentation was also revamped a bit to be more clear.

@pwielders do we perhaps want to make another config file with the Domotica paths, so that we can keep both configs?


docs/introduction/build_windows.md
```diff
@@ -1,84 +1,99 @@
-To build Thunder and its components on Windows, you will need Visual Studio installed -  the free community edition is just fine.
+To build the Thunder framework and its components on Windows, you need to have Visual Studio installed. The free Community Edition is sufficient.
```
MFransen69: I see that it is not new here but should we perhaps mention something like "you need to be check if you are entitled to use the free version in your situation, please check the license" before people will say we encouraged them to do so :)
(just checked and sometimes it is even allowed in a commercial situation, e.g. for relatively small companies for developing open source software)

MFransen69 (changes requested):

docs/introduction/build_windows.md
VeithMetro: Good idea, better safe than sorry, I made the change:
_"The free Community Edition is sufficient if you are entitled to use it in your situation, and to make sure of that please check the license."_

MFransen69 (approved):

PR #1737, opened 2024-09-06
Replace SharedBuffer::Semaphore with Core::SharedSemaphore

nxtum:
In SharedBuffer, a class "Semaphore" was being used. This used <semaphore.h>, and the problem was due to sem_clockwait not available on the apple platform and not in the musl library. 

In Sync.h, a class "BinarySemaphore" (previously "BinairySemaphore") was defined, which instead uses <pthread.h>, using mutex and condition variables instead. 

Other differences include constructor values as SharedBuffer::Semaphore would take either TCHAR or a sem_t* depending on platform, whilst Core::BinarySemaphore takes in a bool flag.

Essentially, they try to do the same thing, as both act as a binary semaphore


pwielders:
I think we also need to "move" the shared over process boundary feature into the Binary semaphore than.. Now I do not think that the semaphore (at least on windows), can be shared over process boundaries. That was the name shared between the two processes. I think this feature is now lost and the SharedBuffer no longer works on WIndows...

pwielders:
@nxtum I think @sebaszm came with some good enhancements/improvements to have the SharedSemaphore a static size_t Size() const; operator to expose the size of the (sem_t/windows) struct and to add the max value for this Semaphore..

Source/core/Sync.h
```diff
@@ -181,6 +182,44 @@ namespace Core {
...
+    // ===========================================================================
+
+    class EXTERNAL SharedSemaphore {
+    public:
+        SharedSemaphore() = delete;
+        SharedSemaphore(const Semaphore&) = delete;
+        SharedSemaphore& operator=(const SharedSemaphore&) = delete;
+
+#ifdef __WINDOWS__
+        SharedSemaphore(const TCHAR name[], const uint32_t initValue, const uint32_t maxValue);
```
pwielders, line 195: This could also work on Linux? Right, with the sem_open/sem_close? So why limit this to windows only ?

Source/core/Sync.cpp
```diff
@@ -782,6 +784,188 @@ namespace Core {
...
+            ::CloseHandle(_semaphore);
+        }
+#else
+        sem_destroy(static_cast<sem_t*>(_semaphore));
+#endif
+    }
+
+    size_t SharedSemaphore::Size()
+    {
+#ifdef __WINDOWS__
```
pwielders, line 888: Why would you ASSERT here? Also on window the Semaphore does have a certain size. the size of a HANDLE I guess ? I guess it will nver be called on windows as there we only have the constructor with the names so there is no need to know how big the space needs to be..

Source/core/Sync.h
```diff
@@ -181,6 +182,44 @@ namespace Core {
...
+        SharedSemaphore(void *storage, const uint32_t initValue, const uint32_t maxValue);
+    public:
+#endif
+        ~SharedSemaphore();
+    public:
+        uint32_t Lock(const uint32_t waitTime);
+        uint32_t Unlock();
+        bool IsLocked();
+
+        static size_t Size();
```
pwielders, line 213: I was also expecting a uint32_t Ma() const method ?

Source/core/Sync.cpp
sebaszm: Exactly, this has no useful meaning on Windows and if you think you need to use it on Windows, you're probably doing something wrong, hence I asked to assert here.

Source/core/Sync.cpp
```diff
@@ -782,6 +784,207 @@ namespace Core {
...
+    SharedSemaphore::SharedSemaphore(const TCHAR sourceName[], const uint32_t initCount, VARIABLE_IS_NOT_USED const uint32_t maxCount)
+    {
+        ASSERT(initCount <= 1);
+        ASSERT(maxCount <= 1);
+        ASSERT(initCount <= maxCount);
+#ifdef __WINDOWS__
+        _semaphore = (::CreateSemaphore(nullptr, initCount, maxCount, sourceName));
+#else
+        _name = sourceName;
+        _semaphore = sem_open(sourceName, O_CREAT | O_RDWR, 0644,  initCount);
```
sebaszm: The name of the semaphore should start with a slash /. To unify this with Windows the slash could be prepended in the constructor.

sebaszm: Perhaps we should cater for creating a semaphore with the same name. I,e. differentiate between creating and opening. Now if you mistakenly ask for a semaphore with the same name you'll just open the existing one, and do not know about the fact. Same goes for Windows.

Source/core/Sync.cpp
```diff
@@ -782,6 +784,207 @@ namespace Core {
...
+        ASSERT(_semaphore != SEM_FAILED);
+#endif
+    }
+
+#ifndef __WINDOWS__
+    SharedSemaphore::SharedSemaphore(void* storage, const uint32_t initCount, const uint32_t maxCount)
+        : _semaphore(storage)
+    {
```
sebaszm, line 852: _name is left uninitialized

Source/core/Sync.cpp
```diff
@@ -782,6 +784,207 @@ namespace Core {
...
+
+    SharedSemaphore::SharedSemaphore(const TCHAR sourceName[], const uint32_t initCount, VARIABLE_IS_NOT_USED const uint32_t maxCount)
+    {
+        ASSERT(initCount <= 1);
+        ASSERT(maxCount <= 1);
+        ASSERT(initCount <= maxCount);
+#ifdef __WINDOWS__
+        _semaphore = (::CreateSemaphore(nullptr, initCount, maxCount, sourceName));
+#else
+        _name = sourceName;
```
sebaszm: You only store the pointer to the name. It may become invalid by the time you reach unlink.

sebaszm (changes requested):

Source/core/SharedBuffer.h
```diff
@@ -65,34 +65,6 @@ namespace Core {
...
-        private:
-#ifdef __WINDOWS__
-            HANDLE _semaphore;
-#else
-            sem_t* _semaphore;
-#endif
-        };
         struct Administration {
 
             uint32_t _bytesWritten;
```
pwielders, line 69: Now we did our best to hide the imeplementation details as much as possible, however, here we do refer to the hidden details. Can we not create something like:
 ```
 #ifndef __WINDOWS__
            uint8_t _producerSpace[SharedSemapore::Size()];
            uint8_t _consumerSpace[SharedSemapore::Size()];
#endif
       ```

Source/core/Sync.cpp
pwielders: @nxtum  / @sebaszm I am not religious about t, but at the end this now does leak implementation details through the interface. Seems that the static size_t Size() const operator is leaking the detai that the Size() call is *not* valid sometimes.
At the end, on windows it can not go wrong, even if you would assume that you have to allocate memory, if you than would like to use the constructor (you have to use the constructor that passes in storage space) you get a compilation error cause it does not exist (you even find the issue earlier than running it)...

@sebaszm agree that the meaning on windows is useless but disagree that you can use it incorrectly cause than you would get a compiler error..

So I have a slight preference to *not* show the implementation details and suggest to make it statis, see other remark on hiding sem_t stuff..

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+        return 0;
+#else
+        return sizeof(sem_t);
+#endif
+    }
+
+    uint32_t SharedSemaphore::MaxCount() 
+    {
+        // Currently max count is 1, as it is implemented as a binary shared semaphore
+        return 1;
```
pwielders, line 898: Can we, for windows, not return the given MaxValue ? Cause now whatever we set on windows, it return 1 :-)

Source/core/Sync.h
```diff
@@ -181,6 +182,49 @@ namespace Core {
...
+        static uint32_t MaxCount();
```
pwielders: As the interface, assuming agnostic to the implementation, can set the Max(Count) value, I think this method can not be static. See also remark to, on windows, return the actual set value.

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+    SharedSemaphore::SharedSemaphore(const TCHAR sourceName[], const uint32_t initCount, VARIABLE_IS_NOT_USED const uint32_t maxCount)
+    {
+        ASSERT(initCount <= 1);
+        ASSERT(maxCount == 1);
+
+#ifdef __WINDOWS__
+        _semaphore = (::CreateSemaphore(nullptr, initCount, maxCount, sourceName));
+        ASSERT(_semaphore != nullptr);
+#else
+        _name = "/" + string(sourceName);
```
pwielders, line 832: The ASSERT in the Size(), which I think is debatable, should be better here:         ASSERT(maxCount == 1); as this is the only supported value right now for Linux..

pwielders (commented):
Questions and discussions added :-) Maybe a confcall/skype chat in TPX ?

Source/core/Sync.cpp
sebaszm: The code that needs to use Size(), ie. code that will use shared memory and not a name will have to be flagged #ifndef WINDOWS anyway, so I think we could remove the Size() method on Windows completely.

Source/core/SharedBuffer.h
sebaszm: Both sem_t* and HANDLE can be hidden behind void*. Again no purpose on Size() on Windows.

Source/core/Sync.cpp
sebaszm: If we allow setting maxCount to more than 1 on Windows, it means that the code working windows could not work on Linux.

Source/core/Sync.h
sebaszm: Agree it should not be static as one day we could implement maxCount>1 on Linux. No reason to limit it on the interface.

Source/core/Sync.cpp
pwielders: That would for me also be a viable option... Than I will run  into this at compile time and not runtime..

pwielders (dismissed):

Source/core/Sync.cpp
```diff
@@ -782,6 +784,223 @@ namespace Core {
...
+    //----------------------------------------------------------------------------
+    //----------------------------------------------------------------------------
+
+    //----------------------------------------------------------------------------
+    // CONSTRUCTOR & DESTRUCTOR
+    //----------------------------------------------------------------------------
+
+
+    SharedSemaphore::SharedSemaphore(const TCHAR sourceName[], const uint32_t initCount, VARIABLE_IS_NOT_USED const uint32_t maxCount)
+    {
```
sebaszm, line 825: _semaphore is not initialized

Source/core/Sync.cpp
```diff
@@ -782,6 +784,223 @@ namespace Core {
...
+
+        if (initCount != 0 && maxCount != 0) {
+            VARIABLE_IS_NOT_USED int result =  sem_init(static_cast<sem_t*>(_semaphore), 1, initCount); 
+            ASSERT(result != -1);
+        }
+    }
+
+    SharedSemaphore::SharedSemaphore(void* storage) 
+    : _semaphore(storage), _name("")
+    {
```
sebaszm, line 863: assert storage is not null

Source/core/SharedBuffer.cpp
```diff
@@ -26,148 +26,37 @@ namespace Thunder {
...
+    #endif
         , _customerAdministration(PointerAlign(&(reinterpret_cast<uint8_t*>(_administration)[sizeof(Administration)])))
     {
         Align<uint64_t>();
     }
     SharedBuffer::SharedBuffer(const TCHAR name[], const uint32_t mode, const uint32_t bufferSize, const uint16_t administratorSize)
         : DataElementFile(name, mode | File::SHAREABLE | File::CREATE, bufferSize)
-        , _administrationBuffer((string(name) + ".admin"), mode | File::SHAREABLE | File::CREATE, administratorSize + sizeof(Administration) + (2 * sizeof(void*)) + 8 /* Align buffer on 64 bits boundary */)
+        , _administrationBuffer((string(name) + ".admin"), mode | File::SHAREABLE | File::CREATE, sizeof(Administration) + (SharedSemaphore::Size() * 2) + administratorSize + 
+          ((administratorSize % 8 == 0) ? 0 : (8 - (administratorSize % 8))) /* Align buffer on 64 bits boundary */)
```
sebaszm: alignment attempt assumes sizeof(administrtation) and Size() return aligned sizes

sebaszm (changes requested):

Source/core/Sync.cpp
nxtum: I initialise _semaphore later as: 
_semaphore = (::CreateSemaphore(nullptr, initCount, maxCount, sourceName)); // windows
_semaphore = sem_open(_name.c_str(), O_CREAT | O_RDWR, 0644,  initCount); // linux

pwielders (dismissed):
Yee we saved potentially 4 bytes on a 32 bit platform! Looks very good now!

Source/core/SharedBuffer.cpp
```diff
@@ -26,148 +26,38 @@ namespace Thunder {
...
         , _consumer((string(name) + ".consumer").c_str())
-#else
-        , _producer(&(_administration->_producer))
-        , _consumer(&(_administration->_consumer))
-#endif
+    #else
+        , _producer(reinterpret_cast<uint8_t*>(_administration) + sizeof(Administration))
+        , _consumer(reinterpret_cast<uint8_t*>(_administration) + sizeof(Administration) + SharedSemaphore::Size())
+    #endif
         , _customerAdministration(PointerAlign(&(reinterpret_cast<uint8_t*>(_administration)[sizeof(Administration)])))
```
sebaszm: customerAdministration needs to be adjusted, now it overwrites the semaphore structures

Source/core/SharedBuffer.cpp
```diff
@@ -26,148 +26,38 @@ namespace Thunder {
...
-        , _consumer(&(_administration->_consumer))
-#endif
+    #ifdef __WINDOWS__
+        , _producer((string(name) + ".producer").c_str(), 1, 1)
+        , _consumer((string(name) + ".consumer").c_str(), 0, 1)
+    #else
+        , _producer(reinterpret_cast<uint8_t*>(_administration) + sizeof(Administration), 1, 1)
+        , _consumer(reinterpret_cast<uint8_t*>(_administration) + sizeof(Administration) + SharedSemaphore::Size(), 0, 1)
+#   endif
         , _customerAdministration(PointerAlign(&(reinterpret_cast<uint8_t*>(_administration)[sizeof(Administration)])))
```
sebaszm: Same here wrt customerAdministration

Source/core/SharedBuffer.cpp
```diff
@@ -26,148 +26,38 @@ namespace Thunder {
...
+        , _consumer(reinterpret_cast<uint8_t*>(_administration) + sizeof(Administration) + SharedSemaphore::Size())
+    #endif
         , _customerAdministration(PointerAlign(&(reinterpret_cast<uint8_t*>(_administration)[sizeof(Administration)])))
     {
         Align<uint64_t>();
     }
     SharedBuffer::SharedBuffer(const TCHAR name[], const uint32_t mode, const uint32_t bufferSize, const uint16_t administratorSize)
         : DataElementFile(name, mode | File::SHAREABLE | File::CREATE, bufferSize)
-        , _administrationBuffer((string(name) + ".admin"), mode | File::SHAREABLE | File::CREATE, administratorSize + sizeof(Administration) + (2 * sizeof(void*)) + 8 /* Align buffer on 64 bits boundary */)
+        , _administrationBuffer((string(name) + ".admin"), mode | File::SHAREABLE | File::CREATE, sizeof(Administration) + (SharedSemaphore::Size() * 2) + administratorSize +
```
sebaszm: _administration may be moved by PointerAlign, I think we need to cater for it too so another + sizeof(void*) to be on the safe side
_customerAdministration is may also be adjusted by PointerAlign, so another (sizeof(void*)...

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+        ASSERT(initCount <= 1);
+        ASSERT(maxCount == 1);
+
+#ifdef __WINDOWS__
+        _semaphore = (::CreateSemaphore(nullptr, initCount, maxCount, sourceName));
+        ASSERT(_semaphore != nullptr);
+#else
+        _name = "/" + string(sourceName);
+        _semaphore = sem_open(_name.c_str(), O_CREAT | O_RDWR, 0644,  initCount);
```
sebaszm: I think it also needs O_EXCL, so we block accidental reuse of the semaphore with this constructor dedicated to creating

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+
+#ifndef __WINDOWS__
+    SharedSemaphore::SharedSemaphore(void* storage, const uint32_t initCount, const uint32_t maxCount)
+        : _semaphore(storage), _name("")
+    {
+        ASSERT(storage != nullptr);
+        ASSERT(initCount <= 1);
+        ASSERT(maxCount == 1);
+
+        if (initCount != 0 && maxCount != 0) {
```
sebaszm: This the culprit: the semaphore still needs initialization even if initCount is 0! (so only check for maxCount != 0)

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+            VARIABLE_IS_NOT_USED int result =  sem_init(static_cast<sem_t*>(_semaphore), 1, initCount);
```
sebaszm: Seems that depending on implementation sem_init may need the sem_t zero-cleared. Perhaps it's better to stay on the safe side?...

Source/core/Sync.cpp
```diff
@@ -782,6 +784,224 @@ namespace Core {
...
+        }
+#else
+        VARIABLE_IS_NOT_USED int result = sem_post(static_cast<sem_t*>(_semaphore));
+        ASSERT((result == 0) || (errno == EOVERFLOW));
+#endif
+        return ERROR_NONE;
+    }
+
+    bool 
+    SharedSemaphore::IsLocked()
```
sebaszm: Maybe it's better to change this to Count() and return the current semaphore lock value? IsLocked() limits this to a binary semaphore interface again.

sebaszm (changes requested):
(also please mind the whitespaces)

sebaszm (dismissed):

Source/core/Sync.cpp
pwielders: I do agree here with @sebaszm but see it is not adapted, so I guess this is the last bit we need to change. I assume that if Cunt() const -> 0 it IsLocked() const  -> true

Source/core/Sync.h
```diff
@@ -181,6 +182,45 @@ namespace Core {
...
+        */
+        SharedSemaphore(void *storage, const uint32_t initValue, const uint32_t maxValue);
+        SharedSemaphore(void *storage);
+    public:
+#endif
+        ~SharedSemaphore();
+    public:
+        uint32_t Lock(const uint32_t waitTime);
+        uint32_t Unlock();
+        bool IsLocked();
```
pwielders: I think the method should be uint32_t Count() const;

pwielders (dismissed):

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+    //----------------------------------------------------------------------------
+
+    //----------------------------------------------------------------------------
+    // CONSTRUCTOR & DESTRUCTOR
+    //----------------------------------------------------------------------------
+
+#ifdef __WINDOWS__
+    class WindowsAPI {
+    public:
+        WindowsAPI(WindowsAPI&&) = delete;
```
sebaszm, line 800: Missing destructor even if it's default :)

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+        WindowsAPI(const WindowsAPI&) = delete;
+        WindowsAPI& operator=(WindowsAPI&&) = delete;
+        WindowsAPI& operator=(const WindowsAPI&) = delete;
+
+        WindowsAPI() {
+            HMODULE ntdll = GetModuleHandle(_T("ntdll.dll"));
+            _ntQuerySemaphore = reinterpret_cast<_NTQuerySemaphore>(GetProcAddress(GetModuleHandle(_T("ntdll.dll")), "NtQuerySemaphore"));
+            ASSERT (_ntQuerySemaphore != nullptr);
```
sebaszm, line 808: GetModuleHandle is called twice

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+        }
+
+        uint32_t GetSemaphoreCount(HANDLE parameter) {
+            ASSERT (_ntQuerySemaphore != nullptr);
```
sebaszm: This assertion is unnecessary, it would've asserted already in the constructor.

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+            SEMAPHORE_BASIC_INFORMATION basicInfo;
+            NTSTATUS status;
+            status = _ntQuerySemaphore(parameter, 0, &basicInfo, sizeof(SEMAPHORE_BASIC_INFORMATION), nullptr);
+            if (status == ERROR_SUCCESS) {
+                return (basicInfo.CurrentCount);
+            }
+            return(0);
```
sebaszm: can we make one return statement?

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+        WindowsAPI& operator=(WindowsAPI&&) = delete;
+        WindowsAPI& operator=(const WindowsAPI&) = delete;
+
+        WindowsAPI() {
+            HMODULE ntdll = GetModuleHandle(_T("ntdll.dll"));
+            _ntQuerySemaphore = reinterpret_cast<_NTQuerySemaphore>(GetProcAddress(GetModuleHandle(_T("ntdll.dll")), "NtQuerySemaphore"));
+            ASSERT (_ntQuerySemaphore != nullptr);
+        }
+
+        uint32_t GetSemaphoreCount(HANDLE parameter) {
```
sebaszm: const? :)

Source/core/Sync.cpp
```diff
@@ -782,6 +784,245 @@ namespace Core {
...
+            SEMAPHORE_BASIC_INFORMATION basicInfo;
+            NTSTATUS status;
+            status = _ntQuerySemaphore(parameter, 0, &basicInfo, sizeof(SEMAPHORE_BASIC_INFORMATION), nullptr);
+            if (status == ERROR_SUCCESS) {
+                return (basicInfo.CurrentCount);
+            }
+            return(0);
+        }
+
+        _NTQuerySemaphore _ntQuerySemaphore;
```
sebaszm, line 818: This can be private

sebaszm (changes requested):
minor things,  we're almost there!

pwielders (approved):
LGTM :-)

sebaszm (approved):

PR #1747, opened 2024-09-13
Development/bluez headers download

bramoosterhuis:
Bluetooth headers are mandatory on a more generic level. So I moved the find script so its available for all components that require it.  

From bluez5 v5.64 all inclusive language was removed from their project. To deal with that a check was added and that check was started to be copied. To check once, I added the "NO_INCLUSIVE_LANGUAGE" define to the find module and added it in the link interface. Now this define will be propagated to all the components that include the bluez5 headers and use the Bluez5UtilHeaders.

Since Thunder requires only requires the bluez5 headers and the talks to the kernel directly, the version of the bluez5 version is not really a hard link. The headers need to match the kernel structures.  Also a header that is marked private by the bluez5 project is required by Thunder. There for I added  a option to download the headers configure time to prevent patching of the bluez5 on the build and make life easier for the automated builds. The version and git repository can be configured and default to the latest release while writing this.


VeithMetro (approved):
Looks great, can't wait to try it out in Actions if that means I won't have to manually download and install bluez headers 😄

sebaszm (approved):
Thanks for sorting this out!

sebaszm:
Which header is the private one?

bramoosterhuis:
> Which header is the private one?

Thats the ```mgmt.h``` its not installed here:
https://github.com/bluez/bluez/blob/d8a4b126c1d7ef1cf8994681ccb413b59bee29b5/Makefile.am#L73

VeithMetro:
> > Which header is the private one?
> 
> Thats the `mgmt.h` its not installed here: https://github.com/bluez/bluez/blob/d8a4b126c1d7ef1cf8994681ccb413b59bee29b5/Makefile.am#L73

Indeed, that's the one which was needed to be manually download and put where it belongs, but now it should be way easier with these changes

bramoosterhuis:
Lets hear what @pwielders thoughts are about it.

pwielders:
@bramoosterhuis love automation, as always! Good work!

pwielders (approved):

PR #1751, opened 2024-09-18
Use ServiceAdministrator::LoadLibrary to load any libraries

HaseenaSainul:
dlopen is essential in Thunder to manage shared libraries during plugin activation. dlopen has a process wide lock and hence must be managed as a singleton. This change helps streamline all dlopens are done through the ServiceAdministrator.


pwielders:
We should cherry pick so SCmcan track on all commits located where . If we create new patches for the same issue on each branch, merging later on becomes more difficult, and a cherry pick is very easly done.. Suggets to close this and cherry-pick: https://github.com/rdkcentral/Thunder/pull/1748/commits/60ff752d090da1db6c6d8e33702e0fd27f9bf726 this to 4.4.

MFransen69 (approved):

PR #1755, opened 2024-09-18
Development/refine bluez headers handling

bramoosterhuis:
With this change the bluez headers will be part of the core. This fixes dependency complications in other components down the line.


VeithMetro (changes requested):
Like we discussed on Skype, I tried building with these newest changes, and while it works perfectly fine in Lithosphere, I got a compilation error while building with just cmake on my local Ubuntu (probably the same would happen in actions):
```
/home/veith/Thunder/Source/extensions/bluetooth/HCISocket.h:27:10: fatal error: core/bluez/mgmt.h: No such file or directory
   27 | #include <core/bluez/mgmt.h>
      |          ^~~~~~~~~~~~~~~~~~~
```
Which seems strange, since the mgmt.h header seems to be installed correctly:
```
veith@ubuntu:~$ ls build/Thunder/Source/core/bluez
a2mp.h  amp.h  bluetooth.c  bluetooth.h  bluez.pc.in  bnep.h  cmtp.h  hci.c  hci.h  hci_lib.h  hidp.h  iso.h  l2cap.h  mgmt.h  rfcomm.h  sco.h  sdp.c  sdp.h  sdp_lib.h  uuid.c  uuid.h
```
So probably an include path is missing for some reason

VeithMetro (dismissed):
Looks great with the latest changes, managed to build both the extension, and then all of the Bluetooth plugins with just cmake!

Source/core/CMakeLists.txt
```diff
@@ -206,12 +206,27 @@ endif()
 if(BLUETOOTH_SUPPORT)
     target_compile_definitions(${TARGET} PUBLIC __CORE_BLUETOOTH_SUPPORT__)
 
-    find_package(Bluez5UtilHeaders REQUIRED)
+    include(GetBluezHeaders)
 
-    target_link_libraries(${TARGET}
+    GetBluez5UtilHeadersFiles(Bluez5UtilHeadersFiles)
+
+    install(FILES ${Bluez5UtilHeadersFiles}
```
pwielders, line 213: Loved the change to GetBluezHeaders where we removed the 5 from the name, should we not do that everywhere? What if we start working with Bluez6 ?

pwielders (changes requested):
Minor one but to avoid issues in the future, lets remove the major version (5) from the Bluez variables. The FindBluezFile cmake file also does not contain the Major version (Which I think is good)

Source/core/NodeId.h
```diff
@@ -46,9 +46,11 @@
 #endif
 
 #ifdef __CORE_BLUETOOTH_SUPPORT__
-#include <../include/bluetooth/bluetooth.h>
-#include <../include/bluetooth/hci.h>
-#include <../include/bluetooth/l2cap.h>
+PUSH_WARNING(DISABLE_WARNING_PEDANTIC)
+#include "bluez/bluetooth.h"
```
sebaszm: I'm not a hippie to have feelings, but naming the folder bluez feels like Thunder is being dependent on bluez but we're only dependent on the bluetooth SIG-defined magic numbers / structures only.

sebaszm (dismissed):

pwielders:
@sebaszm / @bramoosterhuis I have no feelings at all (rofl) although I do agree on the somment, we do only need the specifics of the bluetooth SIG-defined magic numbers / structures, used and implemented in the kernel and as such, @bramoosterhuis and myself had the discussion f we could not get this info from the kernel header files (if I understand correctly that was even harder) the retrieval of this information from Bluez is a work around. If we want to change this name to something else (not bluetooth, as we already have that) to bluetooth_sig?? or SIG_bleutooth, or whtever, I am fine as well. Not religious about it :-)

sebaszm:
> @sebaszm / @bramoosterhuis I have no feelings at all (rofl) although I do agree on the somment, we do only need the specifics of the bluetooth SIG-defined magic numbers / structures, used and implemented in the kernel and as such, @bramoosterhuis and myself had the discussion f we could not get this info from the kernel header files (if I understand correctly that was even harder) the retrieval of this information from Bluez is a work around. If we want to change this name to something else (not bluetooth, as we already have that) to bluetooth_sig?? or SIG_bleutooth, or whtever, I am fine as well. Not religious about it :-)

bluetooth_data, bluetoot_defs, bluetooth_definitions, bluetooth_defines... not religious at all. Approved it already! :)

Source/core/NodeId.h
bramoosterhuis: Correct, but still there is a include path, so build time there is a dependency. By naming it bluez at least it's clear who's API we expect.

Source/core/CMakeLists.txt
bramoosterhuis: Good point! I made it more specific now, let's welcome bluez6 to earth

sebaszm (approved):

pwielders (approved):

VeithMetro (approved):

PR #1758, opened 2024-09-20
[Core] Improve path normalization

sebaszm:
The signatures are changed to:
`File::Normalize(const string& location, const bool inScopeOnly = false, const bool allowDirs = false)`
and
`Directory::Normalize(const string& location, const bool inScopeOnly = false)`

The isValid output parameter is removed, in case of error empty string is returned. 

Additional, defaulted parameters inScopeOnly and allowDirs specify failure criteria:

- inScopeOnly - true: going past current dir is failure
- allowDirs - false: paths obviously pointing to a directory (ends with / or . or ..) is failure

Directory::Normalize() always outputs a directory ending with slash, especially:
- . becomes ./
- .. becomes ../
- foo/../.. becomes ../
- foo/.. becomes ./
- /foo/.. becomes /

This is now consistent with paths ending with a directory name and always allows concatenating.

Also:
- foo/../../.. becomes ../../
- //foo////bar/././././/////baz////./ becomes /foo/bar/baz

For detailed cases see test code.


pwielders:
I assume that if I ass in ScopeOnly and the path is:
foo/../.. becomes empty ???? So that if you want to esape the current path I can detect that?

pwielders (approved):

PR #1763, opened 2024-09-24
Expose proxies by default

nxtum:
Exposes all proxies by default when no id is specified via: Controller.1.proxies


sebaszm (commented):
how about std::vector::insert  ?

sebaszm (dismissed):

pwielders:
@nxtum / @sebaszm there is probably a bit of a bigger story here. Lifetime :-) There was no addref in the past there is no addref now on the list that we passout, but Ithink it is a bit ticky (a larger flaw) The proxies are reference counted but as we pass them out here we do not increase the reference count which could be "scary"... Lets discuss tomorrow in the Sync up call..

sebaszm:
> @nxtum / @sebaszm there is probably a bit of a bigger story here. Lifetime :-) There was no addref in the past there is no addref now on the list that we passout, but Ithink it is a bit ticky (a larger flaw) The proxies are reference counted but as we pass them out here we do not increase the reference count which could be "scary"... Lets discuss tomorrow in the Sync up call..

Ah, I didn't consider that.

Source/Thunder/PluginHost.cpp
```diff
@@ -712,15 +712,13 @@ POP_WARNING()
                                 printf("Link: %s\n", index.Current().Remote.Value().c_str());
                                 printf("------------------------------------------------------------\n");
 
-                                RPC::Administrator::Proxies proxies;
-
-                                RPC::Administrator::Instance().Allocations(index.Current().ID.Value(), proxies);
-
-                                for (const ProxyStub::UnknownProxy* proxy : proxies) {
+                                RPC::Administrator::Instance().Allocations(index.Current().ID.Value(), [](const std::vector<ProxyStub::UnknownProxy*>& proxies) {
+                                    for (const auto& proxy: proxies) {
```
pwielders, line 716: Is this to test and see if I pay attention ;-) Please remove the printfs, and fill the proxies, cause that is a bit essential  ;-)

nxtum: Hi Pierre, the printfs are from existinng code:
https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginHost.cpp#L721

nxtum: and I am not sure what you mean by filling the proxies:
the old code includes this:
RPC::Administrator::Proxies proxies; 

but i removed it as this part of the function only displays data, so I just directly use: 
[](const std::vector<ProxyStub::UnknownProxy*>& proxies)

I dont think there is a need to pass something like a new vector like
[&proxies]?
Or am i missing something?

pwielders: Ohhh sorry. You are right. My bad. This is the console print. I thought it was the JSONRPC call. Sorry!

pwielders (approved):

sebaszm (approved):

PR #1764, opened 2024-09-24
SetEnv added for plugin Config as well

HaseenaSainul:


Source/plugins/IShell.h
```diff
@@ -235,6 +235,9 @@ namespace PluginHost {
         //! Substituted Config value
         virtual string Substitute(const string& input) const = 0;
 
+        //! Substituted environments list
```
pwielders: I am hoping we do not need to feed this through the IShell. What is the reason for adding this to the IShell ?

Source/com/Communicator.h
```diff
@@ -503,6 +515,9 @@ namespace RPC {
                 if (instance.Threads() > 1) {
                     _options.Add(_T("-t")).Add(Core::NumberType<uint8_t>(instance.Threads()).Text());
                 }
+                if (instance.Environments().empty() == false) {
```
pwielders: how about adding a -E option for each environment variable we need to add, see: https://github.com/rdkcentral/Thunder/pull/1761/files#diff-eb8c33b331954e93d7a9e3832f0aafc222668a0e3290e33259680145c3f98adcR502-R505 I guess the it is save to than say -E <key>="<value>", since key is retricted to be alphanumerical only, this should be safe and to "signal" override or not, use -E <key>#"value" or use a different identifier> like -e? THis would be more in line with command line paramter handling than passing a JSON blob..

Source/Thunder/PluginServer.h
```diff
@@ -1277,7 +1277,24 @@ namespace PluginHost {
             string Substitute(const string& input) const override {
                 return (_administrator.Configuration().Substitute(input, PluginHost::Service::Configuration()));
             }
-           Core::hresult Metadata(string& info /* @out */) const override {
+            string SubstituteList(const string& envList) const override {
```
pwielders: prefer to keep distinct types here for changes in the future, so please define a struct like { string key; string value; bool override; } and work with a std::vector<struct environment>. I prefer to keep this strong tuped as trings are:
a) more resoyrce consuming (JSON)
b) not strong typed,.

pwielders (changes requested):
See: https://github.com/rdkcentral/Thunder/pull/1761 which also extends the data flowing to PluginProcess. Prefer to keep it all strong typed (use a C struct or class) and send each element as a seperate (yet repeating) entry on the command line. See inline comments.

I think it is good to also first include thepull request sted above as this might otherswise be difficult to merge (both add functionality in the same areas)

Source/plugins/Configuration.h
```diff
@@ -92,6 +92,7 @@ namespace Plugin {
             Core::JSON::Boolean Override;
         };
         using EnvironmentList = Core::JSON::ArrayType<Environment>;
+        static constexpr const TCHAR EnvFieldSeparator = _T(';');
```
pwielders: Good improvements! This might sound a bit like nitpicking but the idea is that the information can be used in different places (for the future) without having to parse the information. Maybe for this situation it is not really needed yet, but there are some cases when we might want to have access directly to the key and its value and maybe potentially the state (override yes/no).

To accomplish this I think it is good to define here a struct (define it in the RPC::Object class, as I would really like to put this information in that object as well as it is part of and used by this object).
Such a struct (within the RPC::Object) and would have at least the elements, key (string), value (string) and an enum (to define a type system wide (override == true) or local (override ==false)) or a boolean... Slight preference for the enum as it is easier to extend if we need to in the future.

Than if this struct is part of the RPC::Object, and contained as a std::vector<struct RPC::Object::Environment> in this RPC::Object, I guess you do not need to extend the IShell, as the RPCObject is created in Shell.cpp (Instantiate), forwarded to the PluginServer through the IComlink interface (See: https://github.com/rdkcentral/Thunder/blob/master/Source/plugins/Shell.cpp#L85 ) and there (in the implementation of the PluginServer of Instantiate (here https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L2416) you can apply the substitute of the %...% on the value part of each entry in the class, before sending them out,

So I guess that would rmean we do not have to extend the IShell interface with a *non* COMRPC compatible method. Which I guess is good news...

Also if it the information is in a struct (key, value, type) we can easily wrap this into a IIterator for "Off Host" instantiation (done through COMRPC) in the [Source/Thunder/IRemoteInstantiation.h](https://github.com/rdkcentral/Thunder/pull/1764/commits/bb0c2b13295c2d4f438732971ab501d1c005973d#diff-210ee7d13f88174601f2a5fa5f04e9e677e8410f9156da7414542308a0abfc69) changed above. Since this *must* be posisble to run over COMRPC for off host instantiation.

Source/Thunder/IRemoteInstantiation.h
```diff
@@ -57,7 +57,7 @@ namespace PluginHost {
             const uint8_t threads,
             const int8_t priority,
             const string configuration,
-            const string environments) = 0;
+            const std::vector<string>& environments) = 0;
```
pwielders: As this is a real COMRPC interface (used for "off host" instantiation, we must make this interface simialr to this: https://github.com/rdkcentral/ThunderInterfaces/blob/master/interfaces/IWifiControl.h#L61-L68
So suggest to add a struct to RPC::Object class (within that scope) like:
```
enum Type : uint8_t {
    LOCAL = 0,
    GLOBAL = 1
};

struct Environment {
    string key;
    string value;
    Type type;
}
```

and than here you could state:

```
using IEnvironmentIterator = RPC::IIteratorType<RPC::Object::Environment, ID_ENVIRONMENT_ITERATOR>;

virtual uint32_t Instantiate(
              ...
              IEnvironmentIterator*& environments /* @out */);
```

Realizing this, I also suggest to only do the substitution of the %...% only after this point so that the "host" reciever will fill in its own substitutions. So suggest to do the substitute here: https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L2416

Source/plugins/IShell.h
pwielders: As described above, if we carry this information in the RPC::Object (Here: https://github.com/rdkcentral/Thunder/blob/master/Source/com/Communicator.h#L46 ) I guess there is no need to "substitute" here. It can be done here: https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L2416 as the environment variables are now part of the RPC::Object And no need to expose such functionality on IShell (also for off-host we do not want to substitute them anyway)

Source/Thunder/PluginServer.h
```diff
@@ -1277,16 +1279,33 @@ namespace PluginHost {
...
                 return (_administrator.Configuration().Substitute(input, PluginHost::Service::Configuration()));
             }
-           Core::hresult Metadata(string& info /* @out */) const override {
+            Core::hresult Metadata(string& info /* @out */) const override {
                 Metadata::Service result;
                 GetMetadata(result);
                 result.ToString(info);
                 return (Core::ERROR_NONE);
             }
+            std::vector<RPC::Object::Environment> SubstituteList(const std::vector<RPC::Object::Environment>& environments) const {
```
pwielders: Why do we want to create a "new" list. Can we not just reuse the old list ? It ssaves a lot of memory and if we call this the assumption is that there is no need to "save" the old one..

Source/com/Communicator.h
```diff
@@ -228,6 +246,13 @@ namespace RPC {
         {
             return (_configuration);
         }
+        inline const std::vector<Environment>& Environments() const
+        {
+            return (_environments);
+        }
+        inline void Environments(const std::vector<Environment>& environments) {
```
pwielders, line 253: Maybe make this a moveable ? Or also allow for a movebale Environments ?

Source/plugins/Configuration.h
```diff
@@ -34,6 +34,106 @@ namespace Plugin {
...
+                            env.overriding = index.Current().Override.Value();
+                            environmentList.push_back(env);
+                        } else {
+                            SYSLOG(Logging::Startup, (_T("Failure in Substituting Value of Key:Value:[%s]:[%s]\n"), index.Current().Key.Value().c_str(), index.Current().Value.Value().c_str()));
+                        }
+                    }
+                }
+                return environmentList;
+            }
+            static RPC::Object::Environment Info(const string& env, RPC::Object::Environment::Scope overriding)
```
pwielders, line 108: This seems to be a  parser of a parameter received (env) to an RPC::Object::Environment, why could this not be a constructor of the  RPC::Object::Environment class?
Maybe an idea to also creat the string from the parameter there (in the RPC::Object::Environment) as well. A ToStringMethod, that concattenaats the keyand value, and escaped the '" ?

Source/plugins/IShell.h
HaseenaSainul: done

Source/Thunder/PluginServer.h
HaseenaSainul: removed

Source/com/Communicator.h
HaseenaSainul: done

Source/plugins/Configuration.h
HaseenaSainul: updated

Source/Thunder/IRemoteInstantiation.h
HaseenaSainul: done

Source/com/Communicator.h
HaseenaSainul: addressed

Source/Thunder/PluginServer.h
HaseenaSainul: addressed

pwielders (approved):

PR #1771, opened 2024-10-08
[ProcessContainers] Update Dobby implementation

sebaszm:


Source/extensions/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -115,7 +111,7 @@ namespace ProcessContainers {
 
                     // Dobby stop is async - block until we get the notification the container
                     // has actually stopped
-                    this->InternalLock();
+                    // this->InternalLock();
```
goruklu: can you please remove this line if it's not needed?

Source/extensions/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
```diff
@@ -139,7 +135,7 @@ namespace ProcessContainers {
                         }
                     }
 
-                    this->InternalUnlock();
+                    // this->InternalUnlock();
```
goruklu: remove this line as well

goruklu (changes requested):

Source/extensions/processcontainers/implementations/DobbyImplementation/DobbyImplementation.cpp
sebaszm: InternalLock() was used to protect the internal list of created containers, which is now handled outside of container provider implementations (since there can be multiple providers). Therefore the InternalLock() method has been removed altogether. 

In this context there container list was not modified, so using this lock was not required. But, perhaps, it was reused for a different purpose, so it was only commented out - as a remark, because it may need reviewing whether a critical section is indeed needed here.

I will remove this, since the original purpose of InternalLock is gone.

goruklu (approved):

MFransen69 (approved):

PR #1784, opened 2024-10-25
[Core/WebSocketLink] : Enable unsollicited 'Pong' and rename enum item

msieben:
- Rename 'WEBSERVER' to 'WEBSERVICE' as both client and server utilize state information

- Unsollicited 'Pong' may act as a heart beat


pwielders:
@msieben the enum is a public interface. You *can not* change it in one go. Support both, deprecate the old one and in a next Major Release we can drop it. This should *not* go in 1 go. Code might depend on this enum value.
Please also refer to a ticket this is solving..

pwielders (changes requested):
Keep on supporting the "old enum" value, add the new enum value and use the DEPRECATE keyword on the old enum value so code will raise a warning that the enum needs to be changed. In the next Major Release we can drop it than.

pwielders (approved):

PR #1794, opened 2024-11-21
Development/websocket

msieben:

PR #1795, opened 2024-11-22
Development/secure socket port

pwielders:
Sofar Thunder supported SecureSocket clients where the intiative to enable secure connections was only based on creating a socket as a client. Tests are being written that require a secure socket server. This requires a sifferent set of calls and revealed an issue in the SocketPort.

The server accepts the incoming socket. The bug was that the incoming socket, as it is already open and connected would not call the "Initialize()" method before it was added to the Resource Monitor. This has been fixed in SocketPort.cpp.

Than this new incoming connection must follow a different path in the interaction with the
opensll library. Already added new states like, ACCEPTING and CONNECTING, in stead of EXCHANGE.

However, testing it with a server build using the Thunder functionality still fails. It is
probably due to the functionality/initialization that has to take place in the Initialize()
method of the SecureScketPort where the SSL context is being created. The differentiation
between a client socket and a server socket is already added to the Initialize, I guess it
has to be tweaked.

The client functionality of the SecureSocketPort has been tested with a server on the internet
and is still working oke, so Client functionality has not been broken with these changes!

### DIFF
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -496,9 +496,15 @@ namespace Thunder {
             m_SendOffset = 0;
 
             if ((m_State.load(Core::memory_order::memory_order_relaxed) & (SocketPort::LINK | SocketPort::OPEN | SocketPort::MONITOR)) == (SocketPort::LINK | SocketPort::OPEN)) {
-                // Open up an accepted socket, but not yet added to the monitor.
-                m_State.fetch_or(SocketPort::UPDATE, Core::memory_order::memory_order_relaxed);
-                nStatus = Core::ERROR_NONE;
+
+                if (Initialize() != Core::ERROR_NONE) {
+                    nStatus = Core::ERROR_ABORTED;
+                }
+                else {
+                    // Open up an accepted socket, but not yet added to the monitor.
+                    m_State.fetch_or(SocketPort::UPDATE, Core::memory_order::memory_order_relaxed);
+                    nStatus = Core::ERROR_INPROGRESS;
+                }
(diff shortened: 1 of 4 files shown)


pwielders:
Sofar Thunder supported SecureSocket clients where the intiative to enable secure connections was only based on creating a socket as a client. Tests are being written that require a secure socket server. This requires a sifferent set of calls and revealed an issue in the SocketPort.

The server accepts the incoming socket. The bug was that the incoming socket, as it is already open and connected would not call the "Initialize()" method before it was added to the Resource Monitor. This has been fixed in SocketPort.cpp.

Than this new incoming connection *must* follow a different path in the interaction with the
opensll library. Already added new states like, ACCEPTING and CONNECTING, in stead of EXCHANGE.

However, testing it with a server build using the Thunder functionality still fails. It is
probably due to the functionality/initialization that has to take place in the Initialize()
method of the SecureScketPort where the SSL context is being created. The differentiation
between a client socket and a server socket is already added to the Initialize, I guess it
has to be tweaked.

The client functionality of the SecureSocketPort has been tested with a server on the internet
and is still working oke, so Client functionality has *not* been broken with these changes!

MFransen69 (approved):

msieben:
@Pierre, per my understanding your assumption about server vs client is correct as already discussed with @MFransen69 . I am in the process of adding it as a sequel of https://github.com/rdkcentral/Thunder/pull/1794 though much of the changes there are as I interpret indeed related to the missing trigger for initialisation.  I assume work continues here and the draft PR for review purpose (only) can be closed.

PR #1796, opened 2024-11-22
Development/websocket

msieben:


Source/cryptalgo/SecureSocketPort.h
```diff
@@ -72,19 +72,26 @@ namespace Crypto {
             };
 
         public:
+            // Type of underlying socket
```
pwielders: This enum, together with the enum class context_t split up into server and client do not add any functionality. They are currently used to "determine" the initial state of the _handshaking in the handler but the handler, in the initialize will override the set state, by definition, through the IsOpen() state, which indeed is the only valid way to determine if you are a server, so why all this additional code?
I suggest setting the initial state at ERROR and if during normal runtime you use this state and it is still ERROR, assert where you would not expect it..

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -72,19 +72,26 @@ namespace Crypto {
...
+            Handler(SecureSocketPort& parent, sockettype_t type, Args&&... args)
                 : Core::SocketPort(args...)
                 , _parent(parent)
                 , _context(nullptr)
                 , _ssl(nullptr)
                 , _callback(nullptr)
-                , _handShaking(CONNECTING) {
-            }
+                , _handShaking(type == sockettype_t::CLIENT_SOCKET ? CONNECTING : ACCEPTING)
+                , _type{type}
```
pwielders: As the sockettype_t type -> _type is not used anwhere, just drop it, it just confuses..

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -72,19 +72,26 @@ namespace Crypto {
...
-            Handler(SecureSocketPort& parent, Args&&... args)
+            Handler(SecureSocketPort& parent, sockettype_t type, Args&&... args)
                 : Core::SocketPort(args...)
                 , _parent(parent)
                 , _context(nullptr)
                 , _ssl(nullptr)
                 , _callback(nullptr)
-                , _handShaking(CONNECTING) {
-            }
+                , _handShaking(type == sockettype_t::CLIENT_SOCKET ? CONNECTING : ACCEPTING)
```
pwielders: _handShaking is properly initialized during the Initialize() call t CONNECTING or ACCEPTING, based on the state of the socket if it is being called. Suggest to set it to ERROR here to assert in the INITIALIZE if, for example, it does not have this vale if we "Initialize()". Making sure the Initialize is called only once..

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -72,19 +72,26 @@ namespace Crypto {
...
         public:
+            // Type of underlying socket
+            enum class sockettype_t : uint8_t {
+                   CLIENT_SOCKET
+                ,  SERVER_SOCKET
+            };
+
             Handler(Handler&&) = delete;
             Handler(const Handler&) = delete;
             Handler& operator=(const Handler&) = delete;
```
pwielders, line 128: Please add the:
Handler& operator=(Handler&&) = delete;
for upgrading code we should add it now...

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -107,6 +114,7 @@ namespace Crypto {
 
             // Signal a state change, Opened, Closed or Accepted
             void StateChange() override {
+//                ASSERT(_context != nullptr && _ssl != nullptr);
```
pwielders: Interesting that this ASSERT fires... Maybe during the Close Sequence? Please investigate and remove the assert or make sure that under normal conditions it does not fire (IsClosed() == false) || ((_context != nullptr) && (_ssl != nullptr)));

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -135,17 +143,31 @@ namespace Crypto {
             void* _ssl;
             IValidator* _callback;
             mutable state _handShaking;
+            const sockettype_t _type;
```
pwielders: As described above, I think we can drop this..

Source/cryptalgo/SecureSocketPort.h
```diff
@@ -135,17 +143,31 @@ namespace Crypto {
...
         };
 
     public:
         SecureSocketPort(SecureSocketPort&&) = delete;
         SecureSocketPort(const SecureSocketPort&) = delete;
         SecureSocketPort& operator=(const SecureSocketPort&) = delete;
 
+    protected:
```
pwielders: from here till 166, I think you can drop it has no added value..

Source/cryptalgo/SecureSocketPort.cpp
```diff
@@ -130,43 +130,60 @@ uint32_t SecureSocketPort::Handler::Initialize() {
...
-        TRACE_L1("OpenSSL failed to load certificate store");
+    constexpr uint64_t options = SSL_OP_ALL | SSL_OP_NO_SSLv2;
+
+    VARIABLE_IS_NOT_USED uint64_t bitmask = SSL_CTX_set_options(static_cast<SSL_CTX*>(_context), options);
+
+    ASSERT((bitmask & options) == options);
+
+    if ( ( (    (_handShaking == ACCEPTING)
+              // Load server certifiate and private key
+             && (SSL_CTX_load_verify_locations(static_cast<SSL_CTX*>(_context), "RootCA.pem", nullptr) == 1)
```
pwielders: The missing code in case we are a server :-) I would assume that these three files, "RootCA.pem", "localhost.pem" and "localhost.key" can be made a parameter and I guess/hope we can make the loading of these certificates part of the constructor (of the server) to validate their "loadability" and if not loadable, just stop the server from opening up..

Source/cryptalgo/SecureSocketPort.cpp
```diff
@@ -230,34 +262,44 @@ void SecureSocketPort::Handler::Update() {
...
+                _handShaking = CONNECTED;
+                // If server has sent a certificate, and, we want to do 'our' own check
+                ValidateHandShake();
             }
-        }
+        } // Server
         else if (_handShaking == ACCEPTING) {
             if ((result = SSL_accept(static_cast<SSL*>(_ssl))) == 1) {
-                _handShaking = EXCHANGE;
+                _handShaking = CONNECTED;
```
pwielders: I think the ValidateHandshake will set the _handShaking correctly based on the outcome, so no real need to set it if you call the ValidateHandShake() method, it will become CONNECTED or ERROR I assume.

Source/cryptalgo/SecureSocketPort.cpp
```diff
@@ -230,34 +262,44 @@ void SecureSocketPort::Handler::Update() {
 
         ASSERT(_ssl != nullptr);
 
+        // Client
         if (_handShaking == CONNECTING) {
             if ((result = SSL_connect(static_cast<SSL*>(_ssl))) == 1) {
-                _handShaking = EXCHANGE;
+                _handShaking = CONNECTED;
```
pwielders: I think the ValidateHandshake will set the _handShaking correctly based on the outcome, so no real need to set it if you call the ValidateHandShake() method, it will become CONNECTED or ERROR I assume.

Source/cryptalgo/SecureSocketPort.cpp
```diff
@@ -230,34 +262,44 @@ void SecureSocketPort::Handler::Update() {
...
+                // Client has sent a certificate (optional, on request only)
+                ValidateHandShake();
             }
-        }
-        
+        } // Re-initialie a previous session
         if (_handShaking == EXCHANGE) {
             if ((result = SSL_do_handshake(static_cast<SSL*>(_ssl))) == 1) {
                 ValidateHandShake();
+                _handShaking = CONNECTED;
```
pwielders: ??? Shouldn't the ValidateHandshake method not determine this? What if the HandShake failed ?

Source/cryptalgo/SecureSocketPort.cpp
```diff
@@ -230,34 +262,44 @@ void SecureSocketPort::Handler::Update() {
...
 
         if (result != 1) {
             result = SSL_get_error(static_cast<SSL*>(_ssl), result);
-            if ((result != SSL_ERROR_WANT_READ) && (result != SSL_ERROR_WANT_WRITE)) {
-                _handShaking = ERROR;
-            }
-            else if (result == SSL_ERROR_WANT_WRITE) {
+
+            if ((result == SSL_ERROR_WANT_READ) || (result == SSL_ERROR_WANT_WRITE)) {
+                // Non-blocking I/O: select may be used to check if condition has changed
```
pwielders: If SSL still needs another READ or WRITE to complete the Handshaking, I assume t has *no* use to ValidateHandShake () and just wait for more data and "stay" in the EXCHANGE state...
I can not imagine that if the Error of SSL is, I need more data to write or read that the handshaking is over and thus that you can validate the handshake and even worse that you can move the connection to CONNECTED as I assume the exchange for TLS is still in progress..

pwielders (changes requested):
See comments!

msieben:
@Pierre, again thanks for the elaborated feedback.  Did you notice it is in draft status and possibly still WIP? Interestingly, some feedback is on your 'own' code that was copied and pasted. I will try to provide some feedback if appropriate on each of the sections.

Source/cryptalgo/SecureSocketPort.h
msieben: Good observation. It was not the intended goal to have it overwritten.

Source/cryptalgo/SecureSocketPort.h
msieben: It was not the intended goal as mentioned here https://github.com/rdkcentral/Thunder/pull/1796#discussion_r1858014895

Source/cryptalgo/SecureSocketPort.h
msieben: This is interesting. In a previous draft PR you did not wanted such solution but promoted the use of ASSERT.  Update() already includes a check and ASSERT.

Source/cryptalgo/SecureSocketPort.cpp
msieben: Indeed.  Setting those is still missing in the API.

Source/cryptalgo/SecureSocketPort.cpp
msieben: ValidateHandshake returns void but overrides and thus resets the value internally to error. Should we transfer the responsibility of connected state to the method? What happens if we add additional and start using functionality (later) which does not fit the validation of certificates?

Source/cryptalgo/SecureSocketPort.cpp
msieben: See https://github.com/rdkcentral/Thunder/pull/1796#discussion_r1858044490

Source/cryptalgo/SecureSocketPort.cpp
msieben: See https://github.com/rdkcentral/Thunder/pull/1796#discussion_r1858044490. Although the order is unintentionally reversed here.

Source/cryptalgo/SecureSocketPort.cpp
msieben: This is puzzling item I was planning to discuss as the API documentation states 'When using a nonblocking socket, nothing is to be done, but select() can be used to check for the required condition.' So it suggests checking the descriptor to become ready. Perhaps it can be implemented here, but we should avoid an indefinite block though unlikely.

Source/cryptalgo/SecureSocketPort.h
msieben: Not the state of the socket as in open or closed. The intent is to differentiate between and identify the owner of the socket: a server using the accept method to construct it, or, a client that creates it to connect to a server, Handler::ACCEPTING, Handler::CONNECTING respectively.

Source/cryptalgo/SecureSocketPort.h
msieben: Are you sure? _handShaking is initially used to identify the server or client side of the connection, but will loose that information once connected or in error.

mhughesacn:
Hi: Please comment on the key and pem files.  Are these all just test values with nothing sensitive?  If so, please add a README file to the same directory explaining this.
Thank you!

msieben:
@mhughesacn  These files are part of the unit test suite. They are BASE64 (PEM) encoded self-signed certificates and private keys used only with test_websocket.cpp to set up an SSL/TLS encrypted communication channel on localhost. Certificate fields have fictional values like ST=ThunderRealm, L= ThunderServer, O=ThunderCertificates etc as can be extracted for further investigation using tools like openssl.

mhughesacn:
Thank you for the clarification.  But please will you note somewhere (a new README or at top level if there is a suitable file)  that all key values in this directory contain only test data.  If Comcast Security runs a scan here and complains, we can point them to this to defuse any issue.

msieben:
@mhughesacn See https://github.com/rdkcentral/Thunder/blob/e5d430463fa94e3ccd78589d8d678b5eb88278fb/Tests/README.

Source/core/SocketPort.cpp
```diff
@@ -1144,11 +1144,11 @@ namespace Thunder {
         }
 
         /* virtual */ int32_t SocketPort::Read(uint8_t buffer[], const uint16_t length) const {
-            return (::recv(m_Socket, reinterpret_cast<char*>(buffer), length, 0));
+            return (::recv(m_Socket, reinterpret_cast<char*>(buffer), length, MSG_NOSIGNAL));
```
pwielders: This should be thoroughly tested on production systems or made an option to enable it on a created SocketPort as this might cause regression. 
I wonder if this is related to the investgation @sebaszm did with the closedown of the socket. @sebaszm to comment?
I guess this might change the flow of action in closing down "failing/closing" sockets...

sebaszm: I don't think MSG_NOSIGNAL is valid on recv() (so this has a chance to fail with EINVAL).

msieben: @pwielders, @sebaszm, good observation. I have checked and the flag has not been defined in that context.

pwielders (changes requested):
@msieben tried reviewing the Thunder related code. What I noticed is a degradation of readability (void* are a no go area, what was wrong with forwarded Struct pointer definition). The assignment was only to add test cases for the WebSockets. You have extended this to also include SecureWebSockets, which is a good addition.
Extending the testcases to the SecureWebSockets showed an omission in functionality of Client Side Certificate validation. A feature *NOT* used today in Thunder but useful to have.
Could you please focus on *only* adding this feature in stead of refactoring the whole code, which makes it *very* hard for the reviewers to focus on the actual change.

PR #1799, opened 2024-11-27
Allow jsonrpc for plugin suspend/resume

nxtum:


Source/Thunder/Controller.cpp
```diff
@@ -1010,6 +1009,7 @@ namespace Plugin {
                 }
                 else {
                     result = stateControl->Request(PluginHost::IStateControl::command::RESUME);
+                    NotifySuspendResumeStateChange(callsign, Exchange::Controller::ILifeTime::state::RESUMED);
```
pwielders: I think this is too early, we can only do so if the state is indeed changed. You know that if you would subscribe to the IStateControl::INotification interface. What if the Suspend or Resume fails ?
Lets have a broader discussion on this if we want this as this has some severe impact..

Source/Thunder/Controller.cpp
```diff
@@ -979,6 +978,7 @@ namespace Plugin {
                 }
                 else {
                     result = stateControl->Request(PluginHost::IStateControl::command::SUSPEND);
+                    NotifySuspendResumeStateChange(callsign, Exchange::Controller::ILifeTime::state::SUSPENDED);
```
pwielders: I think this is too early, we can only do so if the state is indeed changed. You know that if you would subscribe to the IStateControl::INotification interface. What if the Suspend or Resume fails ?
Lets have a broader discussion on this if we want this as this has some severe impact..

pwielders (changes requested):
As discussed in the call. The issue is that we need to "wait" for the statechange to actually occure (moving to SUSPENDED/RESUMED) and not already issue this at the moment it is being requested as at that time the plugin did not yet "reach" that state.
To accomodate that we discussed putting the subscribng on the IStateControl::INotification in the PluginHost::Server::Service::AcquireInterface(https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L1479) and the releasing of this interface on the PluginHost::Server::Service::ReleaseInterface(https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L1563). Than hooping through the _administrator in this class (https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L1649) you can get to the Controller to send the statechanges for each plugin to subscribers of the controller.

Note that you probably do not need to create a SinkType<> for the Service to receive the notifications as you can just "extend" the Composit (https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L387) already a member here as _composit (https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L1650) with the IStateControl::INotification interface to get the required changes!

sebaszm (dismissed):
Looks good! 
I do not myself have a better name for StateControlStateChange.

Source/Thunder/PluginServer.cpp
```diff
@@ -1224,6 +1227,30 @@ namespace PluginHost {
...
+        ASSERT(controller != nullptr);
+
+        switch (state) {
+            case IStateControl::state::SUSPENDED:
+                controller->NotifyStateControlStateChange(callsign, Exchange::Controller::ILifeTime::state::SUSPENDED);
+                break;
+            case IStateControl::state::RESUMED:
+                controller->NotifyStateControlStateChange(callsign, Exchange::Controller::ILifeTime::state::RESUMED);
+                break;
+            case IStateControl::state::EXITED:
```
pwielders: Just checked the EXITED state, since I did not understaood what it meant, and it looks like it is not used at all. Could we try to remove it (from interface and here) and see if it still all compiles ?

sebaszm: It is used, if the plugin/application itself requests to quit.

sebaszm (approved):

pwielders (approved):

PR #1805, opened 2024-12-05
[Core] Ensure inline toupper/tolower

sebaszm:
Some standard lib headers (e.g. stdlib.h) end up setting __NO_CTYPE. When ctype.h is processed with this flag different code is compiled, amongst others making toupper/tolower and some isxxx() not inline but calling extern functions (which is slower). Because of include guards to ensure the inlining, ctype.h must be the first header file included for a module.

### DIFF
--- a/Source/core/Portability.h
+++ b/Source/core/Portability.h
@@ -396,6 +396,8 @@ struct SEMAPHORE_BASIC_INFORMATION {
 #endif
 
 #ifdef __LINUX__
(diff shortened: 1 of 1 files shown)


sebaszm:
This behaviour is a potential libstdc++ bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100128

pwielders (approved):

PR #1808, opened 2024-12-11
Facilitate automatic JSON-RPC object lookup

sebaszm:


pwielders (commented):
What a wonderfull code generation are we getting with theseautomatic lookups ! Love it !

pwielders (approved):

PR #1819, opened 2025-01-06
Smartlink crash fix

Karthick-Somasundaresan:


Source/websocket/JSONRPCLink.h
```diff
@@ -460,14 +460,67 @@ namespace Thunder {
...
             using InvokeFunction = Core::JSONRPC::InvokeFunction;
 
+            class Handler {
+                public:
+                using HandlerMap = std::unordered_map<string, InvokeFunction>;
+
+                public:
+                    Handler():_adminLock(),_invokeMap(){}
+                    Handler(const Handler&) = delete;
+                    Handler& operator=(const Handler&) = delete;
```
pwielders, line 471: Also delete the move  constructor and the move assignment operator (latest coding standards)

Source/websocket/JSONRPCLink.h
```diff
@@ -1174,6 +1225,21 @@ namespace Thunder {
             class Connection : public LinkType<INTERFACE> {
             private:
                 using Base = LinkType<INTERFACE>;
+                class EventSubscriber: public Core::Thread {
```
pwielders, line 1230: Do we really need a thread here ? This will always (per connection) require an 8MB callstack. I thnk you can "reuse" the ResouceMonitor Thread (communication thread) as this is probably just to request a registration ? Or if we need a thread, could we do something to limit it to 1 thread for *all* Conenction types per process ?

Source/websocket/JSONRPCLink.h
```diff
@@ -1261,30 +1330,77 @@ namespace Thunder {
...
+                return Base::Unsubscribe(waitTime, eventName);
+            }
+
 
             private:
+                uint32_t SendSubscribeRequest(const string& eventName) {
+                    uint32_t retVal = Core::ERROR_UNAVAILABLE;
+                    Core::JSONRPC::Message response;
+                    const string parameters("{ \"event\": \"" + eventName + "\", \"id\": \"" + Base::Namespace() + "\"}");
+                    auto result = Base::template Invoke<string>(DefaultWaitTime, "register", parameters, response);
```
pwielders, line 1392: Oke, synchronous Invoke, so communication thread is not an option :-) Maybe than 1 thread for all processes that have this ?

pwielders (approved):

PR #1822, opened 2025-01-07
[Messaging] ASSERTS as a new Messaging stream

VeithMetro:
The idea behind these changes has been described more in [this](https://jira.rdkcentral.com/jira/browse/METROL-1076) Jira ticket.

After this PR is merged, the second one to ThunderNanoServices will follow, which adjusts MessageControl accordingly. And then the same will apply to ThunderUI and other repos with some minor adaptations.


Source/core/MessageStore.cpp
```diff
@@ -390,6 +407,87 @@ namespace Core {
...
         }
 
+        uint16_t IStore::Assert::Serialize(uint8_t buffer[], const uint16_t bufferSize) const
+        {
+            uint16_t length = MessageInfo::Serialize(buffer, bufferSize);
+
+            if (length != 0) {
+                const uint16_t extra = static_cast<uint16_t>(sizeof(_processId) + (_processName.size() + 1) + (_fileName.size() + 1) + sizeof(_lineNumber) + (_callstack.size() + 1));
+
+                ASSERT(bufferSize >= (length + extra));
```
sebaszm: Wondering what happens when ASSERT() happens during processing of an assert?

VeithMetro: Depends, if we are talking about Debug and having abort in the ASSERT macro, then I suppose we would only get this ASSERT in the buffer and not the "original" one, since we would abort right after the ```bufferSize >= (length + extra)``` ASSERT, so before the original one can be put into the buffer

sebaszm (changes requested):
It seems to me that ASSERTs on serialize path have non-zero chance of endless recursion and need to be removed (maybe replaced with TRACE_L1 for some sign of the double fault happening) and if needed the assert situation guarded by a condition statement instead.

VeithMetro:
> It seems to me that ASSERTs on serialize path have non-zero chance of endless recursion and need to be removed (maybe replaced with TRACE_L1 for some sign of the double fault happening) and if needed the assert situation guarded by a condition statement instead.

@sebaszm correct, the chance of an endless recursion was definitely not zero - if an another ASSERT happened before the original ASSERT was fully serialized, then we could end up with an endless recursion..

To handle that, I replaced each ASSERT on the messages serialization path with a new macro `INTERNAL_ASSERT`, which, like we discussed on Skype, prints the ASSERT directly onto the console

sebaszm (approved):

Source/core/AssertionControl.cpp
```diff
@@ -0,0 +1,76 @@
...
+            {
+                ANNOUNCE_ASSERT_CONTROL
+            }
+        } ControlRegistration;
+
+    }
+
+    AssertionUnitProxy::AssertionUnitProxy()
+        : _handler(nullptr)
+        , _adminLock(new Core::CriticalSection())
```
pwielders, line 44: Why can this not be a member (composit), Is there a real purpose to the new ?

pwielders: Never mind found the answer, forward declaration.

pwielders (approved):

PR #1824, opened 2025-01-10
[test] trigger coverity

MFransen69:


MFransen69:
Was just for testing coverity. Test passed :)

PR #1825, opened 2025-01-13
Development/documentation

msieben:


docs/rfc/websocket.md
```diff
@@ -0,0 +1,537 @@
...
+    Note over Client,Server: No connection
+
+    loop
+        critical Establish connection
+            alt
+                Client->>Server: Connect to port (default 80)
+            else
+                Client->>Server: Connect to port (default 443)
+
+                Note over Client,Server: TLS handshake
```
MFransen69, line 26: As TLS in principle is not connected to the WebSocket protocol itself I think we either remove that part or otherwise make that clear in the title and add some text to explain that?

MFransen69 (changes requested):

docs/rfc/websocket.md
msieben: It denotes an alternative path as the diagram illustrates:  WebSocket is on top of existing infrastructure, regular or TLS tunnelled  (see https://www.rfc-editor.org/rfc/rfc6455.html#page-4 and https://www.rfc-editor.org/rfc/rfc6455.html#page-11). The diagram does not explain TLS in detail. Solid line delineated (blocks) text is  to be interpreted as comment. Actually, the 'initial part of' the protocol starts with sending an HTTP upgrade request.

MFransen69 (approved):

PR #1841, opened 2025-02-19
[Messaging] Add a new DIRECT_SYSLOG macro

VeithMetro:
It should only be used when the messaging engine is not running yet

### DIFF
--- a/Source/Thunder/Config.h
+++ b/Source/Thunder/Config.h
@@ -691,7 +691,7 @@ namespace PluginHost {
                 _ethernetCard = config.EthernetCard.Value();
                 _delegatedReleases = config.DelegatedReleases.Value();
                 if( config.Latitude.IsSet() || config.Longitude.IsSet() ) {
-                    SYSLOG(Logging::Error, (_T("Support for Latitude and Longitude moved from Thunder configuration to plugin providing ILocation support")));
+                    DIRECT_SYSLOG("Support for Latitude and Longitude moved from Thunder configuration to plugin providing ILocation support\n");
                 }
 
                 _messagingCategoriesFile = config.DefaultMessagingCategories.IsQuoted();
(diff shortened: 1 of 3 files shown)


pwielders:
@VeithMetro does this mean that the use of DIRECT_SYSLOG moves the logging from the Messaging mechanism to the default printf again?
If this is to get messages that might be lost due to the fact that the MessageControl plugin never starts, I think we should reconsider that :-)
The data is stored (in the MMF files) it is currently just not being published. I guess we could do 2 options:
1) Have an application that flushes the data in the MMF (MemoryMapped File) as the plugin would do. And you can run this app after a crash of Thunder application.
2) Integrate this app and in the @exit or just before close down of Thunder, forcefully flush all information found in the MMF.
The App or the additional functionality could be as simple as just "opening" the MMF's ( no need for the doorbell and all the other stuff) and just spit out all info .

Also wonder why the -F (in such error cases) would not be sufficient. If I am not mistake with the -f (or -F) option being passed to Thunder each and every Messageing LINE is also send to the printf at the moment it is being executed and thus does not end up in the MMF. This should allow you to find issues if they occure before the MessageCOntrol plugin is activated..
Maybe a discussion on skype?

VeithMetro:
> @VeithMetro does this mean that the use of DIRECT_SYSLOG moves the logging from the Messaging mechanism to the default printf again? If this is to get messages that might be lost due to the fact that the MessageControl plugin never starts, I think we should reconsider that :-) The data is stored (in the MMF files) it is currently just not being published. I guess we could do 2 options:
> 
>     1. Have an application that flushes the data in the MMF (MemoryMapped File) as the plugin would do. And you can run this app after a crash of Thunder application.
> 
>     2. Integrate this app and in the @EXIT or just before close down of Thunder, forcefully flush all information found in the MMF.
>        The App or the additional functionality could be as simple as just "opening" the MMF's ( no need for the doorbell and all the other stuff) and just spit out all info .
> 
> 
> Also wonder why the -F (in such error cases) would not be sufficient. If I am not mistake with the -f (or -F) option being passed to Thunder each and every Messageing LINE is also send to the printf at the moment it is being executed and thus does not end up in the MMF. This should allow you to find issues if they occure before the MessageCOntrol plugin is activated.. Maybe a discussion on skype?

@pwielders not exactly, it is not about MessageControl, it is about messaging itself (creating MessageUnit).

[Here](https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginHost.cpp#L649) in main() in PluginHost.cpp we need config to create the instances of MessageUnit (which leads to the creation of MMFs), AssertionUnit and WarningReportingUnit. That's why we were not able to see any SYSLOGs or ASSERTs from Config, which happens prior to that in main, and that's why we decided to create this new `DIRECT_SYSLOG` macro, which should only be used in this very specific situation (before messaging is up)

And as a side note, we already have [this](https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/blob/master/MessageControl/BuffersFlush/main.cpp) application to flush the MMFs - I created it with the latest "ASSERTs via messaging" changes 😄

PR #1843, opened 2025-02-21
[Messaging] No message types should be missed even before the messaging is fully initialized

VeithMetro:
Question: do we also want to make sure the same applies to `WarningReporting`, or by design it should not be used prior to the `messaging engine` being up and running (in [this](https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginHost.cpp#L649) line of `main()` - do not confuse with the `MessageControl plugin`, as it can start way later and no messages should ever be lost, since they will be waiting in the buffer)?

### DIFF
--- a/Source/core/AssertionControl.cpp
+++ b/Source/core/AssertionControl.cpp
@@ -68,7 +68,8 @@ namespace Assertion {
             _handler->AssertionEvent(metadata, message);
         }
         else {
-            fprintf(stderr, "%s%s\n", metadata.ToString(Core::Messaging::MessageInfo::abbreviate::FULL).c_str(), message.Data().c_str());
+            ::fprintf(stderr, "%s%s\n", metadata.ToString(Core::Messaging::MessageInfo::abbreviate::FULL).c_str(), message.Data().c_str());
+            ::fflush(stderr);
         }
 
         _adminLock->Unlock();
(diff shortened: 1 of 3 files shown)


sebaszm (dismissed):

sebaszm (dismissed):

Source/core/AssertionControl.cpp
```diff
@@ -60,10 +60,18 @@ namespace Assertion {
...
+        metadata.TimeStamp(Thunder::Core::Time::Now().Ticks());
+
+        // print the ASSERT to stderr if handler is not set, possibly due to the messaging engine not being fully initialized yet
         if (_handler != nullptr) {
             _handler->AssertionEvent(metadata, message);
+            _adminLock->Unlock();
+        }
+        else {
+            _adminLock->Unlock();
+            fprintf(stderr, "%s%s\n", metadata.ToString(Core::Messaging::MessageInfo::abbreviate::FULL).c_str(), message.Data().c_str());
```
pwielders: Don't we also want to ::fflush(stderr); here just to make sure that f we are in a very early crashing situation that at least this line is outputted ?

pwielders (changes requested):
One minor but for people who rely o the output, maybe a crucial small change :-)

pwielders:
> Question: do we also want to make sure the same applies to `WarningReporting`, or by design it should not be used prior to the `messaging engine` being up and running (in [this](https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginHost.cpp#L649) line of `main()` - do not confuse with the `MessageControl plugin`, as it can start way later and no messages should ever be lost, since they will be waiting in the buffer)?

Good question, it should probably not happen but....

pwielders:
oops accidentally closed it bu reopened

Source/core/AssertionControl.cpp
VeithMetro: Good remark! Added the flush

pwielders (approved):

PR #1852, opened 2025-03-04
[Messaging] Make IStore a member of the Controls class instead of a separate static

VeithMetro:
The idea was mentioned in [this](https://jira.rdkcentral.com/jira/browse/METROL-899) Jira ticket

### DIFF
--- a/Source/core/MessageStore.cpp
+++ b/Source/core/MessageStore.cpp
@@ -108,7 +108,7 @@ ENUM_CONVERSION_END(Core::Messaging::Metadata::type)
                 _adminLock.Unlock();
             }
 
-            void SetStorage(Core::Messaging::IStore* storage)
+            void Set(Core::Messaging::IStore* storage)
             {
                 _adminLock.Lock();
 
@@ -517,7 +517,7 @@ namespace Core {
         
         /* static */ void IStore::Set(IStore* storage)
         {
-            ControlsInstance().SetStorage(storage);
+            ControlsInstance().Set(storage);
         }
     } // namespace Messaging
 } // namespace Core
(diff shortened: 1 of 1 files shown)


Source/core/MessageStore.cpp
```diff
@@ -102,9 +104,23 @@ ENUM_CONVERSION_END(Core::Messaging::Metadata::type)
                 _adminLock.Unlock();
             }
 
+            void SetStorage(Core::Messaging::IStore* storage)
+            {
+                ASSERT((_storage == nullptr) ^ (storage == nullptr));
```
sebaszm, line 115: Assert needs to be inside the lock.

Source/core/MessageStore.cpp
```diff
@@ -102,9 +104,23 @@ ENUM_CONVERSION_END(Core::Messaging::Metadata::type)
...
+
+                _adminLock.Lock();
+                _storage = storage;
+                _adminLock.Unlock();
+            }
+
+            Core::Messaging::IStore* Storage() const {
+                return (_storage);
+            }
```
sebaszm, line 120: _storage is not properly protected here. 
But it doesn't seem to me that we need to export Storage out of the controls instance. It is only required during Announce which is also on controls? Then the Announce should use it's insternal _storage and call Default and it already can guard the nullptr.

Source/core/MessageStore.cpp
```diff
@@ -490,8 +504,8 @@ namespace Core {
         {
             ControlsInstance().Announce(control);
 
-            if (_storage != nullptr) {
-                control->Enable(_storage->Default(control->Metadata()));
+            if (ControlsInstance().Storage() != nullptr) {
+                control->Enable(ControlsInstance().Storage()->Default(control->Metadata()));
```
sebaszm: Looks odd and is not properly locked. See comment above.

Source/core/MessageStore.cpp
```diff
@@ -504,13 +518,12 @@ namespace Core {
         }
 
         /* static */ IStore* IStore::Instance() {
-            return (_storage);
+            return (ControlsInstance().Storage());
```
sebaszm: Potentially could be removed?

sebaszm (changes requested):

sebaszm (approved):

PR #1854, opened 2025-03-09
JSON Parser - Stop escaping forward slash '/'

nxtum:
METROL-1117

Most tools (python, js multiple json linters ) seem to accept both but not escape themself.


pwielders:
Note: Please check the JSON spec. If I recall correctly is that this is optional but recommended. I think there was an issue in the past when we added this...

MFransen69:
@nxtum just talked to Pierre on his comment, I now do understand where this is coming from.
I'll give you a call and discuss

pwielders (approved):

PR #1865, opened 2025-04-01
[Portability] Add pid_t definition for Windows and replace each process_t with it

VeithMetro:


Source/core/ProcessInfo.h
```diff
@@ -27,7 +27,7 @@ namespace Thunder {
 namespace Core {
 
     // On 64 bits deployments, this is probably a uint64_t, lets prepare for it :-)
-    using process_t = uint32_t;
+    using process_t = pid_t;
```
pwielders: Lets mark it deprecated, so that if process_t is used, at least it raises a warning..

VeithMetro: I also thought about it and already tried to, but adding DEPRECATED in front raises this compile error:
```
In file included from ProcessInfo.cpp:20:
ProcessInfo.h:30:16: error: expected unqualified-id before ‘using’
   30 |     DEPRECATED using process_t = pid_t;
      |                ^~~~~
```
But now I tried it like this:
```
using process_t DEPRECATED = pid_t;
```
And it seems to work 😄

VeithMetro: Update: it does work on Linux, but not on Windows..

pwielders: Who cares about windows 🤣

VeithMetro: I can think of at least one person... 🤣

pwielders (approved):

PR #1874, opened 2025-05-06
Fix for Assert during postmortem crash.

Karthick-Somasundaresan:


pwielders (dismissed):
@Karthick-Somasundaresan Probable Fix? I guess we tested the scenarios discussed in theory and this patch does indeed solve them in Practise :-)

Karthick-Somasundaresan:
> @Karthick-Somasundaresan Probable Fix? I guess we tested the scenarios discussed in theory and this patch does indeed solve them in Practise :-)

Yeah, my bad, got the title wrong. Changed it now.

pwielders (approved):

PR #1876, opened 2025-05-09
[Core] Move JSON-RPC instance ID to string type

sebaszm:
[DependsOn=development/jsongen-use-string-instance-id]


pwielders (approved):
That looks impressive!

PR #1877, opened 2025-05-09
[JSONRPC] Facilitate custom object lookup

sebaszm:


Source/plugins/JSONRPC.h
```diff
@@ -1020,7 +1020,119 @@ namespace PluginHost {
...
+            return (obj);
+        }
+
+    private:
+        using ToIDCalback = std::function<Core::IUnknown*(const string&)>;
+        using FromIDCallback = std::function<uint32_t(const Core::IUnknown*)>;
+        using Translators = std::pair<ToIDCalback, FromIDCallback>;
+
+        mutable Core::CriticalSection _adminLock;
+        std::map<uint32_t, Translators> _translators;
```
pwielders: unordered_map seems to be more efficient here ?

sebaszm: agree

Source/plugins/JSONRPC.h
```diff
@@ -1137,23 +1137,183 @@ namespace PluginHost {
...
+        JSONRPCSupportsObjectLookup(const JSONRPCSupportsObjectLookup&) = delete;
+        JSONRPCSupportsObjectLookup& operator=(const JSONRPCSupportsObjectLookup&) = delete;
+        JSONRPCSupportsObjectLookup(JSONRPCSupportsObjectLookup&&) = delete;
+        JSONRPCSupportsObjectLookup& operator=(JSONRPCSupportsObjectLookup&&) = delete;
+
+        JSONRPCSupportsObjectLookup()
+            : _adminLock()
+        {
+        }
+        virtual ~JSONRPCSupportsObjectLookup()
```
pwielders: I think this should be ```~JSONRPCSupportsObjectLookup() override``` i.s.o. virtual

pwielders (commented):
A minor remark! maybe good to correct before we merge it..

pwielders:
But looks good. Love templates!

Source/plugins/JSONRPC.h
sebaszm: it surely should

pwielders (approved):

PR #1879, opened 2025-05-16
[core] : check error condition for 'Kill'

msieben:
'ASSERT' based return value and 'errno' checks

### DIFF
--- a/Source/core/Process.h
+++ b/Source/core/Process.h
@@ -560,7 +560,9 @@ namespace Core {
                 TerminateProcess(_info.hProcess, 1234);
             }
 #else
-            ::kill(_PID, (hardKill ? SIGKILL : SIGTERM));
+            ASSERT((errno = 0) == 0);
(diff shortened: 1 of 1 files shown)


Source/core/Process.h
```diff
@@ -560,7 +560,9 @@ namespace Core {
                 TerminateProcess(_info.hProcess, 1234);
             }
 #else
-            ::kill(_PID, (hardKill ? SIGKILL : SIGTERM));
+            ASSERT((errno = 0) == 0);
```
MFransen69, line 563: Why is this in here? Is there a very specific reason? Resetting should not be needed as far as I know (and if so now it is only done in debug) and why then check if it worked? In the Linux manual it does not mention errno is involved in Kill, guess because there is already a return value, so why should we check it?
Just remove this line and remove the errno from the assert below?

MFransen69 (changes requested):

MFransen69:
As there has been no activity for some time let's close this (reopen if needed)

PR #1882, opened 2025-05-27
[com] log comrpc timeout reached on invoke to syslog

MFransen69:


Source/com/IUnknown.h
```diff
@@ -325,15 +325,21 @@ namespace ProxyStub {
...
+                result = channel->Invoke(message, waitTime);
 
-	                // Oops something failed on the communication. Report it.
-	                TRACE_L1("IPC method invocation failed for 0x%X, error: %d", message->Parameters().InterfaceId(), result);
-	            }
-	    }
+                if (result != Core::ERROR_NONE) {
+
+                    if (result == Core::ERROR_TIMEDOUT) {
+                        SYSLOG(Logging::Error, (_T("IPC method Invoke failed due to timeout. Execution of code may or may not have happened. Side effects are to be expected after this message")));
```
pwielders: Maybe a good to report, in this error, the Interface Id and the MethodId. This way, if we receive the logs, we know which interface/method failed/ timed out..
https://github.com/rdkcentral/Thunder/blob/7af707041f3a0389f11195d11ba6911e807b2a60/Source/com/Messages.h#L103
and 
https://github.com/rdkcentral/Thunder/blob/7af707041f3a0389f11195d11ba6911e807b2a60/Source/com/Messages.h#L111
through:
https://github.com/rdkcentral/Thunder/blob/master/Source/core/IPCConnector.h#L444

message->Parameters().InterfaceId()
and
message->Parameters().MethodId()

pwielders (changes requested):

Source/com/IUnknown.h
MFransen69: Great idea, let me add that

pwielders (approved):

PR #1908, opened 2025-07-25
Fix/detect dead web socket

Karthick-Somasundaresan:
This is the fix for [METROL-1149](https://jira.rdkcentral.com/jira/browse/METROL-1149)

Root Cause: The WebSocket is considered active when data is sent over it. This causes sending a PING to also mark the client as active, but it should only be marked as active when a PONG is received. 

Fix: This change ensures that only when data is read from the channel it is recognized as genuine activity from the client.


Source/websocket/WebSocketLink.h
```diff
@@ -566,7 +569,7 @@ POP_WARNING()
 
                 _adminLock.Lock();
 
-                _state |= ACTIVITY;
+                _state |= WRITE_ACTIVITY;
```
pwielders: As we know that we are now PINGING, (_state & PINGED) != 0, lets not set the WRITE_ACTIVITY if a ping is issued. This way the PING must forcefully be answered if, for example, we consider write only (STREAMING) sockets.

Source/plugins/Channel.h
```diff
@@ -211,31 +209,14 @@ namespace PluginHost {
...
-            } 
-            else {
-                // We are a websiocket and had no activity, time to send a ping..
-                _state |= PINGED;
-                Unlock();
-
-                const_cast<BaseClass*>(static_cast<const BaseClass*>(this))->Ping();
+            bool result = HasReadActivity();
+            if (result == false && IsWebSocket() == true && IsPingInProgress() == false) {
+                Ping();
```
pwielders: Lets force this Ping functionality, even a layer up... This way also this HasActivity can remain const and wheter or not we send out a Ping() is delegated to the higer layers which might have a configuration to do this ... or not..

pwielders (changes requested):
I was expecting also the SFINEA changes for the Watchdog on the Resource monitor ?

pwielders (approved):

PR #1918, opened 2025-08-08
[Dangling] Improve the offloading of Dangling proxies notification.

sramani-metro:
When Communication thread receives a broken pipe, dangling notification is sent to the CommunicationServer. The notification to ICOMLink::INotification subscribers was already offloaded to a Job in CommunicationServer, but removing IPlugin::INotification subscribers was still done in comm thread. This will make the comm thread wait on the notifier lock when activation/deactivation of any plugin is in progress and the notifier lock for IPlugin::INotification was already taken by a workerpool thread. This change makes sure the comm thread completely offloads the notification to a dedicated job.

### DIFF
--- a/Source/Thunder/PluginServer.h
+++ b/Source/Thunder/PluginServer.h
@@ -2403,7 +2403,7 @@ namespace PluginHost {
                     // even worse, use the communicator thread..
                     _deadProxiesProtection.Lock();
                     while (_deadProxies.empty() == false) {
-                        std::pair<uint32_t, Core::IUnknown*> entry(std::move(_deadProxies.back()));
+                        std::pair<uint32_t, Core::IUnknown*> entry(_deadProxies.back());
                         _deadProxies.pop_back();
 
                         // Now the communicator thread can continue! In the span of the lock
--- a/Source/Thunder/PluginServer.h
+++ b/Source/Thunder/PluginServer.h
@@ -2208,6 +2208,47 @@ namespace PluginHost {
                     string _observerPath;
                 };
 
+        class DanglingNotifierJob : public Core::IDispatch {
+            protected:
+                DanglingNotifierJob(CommunicatorServer* commServer, RPC::Administrator::Proxies& deadProxies)
+                    : _commServer(commServer)
+                    , _deadProxies(deadProxies)
+                {
+                }
+
+            public:
+                DanglingNotifierJob() = delete;
+                DanglingNotifierJob(const DanglingNotifierJob&) = delete;
+                DanglingNotifierJob& operator=(const DanglingNotifierJob&) = delete;
+
+                ~DanglingNotifierJob() override
+                {
+                    std::vector<ProxyStub::UnknownProxy*>::const_iterator loop(_deadProxies.begin());
+                    while (loop != _deadProxies.end()) {
+                        // To avoid race conditions, the creation of the deadProxies took a reference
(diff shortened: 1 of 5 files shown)


Source/Thunder/PluginServer.h
```diff
@@ -2000,7 +2000,7 @@ namespace PluginHost {
             private:
                 using Observers = std::vector<IShell::ICOMLink::INotification*>;
                 using Proxy = std::pair<uint32_t, const Core::IUnknown*>;
-                using Proxies = std::vector<Proxy>;
+                using Proxies = std::list<Proxy>;
```
pwielders: Why are we revrting back to the std::list std::vector is faster and cheaper here, so is there a real need for std::list?

sramani-metro: In Dispatch we want to pop_front of the list inside the lock and call the observers after releasing the lock. Since lists support front() and pop_front(), i though it will be easier.

Source/Thunder/PluginServer.h
```diff
@@ -2378,20 +2388,26 @@ namespace PluginHost {
...
+                        Proxy entry(_deadProxies.front());
+                        _deadProxies.pop_front();
+                        _adminLock.Unlock();
+
+                        _parent.Dangling(entry.second, entry.first);
 
-                    for (const Proxy& entry : _deadProxies) {
                         for (IShell::ICOMLink::INotification* observer : _requestObservers) {
                             observer->Dangling(entry.second, entry.first);
                             if (entry.second->Release() != Core::ERROR_DESTRUCTION_SUCCEEDED) {
```
pwielders: This is a potential risk, and it probably existed from the beginning but you must move this section:
```
if (entry.second->Release() != Core::ERROR_DESTRUCTION_SUCCEEDED) {
    TRACE(Trace::Warning, (_T("Potentially a Proxy leak on interface %d"), entry.first));
}
```

outside the for loop. Now if there are *more* subscribers to the interface, the second one might get a "dead" or already released interface. If there are no subsribers at al, the release is not happening at all :-)

Source/Thunder/PluginServer.h
```diff
@@ -2378,20 +2388,26 @@ namespace PluginHost {
...
                     _adminLock.Lock();
+                    while (_deadProxies.size() != 0) {
+                        Proxy entry(_deadProxies.front());
+                        _deadProxies.pop_front();
+                        _adminLock.Unlock();
+
+                        _parent.Dangling(entry.second, entry.first);
 
-                    for (const Proxy& entry : _deadProxies) {
                         for (IShell::ICOMLink::INotification* observer : _requestObservers) {
```
pwielders, line 2417: How is the _requestObservers protected ?

Source/Thunder/PluginServer.h
sramani-metro: If we are not concerned about the order of the Dangling notification to observers, we can revert back to vector. I have now changed it.

Source/Thunder/PluginServer.h
sramani-metro: Agreed and addressed.

Source/Thunder/PluginServer.h
sramani-metro: Noticed this is quickly becoming unreadable. Also, to avoid a lock dance in the Dispatch, i now work on the copy (taken inside lock) of _requestObservers.

pwielders (approved):

PR #1920, opened 2025-08-14
IController - Persist and Restore 

nxtum:
Previously there was one Persist function that would save the configs for all plugins without the ability to undo.
Now, you can specify which plugin to persist, as well as remove any/all configs (restore to default)

### DIFF
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -76,6 +76,7 @@ namespace PluginHost {
     /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::ExtensionsConfigDirectory = _T("extensions/");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T("Thunder/plugins/");
     /* static */ const TCHAR* Server::CommunicatorConnector = _T("COMMUNICATOR_CONNECTOR");
 
     static const TCHAR _defaultControllerCallsign[] = _T("Controller");
@@ -1232,7 +1233,7 @@ namespace PluginHost {
         IFactories::Assign(&_factoriesImplementation);
 
         // See if the persitent path for our-selves exist, if not we will create it :-)
-        Core::File persistentPath(_config.PersistentPath() + _T("PluginHost"));
+        Core::File persistentPath(_config.PersistentPath() + _T("Thunder/plugins"));
 
         if (persistentPath.IsDirectory() == false) {
             Core::Directory(persistentPath.Name().c_str()).Create();
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -1232,7 +1232,7 @@ namespace PluginHost {
     {
         IFactories::Assign(&_factoriesImplementation);
 
-        // See if the persitent path for our-selves exist, if not we will create it :-)
+        // See if the persistent path for our-selves exist, if not we will create it :-)
         Core::File persistentPath(_config.PersistentPath() + PluginOverrideDirectory);
 
         if (persistentPath.IsDirectory() == false) {
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -72,8 +72,6 @@ namespace PluginHost {
 #else
     /* static */ const TCHAR* Server::ConfigFile = _T("/etc/" EXPAND_AND_QUOTE(NAMESPACE) "/config.json");
 #endif
-
-    /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::ExtensionsConfigDirectory = _T("extensions/");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
     /* static */ const TCHAR* Server::PluginOverrideDirectory = _T(EXPAND_AND_QUOTE(NAMESPACE) "/services/");
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -76,7 +76,7 @@ namespace PluginHost {
     /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::ExtensionsConfigDirectory = _T("extensions/");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
-    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T("Thunder/plugins/");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T(EXPAND_AND_QUOTE(NAMESPACE) "/plugins/");
     /* static */ const TCHAR* Server::CommunicatorConnector = _T("COMMUNICATOR_CONNECTOR");
 
     static const TCHAR _defaultControllerCallsign[] = _T("Controller");
@@ -1233,7 +1233,7 @@ namespace PluginHost {
         IFactories::Assign(&_factoriesImplementation);
 
         // See if the persitent path for our-selves exist, if not we will create it :-)
-        Core::File persistentPath(_config.PersistentPath() + _T("Thunder/plugins"));
+        Core::File persistentPath(_config.PersistentPath() + PluginOverrideDirectory);
 
         if (persistentPath.IsDirectory() == false) {
             Core::Directory(persistentPath.Name().c_str()).Create();
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -76,7 +76,7 @@ namespace PluginHost {
     /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::ExtensionsConfigDirectory = _T("extensions/");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
-    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T(EXPAND_AND_QUOTE(NAMESPACE) "/plugins/");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T(EXPAND_AND_QUOTE(NAMESPACE) "/services/");
     /* static */ const TCHAR* Server::CommunicatorConnector = _T("COMMUNICATOR_CONNECTOR");
 
     static const TCHAR _defaultControllerCallsign[] = _T("Controller");
--- a/Source/Thunder/PluginServer.h
+++ b/Source/Thunder/PluginServer.h
@@ -1875,18 +1875,19 @@ namespace PluginHost {
             };
 
             using Callsigns = std::unordered_map<string, Plugin>;
+            using PluginOverrideFiles = std::unordered_map<string, string>; // [callsign, filepath]
 
         public:
             Override(const Override&) = delete;
             Override& operator=(const Override&) = delete;
 
-            Override(PluginHost::Config& serverconfig, ServiceMap& services, const string& persitentFile)
+            Override(PluginHost::Config& serverconfig, ServiceMap& services, const string& persitentFolder)
                 : Services()
                 , Prefix(serverconfig.Prefix())
                 , IdleTime(serverconfig.IdleTime())
                 , _services(services)
                 , _serverconfig(serverconfig)
-                , _fileName(persitentFile)
+                , _pluginOverrideFiles()
                 , _callsigns()
             {
                 Add(_T("Services"), &Services);
@@ -1903,6 +1904,10 @@ namespace PluginHost {
                         std::forward_as_tuple(name),
                         std::forward_as_tuple(_T("{}"), "", PluginHost::IShell::startmode::UNAVAILABLE, false)).first);
 
+                    // Create individual plugin configuration files
+                    string filePath = persitentFolder + name + "Override.json";
+                    _pluginOverrideFiles[name] = filePath;
+
                     // Store the override config in the JSON String created in the map
                     Services.Add(index->first.c_str(), &(index->second));
                 }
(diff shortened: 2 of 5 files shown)


Source/plugins/IController.h
```diff
@@ -86,9 +86,17 @@ namespace Controller {
     struct EXTERNAL IConfiguration : virtual public Core::IUnknown {
         enum { ID = RPC::ID_CONTROLLER_CONFIGURATION };
 
+        // @omit
+        Core::hresult Persist() { return Persist(Core::OptionalType<string>());}
```
nxtum: This could be removed, but will need to create 2 PR to update:
LocationSync.cpp https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/blob/master/LocationSync/LocationSync.cpp#L126

ConfigUpdateExample: https://github.com/rdkcentral/ThunderNanoServices/blob/master/examples/ConfigUpdateExample/ConfigUpdateExample.cpp#L63

pwielders: Please mark it as DEPRECATED so people using it will get a warning!

Source/Thunder/PluginServer.h
```diff
@@ -1770,6 +1771,10 @@ namespace PluginHost {
                         std::forward_as_tuple(name),
                         std::forward_as_tuple(_T("{}"), "", PluginHost::IShell::startmode::UNAVAILABLE, false)).first);
 
+                    // Create individual plugin configuration files
+                    string filePath = persitentFolder + name + "Override.json";
+                    _pluginOverrideFiles[name] = filePath;
```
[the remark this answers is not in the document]
nxtum: realised i also spelt persistentFolder as "persitentFolder"

Source/Thunder/PluginServer.cpp
```diff
@@ -73,7 +73,7 @@ namespace PluginHost {
     /* static */ const TCHAR* Server::ConfigFile = _T("/etc/" EXPAND_AND_QUOTE(NAMESPACE) "/config.json");
 #endif
 
-    /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T("PluginHost/");
```
pwielders: Might be good to move the directory now to Thunder/plugins and store the configs there, it is in the same line as the fixed configs are. Maybe you could even swap this line than with the nex line and use the default of the plugins config from there so place this:
```
/* static */ const TCHAR* Server::PluginOverrideDirectory = _T("Thunder/") + Server::PluginConfigDirectory;
```
on the line "after" the  Server::PluginConfigDirectory declaration

Source/Thunder/PluginServer.h
```diff
@@ -1776,6 +1777,17 @@ namespace PluginHost {
                         std::forward_as_tuple(name),
                         std::forward_as_tuple(_T("{}"), "", PluginHost::IShell::startmode::UNAVAILABLE, false)).first);
 
+                    // Create individual plugin configuration files
+                    string filePath;
+                    filePath.reserve(persistentFolder.size() + name.size() + sizeof("Override.json") - 1);
+                    filePath.append(persistentFolder).append(name).append("Override.json");
```
pwielders: If we put them all in a folder (e.g. previous remark Thunder/plugins. I guess we could than use the same sematics as we do in thunder an only append the name with json, so:
``` 
filePath.append(persistentFolder).append(name).append(".json");
```
than it is consistent with the configs we find in /etc/Thunder/plugins as the startin point. Consistency is key ;-)

pwielders (changes requested):
An overall remark. I guess in the past, because all plugins where in 1 file, we added an empty config ('{}') for plugins that had no override or where destroyed. However now the configs are single entries per callsign. Whenever you "add" a file it probably takes up a sector on flash storage per file (much more than two byes to signal {} in the one file approach). I guess it is worth the effort to *not* store an override file if there are no changes to to config. So suggest to add this optimization in the code.
No changes on the plugin, no override file in the thunder/plugins directpry o persitent storage space to safe flash sectors...

pwielders (commented):
Questions/Remark:
1) Why is in the save, the way to create an override a lamda i.s.o. a normal private method?
2) Comparing if the configs are different might be a bit more tricky. Cause officially:
```
{ "fieldA": 1, "fieldB" : 2 } === { "fieldB": 2, "fieldA": 1}
```
This is now detected as different configs.
And probbaly even worse: 
```
{     "field":   1 } === {"field":1}
```
Is already detected as a different config :-)

got 2 suggestions to avoid this (or make it more clear):
1) Use a flag on the service that if the "config" is set through a method, flag it as dirty and consider the config to be changed (even if it was not changed :-)
2) enhance the Core::JSON::VariantContainer class with an ```bool operator ==(const VariantContainer& ) const``` and load each string into the VariantContainer class and compare those (field availability and there corresponding values)

I have a preference for the second option enhance the JSON class :-) might be usefull in more places and the VariantContainer is already used anway, somewhere in the PluginServer code but aI am also interested in @sebaszm opinion here...

Source/Thunder/PluginServer.cpp
```diff
@@ -72,9 +72,8 @@ namespace PluginHost {
 #else
     /* static */ const TCHAR* Server::ConfigFile = _T("/etc/" EXPAND_AND_QUOTE(NAMESPACE) "/config.json");
 #endif
-
-    /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T("Thunder/plugins/");
```
sebaszm: Use EXPAND_AND_QUOTE(NAMESPACE) instead of using framework name directly.

Source/Thunder/PluginServer.cpp
```diff
@@ -1123,7 +1122,7 @@ namespace PluginHost {
         IFactories::Assign(&_factoriesImplementation);
 
         // See if the persitent path for our-selves exist, if not we will create it :-)
-        Core::File persistentPath(_config.PersistentPath() + _T("PluginHost"));
+        Core::File persistentPath(_config.PersistentPath() + _T("Thunder/plugins"));
```
sebaszm: Use the constant instead of repeating the path name.

Source/Thunder/PluginServer.h
```diff
@@ -1776,6 +1777,17 @@ namespace PluginHost {
...
+                    string filePath;
+                    filePath.reserve(persistentFolder.size() + name.size() + sizeof(".json") - 1);
+                    filePath.append(persistentFolder).append(name).append(".json");
+
+                    _pluginOverrideFiles.emplace(
+                        std::piecewise_construct,
+                        std::forward_as_tuple(name),
+                        std::forward_as_tuple(std::move(filePath))
+                    );
+
```
sebaszm: Do we really need to remember the filenames for each callsign? if we only stored the base folder having a callsing we can every time find the right path. I don't think configs are being overiden so frequent to justify this flamboyance with memory.

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
 
-                    while (index.Next() == true) {
+                        // Clear all currently set values, they might be from the previous run.
+                        Clear();
 
-                        Callsigns::const_iterator current(_callsigns.find(index->Callsign()));
+                        // Read the file and parse it into this object.
+                        IElement::FromFile(storage);
+                        _serverconfig.SetPrefix(Prefix.Value());
+                        _serverconfig.SetIdleTime(IdleTime.Value());
```
sebaszm: This part shouldn't be repeated for each plugin config. This clears and looks for prefix and idletime in each of the plugins' config while this is Thunder property.

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
                 return (result);
             }
 
-            bool Save()
+            bool Save(const Core::OptionalType<string>& callsign)
             {
                 uint32_t result = Core::ERROR_NONE;
+                
+                auto CreateOverride = [this, &result](ServiceMap::Iterator& indexService) {
+                    const string currentCallsign = indexService->Callsign();
```
sebaszm: A copy!!

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
             {
                 uint32_t result = Core::ERROR_NONE;
 
-                Core::File storage(_fileName);
+                ServiceMap::Iterator indexService(_services.Services());
+                while (indexService.Next() == true) {
+                    const string currentCallsign = indexService->Callsign();
```
sebaszm: a copy!

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
+                        Core::JSON::ArrayType<Thunder::Plugin::Config>::Iterator indexConfig = _serverconfig.Plugins();
+                        while (indexConfig.Next()) {
+                            if (indexConfig.Current().Callsign.Value() == currentCallsign) {
+                                readOnlyConfig = &indexConfig.Current();
+                                break;
+                            }
+                        }
 
+                        // ONLY persist when the config changes from the default config
+                        if (readOnlyConfig == nullptr || config == readOnlyConfig->Configuration.Value()) {
```
sebaszm: This is a bit naive as it's a string compare not JSON compare and you may wrongly consider a config changed where it is technically holding the same information.

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
+                            return;
```
sebaszm: I don't thik we endorse multiple exit points. :)

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
+                        }
+
+                        Core::File storage(indexFile->second);
+                        if (storage.Create() == true) {
+
+                            // Clear all currently set values, they might be from the previous run.
+                            Clear();
+    
+                            Prefix    = _serverconfig.Prefix();
+                            IdleTime  = _serverconfig.IdleTime();
```
sebaszm: Again these are not plugin properties and are saved for every plugin. I think these need to go to a separate config file.

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
+                    PluginOverrideFiles::const_iterator indexFile = _pluginOverrideFiles.find(currentCallsign);
+                    ASSERT(indexFile != _pluginOverrideFiles.end());
 
-                if (storage.Create() == true) {
+                    Callsigns::iterator indexCallsigns(_callsigns.find(currentCallsign));
+                    ASSERT(indexCallsigns != _callsigns.end());
 
-                    // Clear all currently set values, they might be from the precious run.
-                    Clear();
+                    string config(indexService->ConfigLine());
```
sebaszm: a copy

Source/Thunder/PluginServer.h
```diff
@@ -1791,110 +1803,183 @@ namespace PluginHost {
...
+                            }
+                        }
 
+                        // ONLY persist when the config changes from the default config
+                        if (readOnlyConfig == nullptr || config == readOnlyConfig->Configuration.Value()) {
+                            return;
+                        }
+
+                        Core::File storage(indexFile->second);
+                        if (storage.Create() == true) {
```
sebaszm: Let's not write the override config at all if it's not changed! Let the absence of the override file mean there is no override.

Source/Thunder/PluginServer.h
```diff
@@ -4769,16 +4854,21 @@ namespace PluginHost {
...
+        {
+            Override infoBlob(_config, _services, Configuration().PersistentPath() + PluginOverrideDirectory);
+            return (infoBlob.Destroy(callsign));
         }
+
         uint32_t Load()
         {
-            Override infoBlob(_config, _services, Configuration().PersistentPath() + PluginOverrideFile);
-
+            Override infoBlob(_config, _services, Configuration().PersistentPath() + PluginOverrideDirectory);
```
sebaszm, line 5220: I find this approach inefficient. Even if we are working with a single callsign we still create the Override object and prepare for handling all plugins (ie. saving callsiings, configs, filenames.... etc).

Source/plugins/IController.h
```diff
@@ -86,9 +86,17 @@ namespace Controller {
     struct EXTERNAL IConfiguration : virtual public Core::IUnknown {
         enum { ID = RPC::ID_CONTROLLER_CONFIGURATION };
 
+        // @omit @deprecated
+        DEPRECATED Core::hresult Persist() { return Persist(Core::OptionalType<string>());}
+
         // @alt storeconfig
```
sebaszm, line 89: Perahps we can finally depracate the old JSON-RPC name "storeconfig" using alt:depracated storeconfig.

Source/plugins/IController.h
```diff
@@ -86,9 +86,17 @@ namespace Controller {
     struct EXTERNAL IConfiguration : virtual public Core::IUnknown {
         enum { ID = RPC::ID_CONTROLLER_CONFIGURATION };
 
+        // @omit @deprecated
```
sebaszm: omit and deprecated tags are not needed as this is not a virtual method and will not be considered for code generation at any level.

Source/plugins/IController.h
```diff
@@ -86,9 +86,17 @@ namespace Controller {
...
         // @alt storeconfig
-        // @brief Stores all configuration to the persistent memory
-        virtual Core::hresult Persist() = 0;
+        // @brief Stores configuration to the persistent memory
+        // @param callsign: Callsign to persist
+        virtual Core::hresult Persist(const Core::OptionalType<string>& callsign) = 0;
+
+        // @brief Restores configuration back to default
+        // @param callsign: Callsign to restore
+        virtual Core::hresult Restore(const Core::OptionalType<string>& callsign) = 0;
```
sebaszm: Perhaps we could consider making the callsign defaultly unset with default construction = { } ? Otherwise I need to call it like Restore(Core::OptionalType<string>()), which - fow what it's worth - I don't like. :) Or maybe we should introduce std::nullopt equivalent for Core::OptionalType.

sebaszm (changes requested):

sebaszm:
> 1. Use a flag on the service that if the "config" is set through a method, flag it as dirty and consider the config to be changed (even if it was not changed :-)
> 2. enhance the Core::JSON::VariantContainer class with an `bool operator ==(const VariantContainer& ) const` and load each string into the VariantContainer class and compare those (field availability and there corresponding values)
> 
> I have a preference for the second option enhance the JSON class :-) might be usefull in more places and the VariantContainer is already used anway, somewhere in the PluginServer code but aI am also interested in @sebaszm opinion here...
1. does not solve the problem much if Store() is used with no callsign then all plugins will be marked dirty even if not changed.

Agree, to do this properly we would need to normalize the JSON first. JSON::Variant would be a perfect starting point to achieve this (it is though behind COMPLEMENTARY_CODE_SET flag). We however will encounter nuance like order of elements in array (is [a,b] same as [b,a]? depends), or explicitly setting a value that is same as a default. I think too that it would be a useful feature, but perhaps we should split this into another PR.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
-                    Prefix    = _serverconfig.Prefix();
-                    IdleTime  = _serverconfig.IdleTime();
+                        const Thunder::Plugin::Config* readOnlyConfig = nullptr;
+                        Core::JSON::ArrayType<Thunder::Plugin::Config>::Iterator indexConfig = _serverconfig.Plugins();
+                        while (indexConfig.Next()) {
+                            if (indexConfig.Current().Callsign.Value() == currentCallsign) {
+                                readOnlyConfig = &indexConfig.Current();
+                                break;
+                            }
+                        }
```
sebaszm: This is already implemented as PluginHost::Plugin(callsign), so you don't have to lookup yourself.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
 
-                    // Red the file and parse it into this object.
-                    IElement::FromFile(storage);
+                        result = true;
 
-                    _serverconfig.SetPrefix(Prefix.Value());
-                    _serverconfig.SetIdleTime(IdleTime.Value());
-                    // Convey the real JSON struct information into the specific services.
-                    ServiceMap::Iterator index(_services.Services());
+                        Clear();
```
sebaszm: I think this clears the whole object (ie. all services). Not only the element you're trying to load. So it seems to me that after all iterations you'll end up only with the last one loaded.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                        // Read the file and parse it into this object.
+                        IElement::FromFile(storage);
```
sebaszm: This requires that the saved JSON has extra level of indirection "servces": { "name": { ... } }.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                return result;
             }
 
-            bool Save()
+            bool Save(const Core::OptionalType<string>& callsign)
             {
                 uint32_t result = Core::ERROR_NONE;
 
-                Core::File storage(_fileName);
+                SavePluginHostConfig();
```
sebaszm: The extra pluginhost config is saved each and every time a Save is made to all or any plugin.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
 
-                    // Convey the real information from he specific services into the JSON struct.
-                    ServiceMap::Iterator index(_services.Services());
+                        // ONLY persist when the config changes from the default config
+                        if (readOnlyConfig != nullptr && config != readOnlyConfig->Configuration.Value()) {
+                            Core::File storage(CreateOverridePath(currentCallsign));
+                            if (storage.Create() == true) {
+
+                                // Clear all currently set values, they might be from the previous run.
+                                Clear();
```
sebaszm: Again we clear everything up here.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                                auto indexCallsigns = RegisterService(currentCallsign);
+                                ASSERT(indexCallsigns != _callsigns.end());
+
+                                indexCallsigns->second.Configuration = config;
+                                indexCallsigns->second.SystemRootPath = indexService->SystemRootPath();
+                                indexCallsigns->second.StartMode = indexService->StartMode();
+                                indexCallsigns->second.Resumed = indexService->Resumed();
+        
+                                // Persist the currently set information
+                                IElement::ToFile(storage);
```
sebaszm: I believe this saves ALL config (config of all servcies) to the particular file!

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                    ServiceMap::Iterator index(_services.Services());
+                    while (index.Next() == true) {
+                        CreateOverride(index);
+                    }
+                }
+                return result;
+            }
 
-                        Callsigns::iterator current(_callsigns.find(index->Callsign()));
+            bool Destroy(const Core::OptionalType<string>& callsign)
```
sebaszm: There is no way to delete the pluginhost config.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
-            bool Save()
+            bool Save(const Core::OptionalType<string>& callsign)
             {
                 uint32_t result = Core::ERROR_NONE;
 
-                Core::File storage(_fileName);
+                SavePluginHostConfig();
 
-                if (storage.Create() == true) {
+                auto CreateOverride = [this, &result](const ServiceMap::Iterator& indexService) {
```
sebaszm: nitpick: Perhaps it's more readable if we move this to a meber method insteadof lambda?

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                    _serverconfig.SetPrefix(Prefix.Value());
+                    _serverconfig.SetIdleTime(IdleTime.Value());
                     storage.Close();
-                } else {
-                    result = storage.ErrorCode();
                 }
+            }
 
-                return (result);
+            void SavePluginHostConfig() {
```
sebaszm: Always saves the elements even if they have not changed from defaults.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                Core::File storage(CreateOverridePath("PluginHost"));
+                if (storage.Create()) {
+                    Clear();
+
+                    Prefix    = _serverconfig.Prefix();
+                    IdleTime  = _serverconfig.IdleTime();
+
+                    IElement::ToFile(storage);
```
sebaszm: Again I believe this will save all config to this file.

Source/Thunder/PluginServer.h
```diff
@@ -1777,110 +1760,220 @@ namespace PluginHost {
...
+                    storage.Close();
+                }
             }
 
-            Core::JSON::Container Services;
+            string CreateOverridePath(const string& callsign) const {
```
sebaszm, line 2231: nitpick: I think I would be a lot more readable we simply had persistetFolder + callsign + ".json" :)

Source/plugins/IController.h
```diff
@@ -86,9 +86,17 @@ namespace Controller {
...
 
+        // @alt:deprecated storeconfig
+        DEPRECATED Core::hresult Persist() { return Persist(Core::OptionalType<string>());}
+
         // @alt storeconfig
-        // @brief Stores all configuration to the persistent memory
-        virtual Core::hresult Persist() = 0;
+        // @brief Stores configuration to the persistent memory
+        // @param callsign: Callsign to persist
+        virtual Core::hresult Persist(const Core::OptionalType<string>& callsign) = 0;
```
sebaszm: Why not merge the two Persist method the same way Restore is defined?

sebaszm (changes requested):

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+                            result = storage.ErrorCode();
+                        }
+                    }
+                }
+                return result;
+            }
 
-                        // ServiceMap should *NOT* change runtime...
-                        ASSERT(current != _callsigns.end());
+            bool Save(const Core::OptionalType<string>& callsign)
```
sebaszm: This logic seems incorrect? Unset should mean save everything (incl. pluginhost), set and "" save Controller, set and callsign save that only service (also if controller).

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
-                            if (current->second.SystemRootPath.IsSet() == true) {
-                                index->SystemRootPath(current->second.SystemRootPath.Value());
-                            }
-                            if (current->second.StartMode.IsSet() == true) {
-                                index->StartMode(current->second.StartMode.Value());
-                            }
-                            if (current->second.Resumed.IsSet() == true) {
-                                index->Resumed(current->second.Resumed.Value());
-                            }
+                if (!callsign.IsSet() || callsign.Value() == "PluginHost") {
```
sebaszm: We are repeating "PluginHost" multiple times, move it to a constant. also use the _T() macro! Or I think it would be **even better** use the NAMESPACE macro (expanded simply to  "Thunder"), so we don't have to hardcode anything.

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+                    SavePluginHostConfig();
+                }
+
+                if (callsign.IsSet() == true && callsign.Value() != "PluginHost") {
+                    bool found = false;
+                    const string& raw = callsign.Value();
+                    const string target = raw.empty() ? _T("Controller") : raw;
```
sebaszm: we must not hardcode "Controller" this can change!

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
                     }
 
-                    storage.Close();
-                } else {
-                    result = storage.ErrorCode();
+                    if (found == false) {
+                        result = Core::ERROR_GENERAL;
+                    }
+                }
+                else {
```
sebaszm: Oops! Seems to me that if set to PluginHost it'll still save all services!

Source/Thunder/PluginServer.h
sebaszm: result should be uint32_t

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+                        PersistOverride(*shell, result);
+                    }
                 }
 
-                return (result);
+                return result;
             }
 
-            bool Save()
+            bool Destroy(const Core::OptionalType<string>& callsign)
```
sebaszm: again, mixing bool and uint32_t  on result

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
                     }
+                }
 
-                    // Persist the currently set information
-                    IElement::ToFile(storage);
+                out = std::move(root);
+                return true;
+            };
 
+            auto PersistVariant = [&](const Core::JSON::Variant& configurationValue) {
```
sebaszm: PersistVariant appears to me identical to PersistString, surely variant has a ToString method

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+                ASSERT(it != _callsigns.end());
+
+                it->second.Configuration = config;
+                it->second.SystemRootPath = shell.SystemRootPath();
+                it->second.StartMode = shell.StartMode();
+                it->second.Resumed = shell.Resumed();
+
+                it->second.IElement::ToFile(storage);
+                storage.Close();
+            } else if (result == Core::ERROR_NONE) {
```
sebaszm: Third time the same code

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+
+            void DestroyOverride(const string& target, uint32_t& result)
+            {
+                Core::File storage(CreateOverridePath(target));
+                if (storage.Exists() == true) {
+                    if (storage.Destroy() == false) {
+                        result = storage.ErrorCode();
+                    }
+                }
+            }
```
sebaszm, line 2177: what's the result if all succeeded? why return by param?

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+
+            void LoadPluginHostConfig()
+            {
+                Core::File storage(CreateOverridePath("PluginHost"));
+                if (storage.Exists() && storage.Open(true) == true) {
+                    Clear();
```
sebaszm: I think it clears everything??

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+            }
+
+            void SavePluginHostConfig()
+            {
+                const string& currentPrefix = _serverconfig.Prefix();
+                const uint16_t currentIdleTime = _serverconfig.IdleTime();
+
+                const bool differs = (currentPrefix != _defaultPrefix) || (currentIdleTime != _defaultIdleTime);
+
+                if (differs == false) {
```
sebaszm: Normally we refrain from multiple returns!

Source/Thunder/PluginServer.h
```diff
@@ -1917,111 +1912,330 @@ namespace PluginHost {
...
+                    return;
+                } 
+
+                Core::File storage(CreateOverridePath("PluginHost"));
+                if (storage.Create()) {
+                    Clear();
```
sebaszm: again I think Clear removes everything

sebaszm (changes requested):

Source/Thunder/PluginServer.cpp
```diff
@@ -72,10 +72,9 @@ namespace PluginHost {
 #else
     /* static */ const TCHAR* Server::ConfigFile = _T("/etc/" EXPAND_AND_QUOTE(NAMESPACE) "/config.json");
 #endif
-
-    /* static */ const TCHAR* Server::PluginOverrideFile = _T("PluginHost/override.json");
     /* static */ const TCHAR* Server::ExtensionsConfigDirectory = _T("extensions/");
     /* static */ const TCHAR* Server::PluginConfigDirectory = _T("plugins/");
+    /* static */ const TCHAR* Server::PluginOverrideDirectory = _T(EXPAND_AND_QUOTE(NAMESPACE) "/plugins/");
```
sebaszm: let's rename /plugins/ to  /services/ as also configs for extensions will sitting there!

sebaszm (changes requested):

Source/plugins/IController.h
```diff
@@ -87,8 +87,13 @@ namespace Controller {
...
         // @alt storeconfig
-        // @brief Stores all configuration to the persistent memory
-        virtual Core::hresult Persist() = 0;
+        // @brief Stores configuration to the persistent memory
+        // @param callsign: Callsign to persist
+        virtual Core::hresult Persist(const Core::OptionalType<string>& callsign = {}) = 0;
+
+        // @brief Restores configuration back to default
+        // @param callsign: Callsign to restore
+        virtual Core::hresult Restore(const Core::OptionalType<string>& callsign = {}) = 0;
```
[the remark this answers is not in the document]
sebaszm: This has only one implementation.

Source/Thunder/PluginServer.cpp
```diff
@@ -72,10 +72,9 @@ namespace PluginHost {
...
     /* static */ const TCHAR* Server::CommunicatorConnector = _T("COMMUNICATOR_CONNECTOR");
```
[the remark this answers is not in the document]
sebaszm: Migration support is not required.

sebaszm (approved):

PR #1922, opened 2025-08-18
Metrol 1145:move lookup and async example to master

sivarajappan-siva:


sebaszm (dismissed):

Source/com/Ids.h
```diff
@@ -96,7 +96,8 @@ namespace RPC {
 
         ID_EXTERNAL_INTERFACE_OFFSET               = (ID_OFFSET_INTERNAL + 0x0080),
         ID_EXTERNAL_QA_INTERFACE_OFFSET            = (0xA000),
-        ID_EXTERNAL_CC_INTERFACE_OFFSET            = (0xCC00) // ends on 0xDFFF
+        ID_EXTERNAL_CC_INTERFACE_OFFSET            = (0xCC00), // ends on 0xDFFF
+        ID_EXTERNAL_SAMPLE_INTERFACE_OFFSET        = (0xE000)
```
MFransen69: For consistency in the ordering in this file can't we make it 0XB000, that gives room for roughly 4000 qa and example interfaces, I guess that should be enough?

MFransen69 (changes requested):

Source/com/Ids.h
sivarajappan-siva: Yes,that will be enough.Updated the code accordingly

sebaszm (dismissed):

sebaszm (approved):

MFransen69 (approved):

PR #1927, opened 2025-08-21
[COMRPC] Close the socket when a COMRPC call timesout

sramani-metro:
When a COMRPC call times out, on a channel, most of the times the caller gives up and continues without handling the error. But the other side may still send a response when its not expected. This causes the response message to be processed for a different outbound method from Thunder on the wrong time resulting in a memory corruption and crash. This change closes the SocketPort when there is a COMRPC timeout thereby avoiding subsequent outbound messages on the channel and also triggering fault tolerance code on the communication thread to kick-in and gracefully shudown the plugin.

### DIFF
--- a/Source/com/IUnknown.h
+++ b/Source/com/IUnknown.h
@@ -224,6 +224,10 @@ namespace ProxyStub {
                         result = _channel->Invoke(message, RPC::CommunicationTimeOut);
 
                         if (result != Core::ERROR_NONE) {
+                            if (result == Core::ERROR_TIMEDOUT) {
+                                SYSLOG(Logging::Error, (_T("IPC method Invoke failed due to timeout (Interface ID 0x%X, Method ID 0x%X). Execution of code may or may not have happened. Side effects are to be expected after this message"), message->Parameters().InterfaceId(), message->Parameters().MethodId()));
+                                Shutdown();
+                            }
                             TRACE_L1("Could not remote release the Proxy.");
                             result |= COM_ERROR;
                         }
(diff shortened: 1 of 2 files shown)


pwielders (approved):

Source/com/IUnknown.h
```diff
@@ -224,6 +224,10 @@ namespace ProxyStub {
                         result = _channel->Invoke(message, RPC::CommunicationTimeOut);
 
                         if (result != Core::ERROR_NONE) {
+                            if (result == Core::ERROR_TIMEDOUT) {
+                                SYSLOG(Logging::Error, (_T("IPC method Invoke failed due to timeout (Interface ID 0x%X, Method ID 0x%X). Execution of code may or may not have happened. Side effects are to be expected after this message"), message->Parameters().InterfaceId(), message->Parameters().MethodId()));
```
pwielders: Can we not move the SYSLOG message into the Shutdown? Seems a bit redundant right now ? Suggest to use the message as a parameter to shutdown. If later we want to enhance the logging we will do so for both locations in 1 go..
.

pwielders (changes requested):
Minor change requested, just avoid that the SYSLOG messages are going to deviate over time as the same message is currentl in two locations which both mena the sameand could be placed in the method that actually does the work (shutdown())

Source/com/IUnknown.h
sramani-metro: Done

pwielders (approved):

PR #1933, opened 2025-09-08
[RDKDEV-1144] Thunder leaking sockets

tomasz-karczewski-red:
It was observed that when calling WebKit plugin URL endpoint with curl, the socket on wpeframework side stays in CLOSE_WAIT state indefinetely, which may lead to running out of file descriptors and crash. The reason
seems to be that on Thunder side, the socket was never closed.

reproduction/test scenario:

- assuming Thunder is listening on port 9998
- tested with WebKitBrowser API, but other APIs behave similarly

the script:

    while true; do
    curl http://127.0.0.1:9998/Service/WebKitBrowser/URL -d '{"url":"https://widgets.metrological.com/lightning/liberty/017b288dfb351c11eb33754fc76c12fc#boot"}'
    netstat -tnp 2>&1 | grep :9998 | grep CLOSE_WAIT | wc -l
    done

The above script will show the number of sockets in CLOSE_WAIT state growing. The solution is to Close() the socket on Thunder side as well when Channel is going into CLOSED state.

On SocketPort side, the case where number of bytes read is 0 is 'orderly shutdown', so also 'ERROR' flag in such case (https://man7.org/linux/man-pages/man2/recv.2.html)

### DIFF
--- a/Source/Thunder/PluginServer.h
+++ b/Source/Thunder/PluginServer.h
@@ -4443,6 +4443,11 @@ namespace PluginHost {
 
                         _service.Release();
                     }
+                    if (!IsSuspended()) {
+                        // need to close our side as well, to avoid sockets
+                        // staying indefinetely in CLOSE_WAIT
+                        Close(0);
+                    }
 
                     State(CLOSED, false);
 
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -1240,7 +1240,9 @@ namespace Thunder {
 
                 if (l_Size == 0) {
                     if ((m_State & SocketPort::LINK) != 0) {
-                        m_State = ((m_State & (~SocketPort::OPEN)) | SocketPort::EXCEPTION);
+                        // peer has performed an orderly shutdown
+                        m_State &= (~SocketPort::OPEN);
+                        StateChange();
                     }
                 }
                 else if (l_Size != static_cast<uint32_t>(SOCKET_ERROR)) {


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-5432

pwielders:
@tomasz-karczewski-red / @pradeeptakdas Issue is in further investigation. Theoretical discussion learned that the rootcause here is indeed the missing StateChange calls so the "owning" class/object of the socket can be signalled of a change. Removing the EXCEPTION BIT from the socket port state seems correctish but not complete and should be (SocketPort::SHUTDOWN) as the connection is shutdown from the other side.
So there is a bug, the way to solve it is, however, still a bit uncertain and under investigation.
We will use the given test scenario above (thanks for this simple reproduction scenario @tomasz-karczewski-red ) to validate the fix.

pwielders:
@tomasz-karczewski-red we tried reproducing this on R5 (latest) and it seems we can not reproduce it with the script you supplied. Can you indicate which version of Thunder you used to observe this?

tomasz-karczewski-red:
> @tomasz-karczewski-red we tried reproducing this on R5 (latest) and it seems we can not reproduce it with the script you supplied. Can you indicate which version of Thunder you used to observe this?

hello @pwielders  - I was testing on 4.2 (https://github.com/rdkcentral/Thunder/commit/8fa0abb34e7db00905fcf6ab58e05f9860ca1b50), on 32-bit linux

tomasz-karczewski-red:
@pwielders one more thing: I think this only happens in case where there is no 'connection: close' header. So it might depend on the http client you're using. The one we have on our box doesn't add 'connection: close' by default:

curl -v http://127.0.0.1:9998/Service/WebKitBrowser/URL -d '{"url":"https://widgets.metrological.com/lightning/liberty/017b288dfb351c11eb33754fc76c12fc#boot"}'
* STATE: INIT => CONNECT handle 0x1994090; line 1833 (connection #-5000)
* Added connection 0. The cache now contains 1 members
* family0 == v4, family1 == v6
*   Trying 127.0.0.1:9998...
* STATE: CONNECT => CONNECTING handle 0x1994090; line 1894 (connection #0)
* Connected to 127.0.0.1 (127.0.0.1) port 9998 (#0)
* STATE: CONNECTING => PROTOCONNECT handle 0x1994090; line 2026 (connection #0)
* STATE: PROTOCONNECT => DO handle 0x1994090; line 2049 (connection #0)
> POST /Service/WebKitBrowser/URL HTTP/1.1
> Host: 127.0.0.1:9998
> User-Agent: curl/7.83.0
> Accept: */*
> Content-Length: 98
> Content-Type: application/x-www-form-urlencoded

pwielders:
@tomasz-karczewski-red thanks, I think the addition (there is no 'connection: close') is I guess essential to the test ;-)
@Karthick-Somasundaresan please take this into account when testing as I think that this might be why you could not reproduce it on master as well..

Karthick-Somasundaresan:
Hi @tomasz-karczewski-red , 


The fix for this socket leak is available from Thunder 4.4. Yet the idea of propagating the state change is valid, and it is needed to add more states to convey the current state properly to the layers above. I've raised a PR for this https://github.com/rdkcentral/Thunder/pull/1941. Please have a look into it.

pwielders:
@tomasz-karczewski-red Thanks for the PR and pointing out the missing StateChange call. Will close this PR in favor of https://github.com/rdkcentral/Thunder/pull/1941 as this adds the new  state to do proper signaling to the upper layers.

PR #1939, opened 2025-09-25
[SERVICE DISCOVERY] Change the way the services are discovered from c…

pwielders:
…entralized to decentralized.

As the service discovery was centralized, the build-up of the centralized list had to be protected. Service registration in the centralized setup was done at the moment, an so (dll) is loaded.

The dyamic loader of linux (dl) has a mutex to protect other DL calls to happen while an open or close is in progress. This means that if a library is opend/loaded, the dl mutex (Lock A) is taken. During the loading the the statics in the module are run, as part of this run, the services are registered in the centralized services store, which in turn use a mutex to protect the list of available service (Lock B)

Now if a service is constructed, the centralized services are consulted, to make sure they are not removed, while we creat a service, the list is protected with a mutex (Lock B), if during the instatiation of this service, other libraries need to be loaded, the linux dynamic loader will call dlopen which in turn lock any other action for dl by taking a mutex (Lock A).

If you have a serious QA department that tests properly (automated test that continously start and stop Thunder) you will see that the slim change of the multithreaded Thunder startup might cause a coincident "Load Library" and "Create Service" between the taking of the first lock and be eempted before the second lock was taken, that will lead to a classical ABBA lock.

To resolve it, initially we used a paradigm to make sure that lock B was *always* taken before lock A was taken. Unfortunatly this brings a bit of implementation knowledge to the developer of a plgin that is very hard to explain nor to check / validate. So this might lead to unexpected deadlocks (with a very slim chance) in the future if the B lock was not always taken before the A lock. The goal of this PR is to remove the use of the, till now, nessecary lock B (Lock A is in the linux dynamic loader and can not be controlled by the Thunder coe). This is done by using a decentralized approach. The services (and there factory implementations) are nolonge "registered" centrally, but in stead the Library (if loaded) is now searched for a "IService" entry point. This Iservice is now enhancedwith a Next(), which allows to iterate over all services in this library. This chain is build during the loading of the library (Single thread and protected by the dl mutex ;-) ), Once the library is loaded, the dl's lifetime is protected wth the dl's handle refernce count (implemented in Core::Library) as the lifetime for creation is now protected by the ref count on the library and the list is not subject to change anymore, we are nolonger in the need of "Lock B" which resolves the "under the hood" knowledge!

### DIFF
--- a/Source/Thunder/PluginServer.h
+++ b/Source/Thunder/PluginServer.h
@@ -1382,9 +1382,19 @@ namespace PluginHost {
             void LoadMetadata() {
                 const string locator(PluginHost::Service::Configuration().Locator.Value());
                 if (locator.empty() == false) {
-                    Core::Library loadedLib = LoadLibrary(locator);
-                    if (loadedLib.IsLoaded() == true) {
-                        Core::ServiceAdministrator::Instance().ReleaseLibrary(std::move(loadedLib));
+                    Core::Library loadedLib;
+                    const Core::IService* service(LoadLibrary(locator, loadedLib));
+                    if (service != nullptr) {
+                        ASSERT (loadedLib.IsLoaded() == true);
+                        Core::System::ModuleBuildRefImpl moduleBuildRef(reinterpret_cast<Core::System::ModuleBuildRefImpl>(loadedLib.LoadFunction(_T("ModuleBuildRef"))));
+                        if (moduleBuildRef != nullptr) {
+			    _metadata.Hash(moduleBuildRef());
+                        }
+                        _metadata = service;
+                        if (_metadata.IsValid() == true) {
+                            _precondition.Load(_metadata.Precondition());
+                            _termination.Load(_metadata.Termination());
+                        }
                     }
                 }
             }
@@ -1441,24 +1451,20 @@ namespace PluginHost {
                 return (Core::ServiceType<RPC::StringIterator>::Create<RPC::IStringIterator>(searchPaths));
             }
 
-            Core::Library LoadLibrary(const string& name) {
-                uint8_t progressedState = 0;
-                Core::Library result;
+            const Core::IService* LoadLibrary(const string& name, Core::Library& library) {
+                Core::IService* result(nullptr);
 
                 RPC::IStringIterator* all_paths = GetLibrarySearchPaths(name);
                 ASSERT(all_paths != nullptr);
 
                 string element;
-                while((all_paths->Next(element) == true) && (progressedState <= 2)) {
+                while((all_paths->Next(element) == true) && (result == nullptr)) {
 
                     TRACE_L1("attempting to load library %s", element.c_str());
 
                     Core::File libraryToLoad(element);
 
                     if (libraryToLoad.Exists() == true) {
-                        if (progressedState == 0) {
-                            progressedState = 1;
-                        }
 
                         // Loading a library, in the static initializers, might register Service::Metadata structures. As
                         // the dlopen has a process wide system lock, make sure that the, during open used lock of the
@@ -1467,41 +1473,18 @@ namespace PluginHost {
                         Core::Library newLib(element.c_str());
 
                         if (newLib.IsLoaded() == true) {
-                            if (progressedState == 1) {
-                                progressedState = 2;
-                            }
 
-                            Core::System::ModuleBuildRefImpl moduleBuildRef = reinterpret_cast<Core::System::ModuleBuildRefImpl>(newLib.LoadFunction(_T("ModuleBuildRef")));
                             Core::System::GetModuleServicesImpl moduleServiceMetadata = reinterpret_cast<Core::System::GetModuleServicesImpl>(newLib.LoadFunction(_T("GetModuleServices")));
-                            if ((moduleBuildRef != nullptr) && (moduleServiceMetadata != nullptr)) {
-                                result = newLib;
-                                progressedState = 3;
-                                // DEPRECATED: In the next release cleanup the legacy, where the extended interface was [optional] it should be mandatory!
-                                if ((_metadata.IsValid() == false) || (_metadata.IsExtended() == false)) {
-                                    _metadata = moduleServiceMetadata();
-                                    if (_metadata.IsValid() == true) {
-                                        _precondition.Load(_metadata.Precondition());
-                                        _termination.Load(_metadata.Termination());
-                                    }
-                                    _metadata.Hash(moduleBuildRef());
+                            if (moduleServiceMetadata != nullptr) {
+                                result = moduleServiceMetadata();
+                                if (result != nullptr) {
+                                    library = std::move(newLib);
                                 }
                             }
                         }
                     }
                 }
                 all_paths->Release();
-
-                if (HasError() == false) {
-                    if (progressedState == 0) {
-                        ErrorMessage(_T("library does not exist"));
-                    }
-                    else if (progressedState == 1) {
-                        ErrorMessage(_T("library could not be loaded"));
-                    }
-                    else if (progressedState == 2) {
-                        ErrorMessage(_T("library does not contain the right methods"));
-                    }
-                }
 
                 return (result);
             }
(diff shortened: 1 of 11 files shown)


pwielders:
@sebaszm / @VeithMetro 2 interesting "ohh shit erlebnissen" both where concerning templates (fun but a pain in the ...). See: line (335) in Services.cpp and lines (386-394) in metadata.h. They where the result of tedious template hassles....

pwielders:
ohhh b.t.w testing is in progress ;-) so only "compiles on linux "

Source/core/Library.h
```diff
@@ -41,6 +49,98 @@ namespace Core {
...
+                _current++;
+                // See if we can find a name..
+                if ((_current != static_cast<uint32_t>(~0)) && (LoadIndex(_current) == true)) {
+                    return (true);
+                }
+                _current = static_cast<uint32_t>(~0);
+                return (false);
+            }
+            Library Current() {
+                ASSERT(IsValid() == true)
```
VeithMetro: Semicolon is MIA

Source/core/Library.h
```diff
@@ -41,6 +49,98 @@ namespace Core {
...
+                    if (index < (needed / sizeof(HANDLE))) { 
+                        TCHAR moduleName[MAX_PATH];
+
+                        if (::GetModuleFileNameEx(GetCurrentProcess(), handles[current], moduleName, sizeof(moduleName)/sizeof(TCHAR))) {
+                            _filename = (moduleName);
+                            loaded = true;
+                        }
+                    }
+                }
+                return (loaded)
```
VeithMetro: Same here 😄

Source/core/Library.h
```diff
@@ -41,6 +49,98 @@ namespace Core {
...
+                _current = static_cast<uint32_t>(~0);
+                return (false);
+            }
+            Library Current() {
+                ASSERT(IsValid() == true)
+                return (Library(_filename.c_str()));
+            }
+
+        private:
+            // Index is the inde counted from 1 up.. (0 is an illegal index as it means no handle)
```
VeithMetro: "Index is counted from 1 up…"

Source/core/Library.h
```diff
@@ -41,6 +49,98 @@ namespace Core {
...
+            bool LoadIndex(const uint32_t index) {
+                bool loaded (false);
+                ASSERT (index > 0);
+                HMODULE* handles = reinterpret_cast<HMODULE*>(ALLOCA(sizeof(HMODULE) * index));
+                DWORD    needed;
+                if (::EnumProcessModules(GetCurrentProcess(), handles, sizeof(HMODULE) * index, &needed)) {
+                    if (index < (needed / sizeof(HANDLE))) { 
+                        TCHAR moduleName[MAX_PATH];
+
+                        if (::GetModuleFileNameEx(GetCurrentProcess(), handles[current], moduleName, sizeof(moduleName)/sizeof(TCHAR))) {
```
VeithMetro: `current` without `_`

Source/core/Services.cpp
```diff
@@ -96,28 +71,28 @@ namespace Core {
...
 
     void ServiceAdministrator::FlushLibraries()
     {
-        _dynamicLoaderLock.Lock();
+        _adminLock.Lock();
         while (_unreferencedLibraries.size() != 0) {
             Library lib = _unreferencedLibraries.front();
-            _unreferencedLibraries.pop_front();
-            _dynamicLoaderLock.Unlock();
+            _unreferencedLibraries.pop_back();
```
VeithMetro, line 72: Why change to `pop_back()` here?

Source/core/Services.h
```diff
@@ -56,47 +59,39 @@ namespace Core {
...
         ServiceAdministrator& operator=(const ServiceAdministrator&) = delete;
-        virtual ~ServiceAdministrator();
+        virtual ~ServiceAdministrator() = default;
 
         static ServiceAdministrator& Instance();
 
     public:
         Library LoadLibrary(const TCHAR libraryName[]) {
-            Core::SafeSyncType<Core::CriticalSection> lock(_dynamicLoaderLock);
+            Core::SafeSyncType<Core::CriticalSection> lock(_adminLock);
```
VeithMetro: I thought with these changes, we would be able to get away without locking in both open and close?

VeithMetro (commented):
Some adjustments and a few questions

Source/core/Services.cpp
pwielders: I changed from std::list to std::vector and std:vector has no pop_front.

Source/core/Services.h
pwielders: The "release" of the library should still be done on the idle thread. This lock, however, is now never causing the ABBA lock scenario describe in the text here as on the create the lock is nologer needed. Here it is no use to only protect the std::vector to move the Core::Library references from one thread to the destructor thread.

Source/core/Library.h
pwielders: I guess you figured out by now that I did not build on windows (Release and Debug) and on linux only built the Release :-( I thought I could get away with it :-)

Source/core/Library.h
pwielders: added

Source/core/Library.h
pwielders: added

Source/core/Library.h
pwielders: Adapted

Source/core/Services.cpp
VeithMetro: In that case, don't we need to do `_unreferencedLibraries.back();` one line up?

pwielders: good catch ! Indeed ...

Source/Thunder/PluginServer.h
```diff
@@ -712,22 +704,31 @@ namespace PluginHost {
                 }
                 ~ControlData() = default;
 
-                ControlData& operator=(const Core::IService::IMetadata* info) {
+                ControlData& operator=(Core::IService* info) {
```
sebaszm: Why not const? this suggests that rhs is potentially modified by operator=

Source/core/Services.cpp
```diff
@@ -96,28 +67,28 @@ namespace Core {
...
 
     void ServiceAdministrator::FlushLibraries()
     {
-        _dynamicLoaderLock.Lock();
+        _adminLock.Lock();
         while (_unreferencedLibraries.size() != 0) {
-            Library lib = _unreferencedLibraries.front();
-            _unreferencedLibraries.pop_front();
-            _dynamicLoaderLock.Unlock();
+            Library lib = _unreferencedLibraries.back();
```
sebaszm, line 71: dlclose order is now changed from FIFO to LIFO. Wouldn't this be problematic for libraries with complex dependencies as dlcose's will not be in reverse order of dlopen's?

pwielders: Fair point but I think that this is an issue for the dl application itself as this could potentially also happen if we would have a multithreaded application where one thread could race another thread in releasing a library.
If we want to be safe i.s.o. sorry we can change the removal of the library from the list, using the the _unreferencedLibraries.erase(_unreferencedLibraries.begin()), but I guess that will be a more expensive operation...

Source/Thunder/PluginServer.h
pwielders: Could be done but it requires interfaces changes on IService:
Option1:
```
        virtual void* Create(const uint32_t interfaceNumber) = 0;
        virtual const IMetadata* Metadata() const = 0;
        virtual IService* Next() = 0;
        virtual const IService* Next() const = 0; // <Add
```

Option2:
```
        virtual void* Create(const uint32_t interfaceNumber) const = 0; // < make the create const
        virtual const IMetadata* Metadata() const = 0;
        virtual const IService* Next() const = 0; // Make Next const
```

Prefer Option 2, as option 1 complicates the interface for a technical issue, questionable if the Create is a cons method :-)

Lets do a vote, @sebaszm / @VeithMetro

sebaszm: 2!

VeithMetro: Definitely the second interface looks much cleaner

pwielders: Addressed!

Source/core/Services.cpp
pwielders: @sebaszm / @VeithMetro  vote again?
1) Use ```_unreferencedLibraries.erase(_unreferencedLibraries.begin())``` to go from front to back
2) Pray and hope the the linux dl handles proper unloading :-)

sebaszm: I think it can only burn if one lib calls the other during dlclose (eg static destructors etc) then the order would matter. DL cannot do anything about it. It's theoretically possible, but I think I still wouldn't do 1) just to save potentially dreadful plugin code.

pwielders: @sebaszm  so you opt for 2? @VeithMetro  ?

VeithMetro: What about using deque instead of vector, and then we can still have FIFO?

pwielders: After some discussion we opted for sticking to the std::vector and @pwielders  will create a testapp to see what really happens. If unloading is than an issue, we will use the deque here.

VeithMetro (approved):

pwielders:
@volkan-aslan  can you add your observations from QA perspective.

volkan-aslan:
I have done 200 reboot iterations and I did not observe the issue. So, LGTM :)

sebaszm (approved):

PR #1944, opened 2025-09-30
To handle truncated UTF code on parsing empty null terminated string

sivarajappan-siva:


sebaszm (commented):
codeSize is negative, so the condition is always false (implicitly converted to unsigned int)

Source/core/JSON.h
```diff
@@ -1867,7 +1867,11 @@ namespace Core {
                                 if (codeSize < 0) {
                                     // Oops it is a bad code thingy, Skip it..
                                     // TODO: report an error
-                                    codeSize = -codeSize;
+                                    if(::abs(codeSize) <= length) {
```
sebaszm: We already know codeSize is negative here, so use of ::abs() is superfluous. Also I think you need to cast it, so the comparison is not mixing signed and unsigned integers, you may get warnings otherwise.

sebaszm (changes requested):
We already know codd

Source/core/JSON.h
sivarajappan-siva: I have updated the code by doing cast now.Thanks

sebaszm (approved):

PR #1971, opened 2025-10-27
Mark end of string for an unquoted input

sivarajappan-siva:


MFransen69:
Not needed anymore, fixed in another way

PR #1972, opened 2025-10-28
[METADATADISCOVERY] Make the Metadata discover configurable.

pwielders:
For automatic configuration of the subsystems and plugin version retrieval, thunder loads all plugins into memory at startup without starting them. This allows for a proper build up of the subsystem dependencies (read from the build in code of the plugin) and allows for retrieval of plugin versions, without the plugins being activated. This feature however, moves the additional loading time of each plugin, into the Thunder startup timer. If squash file system (assumption this is the root cause) is used and you have a great deal of plugins that are not activated at startup, increase the Thunder Plugin time by 6 Seconds. Measurements using an ext4 file system show (using 150 reference stack plugins) that the additional loading time is increased by 100mS.

To reduce the startup time of the software it was requested by Comcast to make this feature configurable. With this PR, there is a flag in the config called "discovery". By default it is true Comcast can, in their build, configure this to false. If configured to false, the plugins are no longer pre-loaded to extract the metadata from the plugins, effectively reducing the initial startup time of Thunder.

Note: If disabled, the subsystems functionality, is no longer taken from the plugin and if needed
      must be placed in the plugin config, manual labor, error prone but still available.
      Versions of plugins will not give any output as long as the plugin is not loaded, at least
      once. So the plugin must be at least started once to get this version info.
      The startup time is still required at first startup, so the feature only "moves" the startup
      time from Thunder to whenever the plugin is activated for the first time!

      Bottom line: From Thunder perspective, it is *not* recommended to turn off the "discovery"!

### DIFF
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -922,20 +922,22 @@ namespace PluginHost {
     // -----------------------------------------------------------------------------------------------------------------------------------
     void Server::ServiceMap::Open(std::vector<PluginHost::ISubSystem::subsystem>& externallyControlled) {
         // Load the metadata for the subsystem information..
-        for (auto service : _services)
-        {
-            service.second->LoadMetadata();
-            for (const PluginHost::ISubSystem::subsystem& entry : service.second->SubSystemControl()) {
-                Core::EnumerateType<PluginHost::ISubSystem::subsystem> name(entry);
-                if (std::find(externallyControlled.begin(), externallyControlled.end(), entry) != externallyControlled.end()) {
-                    SYSLOG(Logging::Startup, (Core::Format(_T("Subsystem [%s] controlled by multiple plugins. Second: [%s]. Configuration error!!!"), name.Data(), service.second->Callsign().c_str())));
-                }
-                else if (entry >= PluginHost::ISubSystem::END_LIST) {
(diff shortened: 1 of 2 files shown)


sebaszm (dismissed):

Source/Thunder/PluginServer.cpp
```diff
@@ -922,20 +922,22 @@ namespace PluginHost {
...
-                if (std::find(externallyControlled.begin(), externallyControlled.end(), entry) != externallyControlled.end()) {
-                    SYSLOG(Logging::Startup, (Core::Format(_T("Subsystem [%s] controlled by multiple plugins. Second: [%s]. Configuration error!!!"), name.Data(), service.second->Callsign().c_str())));
-                }
-                else if (entry >= PluginHost::ISubSystem::END_LIST) {
-                    SYSLOG(Logging::Startup, (Core::Format(_T("Subsystem [%s] can not be used as a control value in [%s]!!!"), name.Data(), service.second->Callsign().c_str())));
-                }
-                else {
-                    SYSLOG(Logging::Startup, (Core::Format(_T("Subsytem [%s] controlled by plugin [%s]"), name.Data(), service.second->Callsign().c_str())));
-                    externallyControlled.emplace_back(entry);
+        if (Configuration().MetadataDiscovery() == true) {
```
MFransen69: Add an else where we also syslog the feature is turned off to help in diagnosing issues because of this?

MFransen69 (changes requested):

MFransen69 (approved):

sebaszm (approved):

PR #1987, opened 2025-11-03
[Priority Queue] Distinguish the source of ThreadPool to bring the ASSERT back

VeithMetro:


sebaszm (dismissed):

Source/ThunderPlugin/Process.cpp
```diff
@@ -111,8 +111,8 @@ POP_WARNING()
         WorkerPoolImplementation& operator=(const WorkerPoolImplementation&) = delete;
 
 PUSH_WARNING(DISABLE_WARNING_THIS_IN_MEMBER_INITIALIZER_LIST)
-        WorkerPoolImplementation(const uint8_t threads, const uint32_t stackSize, const uint32_t queueSize, const string& callsign)
-            : WorkerPool(threads - 1, stackSize, queueSize, &_dispatcher, this, threads, threads)
+        WorkerPoolImplementation(const uint8_t threads, const uint32_t stackSize, const uint32_t queueSize, const string& callsign, const uint8_t additionalThreads = 1)
+            : WorkerPool(threads - 1, stackSize, queueSize, &_dispatcher, this, (threads > 2 ? (threads - 1) : threads), (threads > 2 ? (threads - 1) : threads), additionalThreads)
```
MFransen69: I know additionalThreads will most probably never be different than one but wouldn't it be tetter to also change threads - 1 in to threads - additionalThreads (people might wonder why they behave differently)

MFransen69 (changes requested):

Source/ThunderPlugin/Process.cpp
VeithMetro: @MFransen69 I see your point, but we are specifically subtracting one to have just one thread for high priority jobs (if there are 3 threads or more in total). I don't think this is connected to the fact that in case of ThunderPlugin, we have the main thread which joins in the ThreadPool

MFransen69: Sorry I was not clear I see now, I meant the first one threads - 1 which goes for the total # workerpool threads
I will call you that is easier

MFransen69 (approved):

PR #1989, opened 2025-11-03
[Tests/unit/core] : pass a proper length for 'TriState' object creati…

msieben:
…on from strings


MFransen69:
After discussing with Mischa we concluded although indeed things are fishy here we need to fix the TriState and not necessarily the test

PR #2002, opened 2025-11-12
[RACECONDITION] Racecondition on destruction of a socket if the Close…

pwielders:
…d() was executed.

If the Closed() method, was run, it would set the m_State = 0, which meant that the SocketPort class was classed and eligable for destruction. In a race condition, this signalling allowed a server to "cleanup" the children (Clients). In the unlucky case that after the setting of the:

[ReourceMonitorThread]
SocketPort::Closed() {
    ...
    m_State -> 0
    ...
}
                                             [Server Cleanup Thread]
                                             Delete all Clients that are Closed()
SocketPort::Events() {
    ...
    m_State &= ~SocketPort::MONITOR;  CRASH!
    ...
}

This PR prevents this race condition from happening.

As this is in the core of Thunder, the biggest risk is the Lock in the destructor. It might casue (if the socket is used incorrectly) to ABBA locks. Request thorough testing !

### DIFF
--- a/Source/core/SocketPort.cpp
+++ b/Source/core/SocketPort.cpp
@@ -423,6 +423,14 @@ namespace Thunder {
         {
             TRACE_L5("Destructor SocketPort <%p>", (this));
 
+            // The Closed(), internal method, run on the ResourceMonitor thread,
+            // will lock and clear the m_State if applicable. We want to make sure
+            // that we can take the lock again, to make sure that the Closed() 
+            // released it. All methods, calling the Closed(), will nolonger refer 
+            // to any members or methods in this class, so it is safe to destruct
+            // after we have aquired the lock!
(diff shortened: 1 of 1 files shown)


pwielders:
@Karthick-Somasundaresan / @sramani-metro as this is your analyses, it would be good to review this thoroughly if this fixes your use case. It is a theoretical fix, the proof of the pudding is in the eating :-) 

@MFransen69 might be interesting for the 5.3 release. It is already in I think from the beginning, as you must be very unlucky to run into it but I guess @Karthick-Somasundaresan / @sramani-metro did have some BackTraces pointing to this issue (I think)

pwielders:
Closed. Follow up is: https://github.com/rdkcentral/Thunder/pull/2004

PR #2005, opened 2025-11-15
Add CustomCode feature

MFransen69:
[DependsOn=ThunderTools:development/MoveSomeFrameCodeToNumbers]

Add support for Custom (Error) Codes next to the Thunder predefined ones

Note: will be in COMPLEMENTARY_CODE_SET section

Fully backwards compatible.
Tested both on Linux and Windows
Example of feature added to ThunderNanoServices ()

Note related ThunderTools PR: https://github.com/rdkcentral/ThunderTools/pull/210

### DIFF
--- a/Source/core/Number.h
+++ b/Source/core/Number.h
@@ -1274,6 +1274,9 @@ public:
 
 } // namespace std
 
+using uint24_t = Thunder::Core::UInt24;
+using int24_t = Thunder::Core::SInt24;
+
 template <typename NEW_TYPE, typename ORIGINAL_TYPE>
 NEW_TYPE int_cast(const ORIGINAL_TYPE& input)
 {
--- a/Source/core/Number.h
+++ b/Source/core/Number.h
@@ -25,6 +25,8 @@
 #include "TextFragment.h"
 #include "TypeTraits.h"
 
+#include <limits>
+
 namespace Thunder {
 namespace Core {
     extern "C" {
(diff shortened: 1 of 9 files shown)


Source/core/Number.h
```diff
@@ -1049,15 +1051,243 @@ namespace Core {
...
+        }
+
+        return success;
+
+    }
+
 } // namespace Core
+} // namespace Thunder
+
+namespace std { // seems to be allowed/mandatory to specialize inside the std namespace
```
mhughesacn, line 1179: Hi @MFransen69 : Blackduck is finding a code  match here from lines 1185-1273, to someone else's Template structure numeric_limits.  But it looks like standard setup for various data types SInt24, UInt24.  Please will you confirm that these lines are simply setup that must be done in this manner, and should not be considered to be copied code.

MFransen69: Hi @mhughesacn 

Yes, it is indeed logical it finds matching code is it is a specialization of a C++ standard template. It is  the only way it should be done to add this functionality for your own defined type, which SIn24 and UInt24 are, they are defined in Thunder code (above).
So I can confirm these lines need to be done this way and should not be considered copied code

pwielders (dismissed):

mhughesacn:
Hi @MFransen69 : Thank you for your response.  I just cleared off the BD complaint.

VeithMetro (approved):

PR #2008, opened 2025-11-18
Development/custom codes2

MFransen69:
Now full support for uint24_t and in24_t and Custom Code uses this now (of course still inside non complimentary code section :) )
To be able to do this I had to split off the error code from Portability as the U/SInt24 in its separate file still needed ASSERT.

So interface better indicates that only 24bits types are used where and overflow situation (Release, debug will still assert) no longer needs to be checked by a certain magic value externally which is much better but with a method now, added a templated version just like RealSize in case it might be needed more  types (e.g. in23_t :) ) in the future in some templated code (also abstracts away the need to call a specific method) (btw did not also make RealSize use the sfinae check to see if the method is there for a reason)

BTW Built for both Linux and Windows, for now only tested on Windows as the Linux lithospere update broke my build and I'll fix that tomorrow (Today that is ;) ). But there is nothing Linux specific on it so it should work anyway.

@mhughesacn I had to move the template code that gave the false warning a few days ago to a new file so you might see the same black duck error again, but it is the same code so again it is just a template specialization for our own types and needs to be done this way so it is a false warning.

### DIFF
--- a/Source/core/AccessControl.h
+++ b/Source/core/AccessControl.h
@@ -22,6 +22,7 @@
 #include "Module.h"
 #include "NodeId.h"
 #include "Portability.h"
+#include "Errors.h"
 
 #ifdef __POSIX__
 #include <grp.h>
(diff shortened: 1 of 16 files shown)


pwielders (approved):

sebaszm (approved):

mhughesacn:
Thank you @MFransen69 - Clearing this

PR #2015, opened 2025-11-23
Development/53releasenotes

MFransen69:


docs/introduction/config.md
```diff
@@ -34,6 +34,7 @@ This section documents the available options for Thunder. This is different from
...
 | redirect                          | Redirect incoming HTTP requests to the root Thunder URL to this address (please note it must contain the resource that is required e.g. index.html ) | string    | http://127.0.0.1/Service/Controller/UI/index.html            | http://127.0.0.1/Service/Controller/UI/index.html     |
 | idletime                          | Amount of time (in seconds) to wait before closing and cleaning up idle client connections. If no activity occurs over a connection for this time Thunder will close it. | integer   | 180                                                          | 180                                                   |
+| discovery                         | enable loading of the plugin metadata on Thuder startup, when disabled metadata will only be available after plugin activation (only turn this feature off when loading the metadata does not work in special circumstances) | bool   | true                                                            |
```
nxtum: spelt Thuder

nxtum (dismissed):

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+# Thunder 
+
+##  Process Changes and new Features 
+
+### Feature: Windows build artifacts cleanup
+
+The Windows build (Visual Studio Projects) now fully cleanup the build artifacts when the solution is cleaned guaranteeing a complete fresh build after cleaning.
+
+### Feature: Scripts before thunder start in CMake
```
nxtum: thunder not capitalised

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+A feature was added to the CMake files to enable the inclusion of scripts to be executed before Thunder startup (Linux).
```
nxtum: Think this might be a little vague, I don't understand what it really means

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: action build with Performance Monitoring feature aan
```
nxtum: spelling mistake "aan"

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: Copilot scanning enabled and issues fixed
+
+Next to Coverity also Copilot is now enabled for all Thunder repositories to check the pull requests for issues. The few remarks found when enabled have been cleared up.
+
+## Changes and new Features
+
+## Feature: WorkerPool PriorityQueue
+
+The Thunder WorkerPool (and therefore the ThreadPool as well) now has the ability to use priorities for its jobs. Three priorities ara available, high medium and low.
```
nxtum: Three priorities are

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+A feature was added to the CMake files to enable the inclusion of scripts to be executed before Thunder startup (Linux).
+
+### Feature: action build with Performance Monitoring feature aan
+
+The GitHub actions are extended with a build where the Performance Monitoring feature (enables code that will enable a special plugin to measure JSON-RPC performance) is turned on to verify it is not broken by a pull request.
+
+### Feature: Copilot scanning enabled and issues fixed
+
+Next to Coverity also Copilot is now enabled for all Thunder repositories to check the pull requests for issues. The few remarks found when enabled have been cleared up.
```
nxtum: maybe would reword this as:

Copilot is now enabled for all Thunder repositories alongside Coverity, and the initial reported issues have been addressed.

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+## Changes and new Features
+
+## Feature: WorkerPool PriorityQueue
+
+The Thunder WorkerPool (and therefore the ThreadPool as well) now has the ability to use priorities for its jobs. Three priorities ara available, high medium and low. 
+In the Thunder config the total number of WorkerPool threads used by THunder itself, as well as the thresholds of maximum parallel scheduled jobs for medium and low priority jobs can be configured (note if not overridden here the already existing THREADPOOL_COUNT is still taken into account).
```
nxtum: Thunder spelt as THunder

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+See more info [here](https://rdkcentral.github.io/Thunder/introduction/config/).
+
+The scheduling algorithm used by the Thunder WorkerPool/ThreadPool can be compile time configured to be either static or dynamic. With static only the maximum number of jobs for that priority will be scheduled in parallel, with dynamic additionally the total load of the system is also taken into account.
+When both static and dynamic are (compile time) compile time disabled the previous non priority solution will be used.
```
nxtum: compile time repeated

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+For the previous it is advised to always configure the lower priority threshold to at least one lower than the total amount of worker pool threads to guarantee at least one thread always to be available for the high priority jobs.
+In case of an incorrect configuration, e.g. medium and/or low thresholds higher than the total amount of threads will give an ASSERT.
```
nxtum: maybe trigger an ASSERT

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+As this a relatively new but potentially complex feature to Thunder the Thunder documentation will be extended with a more detailed description than the above.
```
nxtum: missing is after this

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: Custom (Error) Codes.
+
+Before Thunder 5.3 error codes used in Thunder were predefined, which sufficed up to now. These error codes (due to the "JSON-RPC in terms of COM-RPC" feature) are also translated into JSON-RPC error codes and there was a request to support a more flexible error scheme (mainly to have more flexibility in the error codes reported in the JSON-RPC error object).
+So with this new Custom Codes a mechanism was introduced to have custom error codes that are consistent for both COM-RPC and JSON-RPC (or any other future protocol that will be implemented in terns of COM-RPC), allow for custom (not hardcoded inside Thunder) code to string translation and allow for direct influence on the error code returned by JSON-RPC.
```
nxtum: With these new Custom Codes,

terms spelt as terns

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+We now have added a WarningReporting message when the FlowControl feature (see above and in the Thunder 5.1 release notes) leads to longer than configured handling times.
+
+### Change: COMRPCTerminator thread only started when needed
+
+The COM-RPC terminator thread (a cleanup thread used within Thunder) is now only started when required, leading to one less thread in processes that so not need it (like ThunderPlugin)
+
+### Feature: allow per callsign registration IShell
+
+It is now possible when registering via COM-RPC on the IShell interface to be notified on plugin state changes to be notified only for the state changes of a specific plugin instead of changes for all plugins by optionally providing the callsign of the plugin for which the state changes should be sent.
```
nxtum: Think i would change wording as to not repeated to be notified twice.

"When registering for plugin state notifications on the IShell interface over COM-RPC, it is now possible to subscribe to state changes for a single plugin by optionally providing its callsign, instead of receiving notifications for all plugins."

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: metadata loading configurable
+
+One can now disable the loading of Plugin metadata (plugin version, subsystem being controlled and dependencies) with the Thunder configuration flag "discovery". With this flag set to true (default) means loading the metadata is enabled.
+When disabled this means this information will not be loaded before the plugin is actually activated making some features of Thunder not work correctly.
+This feature on request as on some devices just loading the library to get this information took much longer than in normal situations (so only turn this flag to false in special situations). 
+
+### Feature: Forgiving JSON-RPC method Pascal/CamelCase handling
+
+A build option was introduced to enable "forgiving" JSON-RPC method casing handling in Thunder. When turned on, default is off, Thunder will accept the method name both as camelCase and PascalCase (to be more forgiving as method name casing was not always done consistently in interfaces).
```
nxtum: would probably put default is off into ()

docs/plugin/interfaces/tags.md
```diff
@@ -143,7 +143,7 @@ When the function returns, the parameter will have the modified length value. Th
 
 #### @restrict
 
-Specifies a valid range for a parameter (e.g. for buffers and strings it could specify a valid size). Ranges are inclusive.
+Specifies a valid range for a parameter (e.g. for arrays, std::vector, iterators and strings it could specify a valid size). Ranges are inclusive.
```
VeithMetro: I think this sounds a bit better: `(e.g. it can indicate a valid size for buffers, strings, arrays, std::vector, or iterators)`

docs/plugin/interfaces/tags.md
```diff
@@ -611,6 +619,64 @@ Example list:
 
 <hr>
 
+#### Wrapped
+
+This tag can be placed at class or method level, where at class level is by far preferable as it prevents inconsistencies in JSON-RPC function handling.
+Wrapped will for a single output parameter also add the parameter name to the result, making it always a JSON object.
+Note can also be used for array, std::vector, iterator, POD etc. single output parameter.
+Incorrect or inconsistent usage will lead to an error to be generated by the code generators. Of course the documentation generators do take the wrapped tag into account.
```
VeithMetro: Perhaps let's use “raise an error” instead of “generate”:
_“Incorrect or inconsistent usage will lead to an error raised by the code generators.”_

docs/plugin/interfaces/tags.md
```diff
@@ -399,7 +400,11 @@ Indicates that enumerator lists should be packed into into a bit mask.
 #### @index
 Used in conjunction with @property. Allows a property list to be accessed at a given index.
 
-Index should be the first parameter in the function. 
+Index should be the first parameter in the function. It is allowed to make the index a Core::OptionalType<>.
+
+The index tag can also be allowed to indicate an event uses an index parameter so clients can subscribe to an event only for a certain value of that index: for more info see [here](../interfaces/#indexed-events)
```
VeithMetro: I think that sounds a bit more clear:
_"The @index tag can be used to mark an event parameter, so clients can subscribe to the event for a specific index value"_

docs/plugin/interfaces/tags.md
```diff
@@ -399,7 +400,11 @@ Indicates that enumerator lists should be packed into into a bit mask.
...
+
+As for a short amount of time the @index for events used a dot (.) as a separator in the clientid instead of a @ in the designator, @index:deprecated can be put at the indexed event parameter to indicate this one should use the old deprecated index format (see the generated documentation for what the interface would be like but only use this for backwards compatibility reasons).
```
VeithMetro: What about something like this?
_"For backward compatibility, use @index:deprecated on the indexed event parameter to indicate the old designator format (dot '.' in the client ID) should be used."_
I think it carries the same messages, but it is a bit more confined and clear

docs/plugin/interfaces/interfaces.md
```diff
@@ -133,13 +161,241 @@ Examples:
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 42,
+  "result": null
+}
+```
+
+Unregistrering from an even can be done by the client by calling the "unregister"" function, which like the register is generically available for all plugins:
```
VeithMetro: _"Unregistering"_ and _"event"_ typos, and also let's remove one unnecessary quote from _"unregister""_

docs/plugin/interfaces/interfaces.md
```diff
@@ -133,13 +161,241 @@ Examples:
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 42,
+  "method": "MyPLugin.1.unregister",
```
VeithMetro: _"MyPlugin"_

docs/plugin/interfaces/interfaces.md
```diff
@@ -133,13 +161,241 @@ Examples:
...
+
+Note depending on the interface more than one helper function is generated (e.g. also one where you can use the generated params class), check the generated J< interface >.h file for all the options.
+
+##### indexed events
+
+With the above and the generated event notification functions the notification will be sent to all registered clients. It is possible to only notify some clients register for an event.
+If one wants to send some notifications only to some registered clients there must be a parameter in the event that a client uses to indicate if it wants to only receive the notification when that parameter has a certain value.
+That parameter can be flagged with the @index tag to indicate it is used to discriminate registered clients.
+
+Let's discus with an example:
```
VeithMetro: _"discuss"_

docs/plugin/interfaces/interfaces.md
```diff
@@ -133,13 +161,241 @@ Examples:
...
+        };
+
+        virtual Core::hresult Register(INotification* const sink, const string& callsign /* index */) = 0;
+        virtual Core::hresult Unregister(const INotification* const sink, const string& callsign /* index */) = 0;
+
+        virtual Core::hresult Test() const = 0;
+    };
+```
+
+From the COM-RPC interface it is clear one can subscribe to be notified on changes for a specific callsign. The @index is put at the COM-RPC Register and Unregister methods, the code generators will do a name lookup up, so in this case "callsign" to find the connected event.
```
VeithMetro: _"name lookup up"_ → _"name lookup"_

docs/plugin/interfaces/interfaces.md
```diff
@@ -204,7 +460,7 @@ For a more detailed view, visit [Messenger.h](https://github.com/WebPlatformForE
 Object lookup defines the ability to create a JSON-RPC interface to access dynamically created objects (or sessions).
 This object interface is brought into JSON-RPC scope with a object ID. This translates the Object Oriented domain (used in COM-RPC interfaces) to the functional domain (JSON-RPC).
 
-Object lookup will happen automatically by the generator when a method is found on the COM-RPC interface that returns an COM-RPC Interface also tagged as JSON-RPC interface as an out parameter and it is expected also a method that takes the same interface as input parameter is available to be able to destroy the created object.
+Object lookup will happen automatically when the @encode:autolookup tag is added to the interface that is to be the session object. The the generator will look for method on that returns that session type as an out parameter as COM-RPC Interface (on an interface in the same file which of course also must be tagged as JSON-RPC interface) and it is expected that there is a method that takes the same interface as input parameter to be available as well to be able to destroy the created object.
```
VeithMetro: _"The the generator"_ → _"The generator"_
And perhaps paraphrasing the whole sentence like that would sound a bit better:
_"The generator will look for methods that return the session type as an out parameter (a COM-RPC interface, which has to be in the same file and must have the JSON-RPC tag) and a corresponding method that accepts the same interface to destroy the created object."_

docs/plugin/interfaces/interfaces.md
```diff
@@ -133,13 +161,241 @@ Examples:
...
+It will now always include the callsign as parameter (as if the event was for a client that registered for all callsigns the @ part will not be in the event so the callsign must be included as parameter to indicate for which callsign this event is).
+
+As always these examples will also be included in the generated documentation for this interface
+
+And again static helper functions will be included in the generated J< interface >.h file for sending the notifications.
+
+##### statuslisteners
+
 Tagging a notification with @statuslistener will emit additional code that will allow you to be notified when a JSON-RPC client has registered (or unregistered) from this notification. As a result, an additional IHandler interface is generated, providing the callbacks.
+Note the unregistered notification will also be triggered when it is the result of the channel closing on which the client registered without deregistering explicitly beforehand.
```
VeithMetro: Perhaps something like this instead?
_"Note: the 'unregistered' notification is also triggered if the client disconnects (channel closed) without explicitly calling Unregister."_

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+    This file contains the interface that a library can implement in case the "custom error codes" feature is used in Thunder and code to string comversion is desired
```
VeithMetro: _"comversion"_ → _"conversion"_

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+As the range for the custom code is too big to have pre created static texts for all of them, it is desirable to create a dynamic string representation when a static one is not provided by the conversion library (or the conversion library feature is not used at all) to include the actual custom code in the message.
+This cannot be done in the ErrorToString as it returns a pointer to the text, therefore an extended version is created that will include the custom code number.
+(changing the ErrorToString was not desirable as that would break backwards compatibility unnecessarily)
+
+## String representation
+
+To be able to provide for a custom error code to string conversion (e.g. used for the JSON-RPC error text in the error object or to return a correct string representation when Core::ErrorToString is called) the following infrastructure will be added to Thunder:
+
+In Thunder sources/core a header file "ICustomErorCodes.h" provides the interface to be implemented by an external library to support the custom error codes used in the plugin code for that Thunder instance.
```
VeithMetro: _"ICustomErorCodes.h"_ → _"ICustomErrorCodes.h"_

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+
+* The custom error code will be placed as is in the JSON RPC error object "code" tag. (so code 0 means an invalid custom code (too big) was passed). Note it is the responsibility of developers setting the codes not to overlap with JSON RPC specification reserved codes or Thunder codes (1 to 100 in Thunder 4.4 or -31000 to -31999 in Thunder 5 and above) this on explicit request so that codes can be used for API backwards compatibility (a TRACE will be added to warn for these cases though)
+* the JSON-RPC error object "message" tag will contain the string returned by the CustomCodeToString call from the conversion library if configured and does return a text for that specific code, otherwise it will be set to "Undefined Custom Error: XXXX" where XXXX will be the Custom Code set. (of course when the "custom JSON-RPC error messages" feature was used for this call that will override this default behaviour)
+* if the value set with CustomCode() overflowed (too big for a 24bit number) and ASSERTS are not enabled ion the build the error code will be set to 0 and the object "message" tag will be set to "Invalid Custom ErrorCode set"
```
VeithMetro: _"ion"_ → _"in"_

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+} // extern "C"
+#endif 
+
+```
+
+It should suffice to have static strings which prevents unnecessary string copying and keeps the memory management simple.
+
+There is a Thunder config option called "customcodelibrary" that allows you to point Thunder to the library implementing the above interface.
+If it is not provided, or in case CustomCodeToString returns a nullptr for the particular code and a string representation for a custom error code is required the error code will be translated into "Undefined Custom Error: XXXX"  where XXXX is the custom error code.
+Note the existing ErrorToString will return a string without the number included when the string is not provided by the library as that cannot be supported without breaking backwards compatibility, ErrorToStringExtended will howeever (Thunder will of course use the new ErrorToStringExtended version to get the string that will be added to the JSON-RPC error text tag).
```
VeithMetro: _"howeever"_ → _"however"_

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+### Core::IsCustomCode
+
+With IsCustomCode one can find out if a Core::hresult is a custom code and what the code is:
+
+```cpp
+  int24_t IsCustomCode(const Core::hresult code);
+```
+
+ This will return the custom error code as signed 24 bit number when the "custom error bit" was set and 0 if the hresult was not a flexible error. 
+ (if the hresult code would have the value 0x1000000 ("custom code bit" set with code 0) IsCustomCode will return an in24_t with overflow value set which can be checked with Core::Overflowed, see the example below.
```
VeithMetro: _"an in24_t"_ → _"an int24_t"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+See [here](https://github.com/rdkcentral/Thunder/blob/master/ReleaseNotes/ThunderReleaseNotes_R5.2.md) for the release notes of Thunder R5.2.
+This document describes the changes in Thunder, ThunderTools and ThunderInterfaces as they are related. 
+
+# Thunder 
+
+##  Process Changes and new Features 
+
+### Feature: Windows build artifacts cleanup
+
+The Windows build (Visual Studio Projects) now fully cleanup the build artifacts when the solution is cleaned guaranteeing a complete fresh build after cleaning.
```
VeithMetro: _"cleanup"_ → _"cleans up"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
VeithMetro: Same here, perhaps something like this?
_"A feature was added to the CMake files to enable inclusion of scripts that are executed before Thunder starts (Linux)."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
VeithMetro: Probably meant _"on"_ 😄. Also, maybe let's change _"action"_ to _"Actions"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: Scripts before thunder start in CMake
+
+A feature was added to the CMake files to enable the inclusion of scripts to be executed before Thunder startup (Linux).
+
+### Feature: action build with Performance Monitoring feature aan
+
+The GitHub actions are extended with a build where the Performance Monitoring feature (enables code that will enable a special plugin to measure JSON-RPC performance) is turned on to verify it is not broken by a pull request.
```
VeithMetro: Maybe something like this would sound better?
_"The GitHub Actions workflow now includes a build with the Performance Monitoring feature enabled (this adds code necessary for a plugin that measures JSON‑RPC performance) so pull requests are checked for regressions in this area."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
VeithMetro: Probably meant:
_"When both static and dynamic are disabled at compile time, the previous non‑priority solution is used."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+When both static and dynamic are (compile time) compile time disabled the previous non priority solution will be used.
+
+Within Thunder itself at the moment only low and high priority jobs are used. When a COM-RPC request is received from a channel (so in an IPC situation, not in process) that has already an outgoing request, the job to handle the incoming request will be submitted with high priority, this to mitigate deadlocks that were seen previously when the thread pool was already filled with COM-RPC jobs that all required a call back into Thunder to be handled before they would be released (note any returning COM-RPC call will be elevated to high when there is already an outgoing call, not only the ones as the result from an outgoing call). Note: this will only mitigate this as one keeps calling (out of process) COM-RPC calls from the returning call depending on the amount of WorkerPool threads configured one might run out of threads to handle them at some point.
```
VeithMetro: To avoid one very long sentence, I would rewrite this section to something like this:
_"Currently, only low and high priority jobs are used. If a COM‑RPC request is received from a channel that already has an outgoing request (IPC, not in‑process), the job to handle the incoming request is submitted with high priority. This mitigates deadlocks when the thread pool is filled with COM‑RPC jobs that require callbacks into Thunder before they complete. Note: if returning COM‑RPC calls keep spawning out‑of‑process calls, you may still exhaust WorkerPool threads, depending on the configured thread count."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+For the previous it is advised to always configure the lower priority threshold to at least one lower than the total amount of worker pool threads to guarantee at least one thread always to be available for the high priority jobs.
```
VeithMetro: Maybe like this?
_"For this reason, it is advised to configure the low‑priority threshold to be at least one less than the total number of worker pool threads to ensure at least one thread is always available for high‑priority jobs."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
VeithMetro: Maybe let's also add a comma before the second part of this sentence (before "_the Thunder documentation will..."_), and we could get rid of _"to Thunder"_ as well

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+We now have added a WarningReporting message when the opening and/or closing of a WebSocket takes longer than an (configurable as always with WarningReporting) minimum amount of time.
+
+### Change: Open COM-RPC port later
+
+We moved the opening of the COM-RPC port to later in the startup of Thunder to prevent issues from clients trying to access it too soon.
+
+### Feature: make Thunder uClibc compatible
+
+Changes where made to the Thunder code to make it uClibc compatible.
```
VeithMetro: _"where"_ → _"were"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+https://github.com/rdkcentral/Thunder/commit/eea443cde3827ec3a44fcc8790176f39bf674737#diff-cb405958fdffcb85a8aeaa1bd1204df978e2525eb9bbab4db014266c39825d73
+
+### Feature: WarningReporting for JSON-RPC FlowControl
+
+We now have added a WarningReporting message when the FlowControl feature (see above and in the Thunder 5.1 release notes) leads to longer than configured handling times.
+
+### Change: COMRPCTerminator thread only started when needed
+
+The COM-RPC terminator thread (a cleanup thread used within Thunder) is now only started when required, leading to one less thread in processes that so not need it (like ThunderPlugin)
```
VeithMetro: _"so not need it"_ → _"do not need it"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+It is now possible when registering via COM-RPC on the IShell interface to be notified on plugin state changes to be notified only for the state changes of a specific plugin instead of changes for all plugins by optionally providing the callsign of the plugin for which the state changes should be sent.
+This change was made in a way that is backwards compatible with existing code.
+
+Note the applicable SmartInterfaceTypes (which use the IPlugin::INotification internally) have been adopted to use this feature reducing COM-RPC traffic and overhead.
+
+### Feature: allow per callsign state change notifications in IController and improvements
+
+The IController state notifications have been improved (so this also affects the JSON-RPC plugin state notifications)
+
+It is now possible to register only to be updated for the state changes of a specific plugin instead of state changes of all plugins, while the latter of of course is also still possible, by optionally providing the callsign of the plugin for which one wants to receive the notifications.
```
VeithMetro: _"of of course"_ → _"of course"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+This also applies to the JSON-RPC interface, for details here best to consult the generated documentation for the IController interface available in the Thunder repository.
+
+StateChange (which notifies of changes to plugin state that not includes Plugin::IStateControl) will when registering send a notification for all plugins already activated at that moment (when registering to be notified for all plugin state changes). If registering for a specific plugin it will sent a Activated notification in case that specific plugin was already activated at the moment of registering.
```
VeithMetro: _"that not includes"_ → _"that does not include"_
_"it will sent"_ → _"it will send"_
Also, perhaps let's add commas for better readability here:
_"will when registering send"_ → _"will, when registering, send"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+StateControlStateChange (which notifies of changes to plugin state from Plugin::IStateControl, so suspended and resumed) will when registering for notifications for a specific plugin immediately notify about the current state for that plugin and send no current state notification when registering for all plugins.
+
+### Feature: MAC Address class
+
+A class was added to Thunder core to represent a MAC Address (and as can be read below is also fully supported by the Thunder code generators)
```
VeithMetro: maybe _"as noted below"_ instead of _"as can be read below"_?

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: version handling for distributed plugins
+
+The plugin version handling now takes the Composite plugin into account. One can now use versioned methods also for the "composited" plugin by putting the version after the composite plugin delimiter.
+
+### Feature: metadata loading configurable
+
+One can now disable the loading of Plugin metadata (plugin version, subsystem being controlled and dependencies) with the Thunder configuration flag "discovery". With this flag set to true (default) means loading the metadata is enabled.
+When disabled this means this information will not be loaded before the plugin is actually activated making some features of Thunder not work correctly.
+This feature on request as on some devices just loading the library to get this information took much longer than in normal situations (so only turn this flag to false in special situations).
```
VeithMetro: I think this sounds better:
_"This feature was added on request because, on some devices, loading the library to read metadata took much longer than normal. Only set this flag to false in special circumstances."_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
VeithMetro: Me too, or put _"disabled by default"_ into ()

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+One can now disable the loading of Plugin metadata (plugin version, subsystem being controlled and dependencies) with the Thunder configuration flag "discovery". With this flag set to true (default) means loading the metadata is enabled.
+When disabled this means this information will not be loaded before the plugin is actually activated making some features of Thunder not work correctly.
+This feature on request as on some devices just loading the library to get this information took much longer than in normal situations (so only turn this flag to false in special situations). 
+
+### Feature: Forgiving JSON-RPC method Pascal/CamelCase handling
+
+A build option was introduced to enable "forgiving" JSON-RPC method casing handling in Thunder. When turned on, default is off, Thunder will accept the method name both as camelCase and PascalCase (to be more forgiving as method name casing was not always done consistently in interfaces).
+
+### Change: ICOMLink::INotifcation Revoked notification
```
VeithMetro: _"INotifcation"_ → _"INotification"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+When in the past the ICOMLink::INotifcation Revoked method was renamed to Dangling it was not removed or made deprecated and now might also confuse people with the Revoked notification when an offered interface is revoked from COM-RPC communicator.
```
VeithMetro: _"INotifcation"_ → _"INotification"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+It is now made deprecated and should in this context not be used anymore (it cannot yet be removed as that would break plugins that implement ICOMLink::INotifcation). When used make sure only Dangling is handled from this notification, Revoked will be removed in Thunder 6.
+
+### Feature: Syslog on COM-RPC timeout
+
+When a COM-RPC call is canceled because is takes longer than the configured COM-RPC timeout this is now logged as a Syslog message to help the investigation of situations where this happens.
```
VeithMetro: _"because is takes"_ → _"because it takes"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: Syslog on COM-RPC timeout
+
+When a COM-RPC call is canceled because is takes longer than the configured COM-RPC timeout this is now logged as a Syslog message to help the investigation of situations where this happens. 
+Note: in this case we now also close the connection as can be read above.
+
+### Change: Improvements to Library and Service Discovery handling
+
+The way Thunder loads the available services from Plugin libraries was improved in general (preventing deadlocks for example which were seen in rare cases by our QA department).
+This now also makes it possible to have services with a duplicate name which previously would have lead to issues.
```
VeithMetro: _"lead"_ → _"led"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+The reporting of the failure for a plugin to activate has been improved and an additional failure state was added called "INSTANTIATION_FAILED" if instantiating the plugin before the IPlugin::Initialize is called failed (which would be a very rare error).
+
+### Feature: Logging second invoke on COM-RPC channel
+
+Now a message is printed to Syslog when a second COM-RPC invocation is made to a channel warning about this possible deadlock (so in an IPC situation, not in process). This to help in the investigation of these plugin issues.
+
+### Feature: SinkType now has WaitReleased
+
+The SinkType now has a WaitReleased method. This can be used in the special situation when a Service is put into the SinkType but there can be a possible racecondion between closing the channel(s) from which the Service is used and receiving the actual Release call(s) from the using side(s).
```
VeithMetro: _"racecondion"_ → _"race condition"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+With WaitReleased the server side can now wait for the Release() to be received from the client side(s), with timeout. 
+See for a scenario in which this is useful [here](https://github.com/rdkcentral/ThunderPluginActivator/blob/89f789624305f4473fb5b605e8ae3cc4fa39fd15/source/COMRPCStarter.cpp#L94) and for the actual call to WaitReleased [here](https://github.com/rdkcentral/ThunderPluginActivator/blob/89f789624305f4473fb5b605e8ae3cc4fa39fd15/source/COMRPCStarter.cpp#L152). 
+
+### Change: Improved dead proxies handling
+
+Previously calling a COM-RPC method (over an IPC channel) from the (PluginServer) ICOMLink::INotification Dangling notification could lead to a possible deadlock this been fixed now.
```
VeithMetro: _"this been fixed"_ → _"this has been fixed"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+This means that all Dangling handling code (so both in PluginServer as well as CommunicatorServer) is now safe for this usage pattern.
+
+### Change: THUNDER_PERFORMANCE code made to work again
+
+The Thunder code using the THUNDER_PERFORMANCE has been made working again (this enables code in Thunder that can be used by a special plugin developed in the past to do JSON-RPC performance measurements)
```
VeithMetro: _"has been made working again"_ sounds a bit off, perhaps something like _"is functional again"_ or just _"is now working properly"_?

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+The flag has a generic name so that future possible performance measurement code can also be put inside this flag.
+
+### Change: Connector timeout can be specified.
+
+It is now possible to set the timeout to bes used in the SmartLinkType connect and disconnect (before it was always Core::infinite so it would wait indefinitely for the event to happen)
```
VeithMetro: _"bes"_ → _"be"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Change: General bug fixes
+
+A number of issues found in Thunder 5.2 have been corrected in Thunder 5.3.
+
+## Breaking Changes
+
+In principle Thunder 5.3 does not contain breaking changes. There are some changes that are not expected to cause any issue but that do deviate from how they behaved before Thunder 5.3, they are listed below.
```
VeithMetro: Maybe this would sound better?
_"Some changes alter previous behavior but are not expected to cause issues; they are listed below"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Feature: close channel on COM-RPC timeout
+
+To prevent issue as a result of "collateral damage" after a COM-RPC time out has happened we now close the channel on which the COM-RPC time happen automatically.
```
VeithMetro: _"time out"_ → _"timeout"_
_"on which the COM-RPC time happen"_ → _"on which the COM-RPC timeout occurred"_

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,444 @@
...
+The Thunder Core class TriState has been removed. It was suboptimal and nowadays alternatives exist like OptionalType<bool>. 
+We do not expect this to be used anywhere nor was it used in Thunder internally.
+
+# Thunder Tools
+
+## Changes and new Features 
+
+### Feature: event indexes improvement
+
+Using indexes for event has improved.
```
VeithMetro: _"Using indexes for events has been improved."_

VeithMetro (commented):
Overall great job! Just some typos and a few potential improvements

ReleaseNotes/ThunderReleaseNotes_R5.3.md
MFransen69: Haha indeed this was Dutch :)

docs/utils/customcodes.md
MFransen69: Ah ships, this is also in the code header file not only in the document, good you noticed :)

docs/plugin/interfaces/tags.md
```diff
@@ -611,6 +619,64 @@ Example list:
...
+This tag can be placed at class or method level, where at class level is by far preferable as it prevents inconsistencies in JSON-RPC function handling.
+Wrapped will for a single output parameter also add the parameter name to the result, making it always a JSON object.
+Note can also be used for array, std::vector, iterator, POD etc. single output parameter.
+Incorrect or inconsistent usage will lead to an error raised by the code generators. Of course the documentation generators do take the wrapped tag into account.
+
+Remark: of course it is preferable to keep the JSON-RPC interface as whole consistent so for that reason be hesitant when using this tag (it was added as there are interface where workarounds are used to achieve the wrapped effect).
+
+#### Example
+
+Without the wrapped tag in case of a single out put parameter the result will be as short as possible to make it as efficient as possible:
```
nxtum: out put to be joined together as output

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
+# Thunder Custom (Error) Codes
+
+## Introduction
+
+Error codes used in Thunder are normally predefined, which suffices up to now. These error codes (due to the "JSON-RPC in terms of COM-RPC" feature) are also translated into JSON-RPC error codes. The custom code feature supports a more flexible error scheme to be used (mainly to have more flexibility in the error codes reported in the JSON-RPC error object).
+There is already a feature for "custom JSON-RPC error messages" which allows for context dependent JSON-RPC messages but also allows to override the error code returned in some form. 
+This however leads to inconsistent error codes between COM-RPC and JSON-RPC for the same handler and is not designed to be used on large scale but for exceptional cases only.
+So with Custom Codes we support a way to have custom error codes that are consistent for both COM-RPC and JSON-RPC (or any other future protocol that will be implemented in terns of COM-RPC), allow for custom (not hardcoded inside Thunder) code to string translation and allow for direct influence on the error code returned by JSON-RPC.
```
nxtum: terms spelt as "terns"

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
+# Thunder Custom (Error) Codes
+
+## Introduction
+
+Error codes used in Thunder are normally predefined, which suffices up to now. These error codes (due to the "JSON-RPC in terms of COM-RPC" feature) are also translated into JSON-RPC error codes. The custom code feature supports a more flexible error scheme to be used (mainly to have more flexibility in the error codes reported in the JSON-RPC error object).
+There is already a feature for "custom JSON-RPC error messages" which allows for context dependent JSON-RPC messages but also allows to override the error code returned in some form. 
+This however leads to inconsistent error codes between COM-RPC and JSON-RPC for the same handler and is not designed to be used on large scale but for exceptional cases only.
```
nxtum: is not designed to be used on _a_ large scale

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+This cannot be done in the ErrorToString as it returns a pointer to the text, therefore an extended version is created that will include the custom code number.
+(changing the ErrorToString was not desirable as that would break backwards compatibility unnecessarily)
+
+## String representation
+
+To be able to provide for a custom error code to string conversion (e.g. used for the JSON-RPC error text in the error object or to return a correct string representation when Core::ErrorToString is called) the following infrastructure will be added to Thunder:
+
+In Thunder sources/core a header file "ICustomErrorCodes.h" provides the interface to be implemented by an external library to support the custom error codes used in the plugin code for that Thunder instance.
+
+For now there is only only function in this interface called "CustomCodeToString":
```
nxtum: only repeated twice

docs/utils/customcodes.md
```diff
@@ -0,0 +1,224 @@
...
+
+#ifdef __cplusplus
+#include <cstdint>
+extern "C" {
+#else
+#include <stdint.h>
+#endif
+
+// called from within Thunder to get the string representation for a custom code
+// note parameter code is the pure (signed) Custom Code passed to Thunder. no additional bits set (so for signed numbers 32th bit used as sign bit).
```
nxtum: 32nd

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+
+##  Process Changes and new Features 
+
+### Feature: Windows build artifacts cleanup
+
+The Windows build (Visual Studio Projects) now fully cleans up the build artifacts when the solution is cleaned guaranteeing a complete fresh build after cleaning.
+
+### Feature: Scripts before Thunder start in CMake
+
+A feature was added to the Thunder CMake files to enable the inclusion of scripts to be executed before Thunder is started by the (Linux) startup script. These scripts should be placed [here](https://github.com/rdkcentral/Thunder/tree/master/Source/Thunder/scripts) and see there as well for some exampes.
```
nxtum: examples spelt as "exames"

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+
+In case of an incorrect configuration, e.g. medium and/or low thresholds higher than the total amount of threads will trigger an ASSERT.
+
+As this is a relatively new but potentially complex feature, the Thunder documentation will be extended with a more detailed description than the above.
+
+### Feature: Custom (Error) Codes.
+
+Before Thunder 5.3 error codes used in Thunder were predefined, which sufficed up to now. These error codes (due to the "JSON-RPC in terms of COM-RPC" feature) are also translated into JSON-RPC error codes and there was a request to support a more flexible error scheme (mainly to have more flexibility in the error codes reported in the JSON-RPC error object).
+So with this new Custom Codes feature a mechanism was introduced to have custom error codes that are consistent for both COM-RPC and JSON-RPC (or any other future protocol that will be implemented in terns of COM-RPC), allow for custom (not hardcoded inside Thunder) code to string translation and allow for direct influence on the error code returned by JSON-RPC.
```
nxtum: terms spelt as "terns"

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+Starting 5.3 this is also done when the unregister is a result of the channel from which the client registered closing without the client calling unregister explicitly before that (note the internal cleanup inside Thunder did already happen, there were no leaks, just the statuslistener was not called when this happened).
+
+### Beta Feature: PSG with interface parsing
+
+NOTE: Tym updates the Thunder documentation add a link to it here as well (see below).
+
+The Plugin Skeleton Generator has been greatly extended to now parse the IDL interface file(s) you want the generated plugin code to implement and therefore generates code already completely providing all methods for the Plugin, only the implementation needs to be added!
+It does generate code for both COM-RPC and JSON-RPC interfaces just by checking if the interface IDL header file indicates also specifies a JSON-RPC interface should be generated.
+It has also been extended with example code for dangling proxies (if applicable fot the interface it implements) and support for subsystem handling code generation.
```
nxtum: for spelt as "fot"

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+### Feature: there is now an example section for interfaces
+
+The Thunder interfaces now contain a special section for example interfaces that are not meant to be used in production, see [here](https://github.com/rdkcentral/ThunderInterfaces/tree/master/example_interfaces)
+
+This also includes new example interfaces for features added in Thunder 5.3.
+
+### Change: interfaces updated for lower case legacy
+
+A number of interfaces being used by contributed plugins still used the JSON meta file solution.
+These have been changed in to full IDL header files so the legacy-lowecase feature could be used to make the interfaces backwards compatible in JSON-RPC casing now the default for this has changed.
```
nxtum: legacy-lowercase

docs/plugin/interfaces/tags.md
```diff
@@ -564,9 +569,12 @@ For more details, click [here](../interfaces/#asynchronous-functions)
 
 #### @encode
 
-This tag encodes data into an alternate format.
+This tag encodes or decodes (if an input parameter) data into/from an alternate format.
 
-* `@encode:base64` encodes arrays as base64 JSON-RPC arrays, on the condition that the array base is type `uint8_t`.
+* `@encode:base64` encodes/decodes arrays (or std::vector or iterator) as base64 JSON-RPC string, on the condition that the array base is type `uint8_t`.
```
sebaszm: encode:hex encode:base64 work for C-style arrays, std::vector and ptr+len buffers if the type of the array/vector is uint8_t or char. They do not work yet for iterators.

docs/plugin/interfaces/tags.md
```diff
@@ -564,9 +569,12 @@ For more details, click [here](../interfaces/#asynchronous-functions)
...
+* `@encode:hex` encodes/decodes arrays (or std::vector or iterator) as a Hex JSON-RPC string, on the condition that the array base is type `uint8_t`.
+
+@encode:autolookup is another form of encode: it indicates this interface is an interface used as a session object for another interface. See for more info here: [here](../interfaces/#object-lookup)
```
sebaszm: ... and encode:lookup

docs/plugin/interfaces/tags.md
```diff
@@ -611,6 +619,64 @@ Example list:
 
 <hr>
 
+#### Wrapped
+
+This tag can be placed at class or method level, where at class level is by far preferable as it prevents inconsistencies in JSON-RPC function handling.
+Wrapped will for a single output parameter also add the parameter name to the result, making it always a JSON object.
+Note can also be used for array, std::vector, iterator, POD etc. single output parameter.
```
sebaszm: ok, when the result is already a POD wrapped does nothing... but perhaps if wrap is beside a method it should still wrap it, but I don't think the class-wide wrap should do anything (why need another indirection level?)

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+NOTE: add link to documentation.
+
+### Feature: new buffer encoding options
+
+NOTE: NOT Document do we still want to remove before the release the encode:mac now we have the natively supported type? (any drawbacks here?)
+(after this decision I'll update the Thunder documentation)
+
+There are new encoding tags supported for @encode (next to the already existing base64):
+
+@encode:hex will encode/decode the buffer as hex value into/from the JSON_RPC string (so works for both input and out parameters). Buffer can be an array, std::vector or even an iterator
```
sebaszm: again iterator is not supported, but buffer+len is

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+It does generate code for both COM-RPC and JSON-RPC interfaces just by checking if the interface IDL header file indicates also specifies a JSON-RPC interface should be generated.
+It has also been extended with example code for dangling proxies (if applicable fot the interface it implements) and support for subsystem handling code generation.
+
+Note due to all the permutations possible with the interfaces it can encounter the PSG remains in beta. If you see issues or have doubts on the code it generates please contact us.
+
+Note: see here for the (updated) Thunder documentation on the Plugin Skeleton Generator here
+
+### Feature: wrapped format
+
+The newly added wrapped tag will for a single output parameter also add the parameter name to the result, making it always a JSON object. It can also be used for arrays, std::vector, iterator and POD's.
```
sebaszm: PODs is not true. need to discuss this

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+If the string is empty this will already generate an error when validating the input in the generated proxy stub code.
+
+NOTE: self still add to Thunder documentation 
+
+### Feature: collated iterators
+
+Iterators in the past always got their values one by one via the COM-RPC interface leading to overhead in an IPC situation as multiple COM-RPC calls are needed while most of the times the iterators were used to in the end get all the values.
+It is now possible to make all the iterators get all the values in one go and after that the COM-RPC calls to get the values will only be local calls.
+To enable this mode for all iterators you can pass the flag --collated-iterators to the proxy stub generator.
+With this mode enabled one can by specifying the tag @ASINTERFACE with an iterator make that one work like it did in the past and get the values one by one.
```
sebaszm: @interface not @asinterface

ReleaseNotes/ThunderReleaseNotes_R5.3.md
```diff
@@ -0,0 +1,408 @@
...
+
+Note in Thunder 6 this mode might become the default, that has not been done yet to give the opportunity to put the @ASINTERFACE at the iterators that potentially deal with huge amounts of data where this might pose a too big of a memory penalty.
+
+NOTE: is @interface with capitals? I guess put it as after the iterator name? also if you put it with the typedef (you would not need to repeat it)
```
sebaszm: no, all tags are lowercase

sebaszm (dismissed):

docs/plugin/interfaces/tags.md
MFransen69: I was just adding the custom part, wait for the next update :)

ReleaseNotes/ThunderReleaseNotes_R5.3.md
MFransen69: Wait for the update, this was work in progress  :)

ReleaseNotes/ThunderReleaseNotes_R5.3.md
MFransen69: Wait for the update, this was work in progress  :)

pwielders (approved):

PR #2026, opened 2025-11-25
[Core] move assert to DestroySocket

MFransen69:


pwielders (approved):

MFransen69:
After reconsideration we considered it putting the horse behind the wagon

PR #2042, opened 2025-12-18
Development/customerrorcodes

sivarajappan-siva:


Source/WPEFramework/Config.h
```diff
@@ -477,6 +483,9 @@ namespace PluginHost {
             Observables Observe;
 #ifdef HIBERNATE_SUPPORT_ENABLED
             HibernateConfig Hibernate;
+#endif
+#ifndef __DISABLE_USE_COMPLEMENTARY_CODE_SET__
```
sramani-metro, line 487: Have you verified by enabling the __DISABLE_USE_COMPLEMENTARY_CODE_SET__ switch?

Source/core/Portability.cpp
```diff
@@ -407,5 +409,111 @@ namespace Core {
...
+
+    hresult CustomCode(const int32_t customCode) {
+
+        static_assert(CUSTOM_ERROR == 0x1000000, "Code below assumes 25th bit used for CUSTOM_ERROR");
+
+        hresult result = Core::ERROR_NONE;
+
+        if (customCode != 0) {
+            int24_t code;
+            if (Core::Frame::check_and_cast<int24_t, int32_t>(customCode, code) == true) {
```
sramani-metro: Since we are making

using uint24_t = uint32_t;
using int24_t = int32_t;

The 2 inputs to check_and_cast becomes <int32_t,int32_t> effectively making the method a no-op? check_and_cast will return true for every possible int32_t value that is passed in. It will not enforce the 24 bit constraint at all. 
The next line does result = static_cast<hresult>(code & 0xFFFFFF); which applies the real constraint by ignoring all the higher level bits than 24. Should we log this unexpected input?

sramani-metro (changes requested):

Source/WPEFramework/Config.h
MFransen69: Oh @sivarajappan-siva I must admit I didn't test that myself (blush) :)
It is more a "safety" mechanism so we can disable less essential stuff.
Might you try and hit issues (already caused by my changes) just let me know and I will help

MFransen69:
Merged via the "big bang" 4_4 PR

PR #2055, opened 2026-02-07
Development/don't touch c or cxx flags

bramoosterhuis:


Source/Thunder/PluginHost.cpp
```diff
@@ -559,139 +559,145 @@ POP_WARNING()
...
             } gc;
 
             gc.FromString(messagingSettings);
 
             WarningReporting::WarningReportingUnit::Instance().Defaults(gc.WarningReporting.Settings.Value());
-            #endif
+#endif
         }
     }
+    } // extern "C"
```
VeithMetro, line 572: Just to make sure - we no longer need main to be inside `extern "C"`?

cmake/common/CompileSettingsDebug.cmake
```diff
@@ -31,7 +31,11 @@ include(CMakePackageConfigHelpers)
 #
 if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
     if(CMAKE_CXX_COMPILER_ID MATCHES "^(Apple)?(Clang)|(GNU)$")
-        target_compile_options(CompileSettingsDebug INTERFACE -O0 -g)
+        if(DEFINED FORTIFY_SOURCES AND FORTIFY_SOURCES OR CMAKE_C_FLAGS MATCHES "_FORTIFY_SOURCE" OR CMAKE_CXX_FLAGS MATCHES "_FORTIFY_SOURCE")
```
VeithMetro: I suppose this has been validated against a buildroot build? Just making sure, because the comment in the removed code in `project.cmake.in` mentioned that before "it was a real problem"

cmake/common/CompileSettingsStrict.cmake
```diff
@@ -0,0 +1,73 @@
...
+            OFF -> apply minimal compatibility tweaks (preserve previous behaviour).
+        - Applies appropriate flags for Clang, GCC and MSVC.
+
+    Guidelines:
+        - Keep only compiler-wide, project-relevant flags here. Avoid target-specific
+            or platform-specific logic that belongs in the corresponding target or
+            platform layer.
+]]
+
+option(ENABLE_STRICT_COMPILER_SETTINGS "Enable strict compiler warnings and treat some warnings as errors" OFF)
```
VeithMetro, line 41: As of right now, in Actions, I was adding the strict compiler settings like this:
```
-DCMAKE_CXX_FLAGS="-Wall -Wextra -Wpedantic -Werror"
-DCMAKE_C_FLAGS="-Wall -Wextra -Wpedantic -Werror"
```
As far as I understand, now we just need to make sure this new option is enabled, correct?

cmake/common/CompileSettingsStrict.cmake
```diff
@@ -0,0 +1,73 @@
...
+
+if(ENABLE_STRICT_COMPILER_SETTINGS)
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        message(STATUS "Enabling strict compiler settings for Clang")
+        set(_STRICT_OPTS -Wall -Wextra -Werror=return-type -Werror=array-bounds -Wnon-virtual-dtor -Wmisleading-indentation)
```
VeithMetro: For Clang, we don't have `-Werror` or `-Wpedantic`, so it is less strict than GCC?

I am wondering, because as of right now, we were building in Actions with Clang using the same strictness flags as with GCC (and usually Clang is the one giving us the most useful compile warnings)

Tests/unit/CMakeLists.txt
```diff
@@ -18,10 +18,10 @@
 find_package(GTest REQUIRED)
 
 # TODO: coverage causes attempt to write on invalid location, -keep-static-functions is not recognized by buildroot compiler
-set(CMAKE_CPP_FLAGS "${CMAKE_CPP_FLAGS} --coverage -fno-inline -fno-inline-small-functions -fno-default-inline -I${PARENT_DIR}/Source")
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pthread")
```
VeithMetro, line 22: Do I get it correctly that we no longer need to specify `-std=c++11` in any of the tests because the standard is now propagated from Thunder's CMake?

VeithMetro (commented):
I am by no means a CMake expect, but I think the changes look good (replacing `CMAKE_*_FLAGS` string manipulation with specific CMake idioms, and extracting these strict compiler settings into a dedicated `CompileSettingsStrict.cmake` file). Nice work 😄

Source/Thunder/PluginHost.cpp
bramoosterhuis: I quote [cppreference](https://en.cppreference.com/w/cpp/language/main_function.html)
> The main function has several restrictions (violation of which renders the program ill-formed):
> 1) It cannot be [named](https://en.cppreference.com/w/cpp/language/definition.html#Naming_a_function)  anywhere in the program
> a) in particular, it cannot be called recursively
> b) its address cannot be taken
> c) it cannot be used in a [typeid](https://en.cppreference.com/w/cpp/language/typeid.html) expression or a > [decltype](https://en.cppreference.com/w/cpp/language/decltype.html) specifier(since C++11)
> 2) It cannot be predefined and cannot be overloaded: effectively, the name main in the global namespace is reserved for functions (although it can be used to name classes, namespaces, enumerations, and any entity in a non-global namespace, except that an entity named main cannot be declared with C [language linkage](https://en.cppreference.com/w/cpp/language/language_linkage.html) in any namespace).

also is found in [c++draft](https://eel.is/c++draft/basic.start.main#3.4) 

> 3.4 an entity named main with C language linkage (in any namespace) is ill-formed[.](https://eel.is/c++draft/basic.start.main#3.sentence-6)

But gcc15 is now warning for it and that is an error for us...

Tests/unit/CMakeLists.txt
bramoosterhuis: No, Thunder's CMake uses set_target_properties with CXX_STANDARD which is not propagated to consumers, so the tests still need to set it explicitly. Fixed by setting CMAKE_CXX_STANDARD and CMAKE_CXX_STANDARD_REQUIRED in the CMake.

cmake/common/CompileSettingsStrict.cmake
bramoosterhuis: Yes, exactly — that's the whole point of the option. Instead of each workflow/CI script manually setting -Wall -Wextra -Wpedantic -Werror in different places (and risking them getting out of sync), it's now defined once in CMake. Enabling -DENABLE_STRICT_COMPILER_SETTINGS=ON is the single point of truth.

cmake/common/CompileSettingsDebug.cmake
bramoosterhuis: Yes, this has been validated against a Buildroot build. Buildroot enables _FORTIFY_SOURCE by default which requires at least -O1 equivalent — -Og satisfies that while still keeping the code debuggable. Fully disabling optimization with -O0 was causing the issue. Debug experience is unaffected.

bramoosterhuis: BTW, enabling this already caught real bugs in URL.cpp — so it's paying off immediately.

Source/websocket/URL.cpp
```diff
@@ -197,11 +199,18 @@ namespace Core
...
+        const uint32_t count = std::min<uint32_t>(remaining, static_cast<uint32_t>(data.length()));
 
-        ::memcpy(&(destination[index]), data.c_str(), (count * sizeof(TCHAR)));
+        if (count > 0) {
+            ::memcpy(&(destination[index]), data.data(), count * sizeof(TCHAR));
+        }
 
-        return (index + count);
+        return index + count;
     }
```
bramoosterhuis, line 216: **`CopyFragment` (root cause of the build failure)**
The function clamped the copy size against the total buffer size instead of the remaining space from the current write position. Near the end of the buffer this caused `memcpy` to write past the end of the `alloca`'d stack buffer, which the compiler correctly flagged.

Source/websocket/URL.cpp
```diff
@@ -358,8 +369,10 @@ namespace Core
                 index = CopyFragment(url, url_len, index + 1, _ref.Value());
             }
 
-            if (index < url_len) {
+            if (index <= url_len) {
                 url[index] = '\0';
+            } else {
+                url[url_len] = '\0';
             }
```
bramoosterhuis, line 378: **`Text()` null terminator and separator writes**
A secondary consequence of the same area — the null terminator was written with `< url_len` instead of `<= url_len`, potentially leaving the string unterminated when the URL fills exactly to capacity. The `@` and `/` separator writes also lacked bounds guards, relying on an implicit assumption about `CopyFragment`'s return value.

Source/websocket/URL.cpp
```diff
@@ -266,7 +275,7 @@ namespace Core
                     _password = Core::TextFragment(urlStr, user + 1, hostname - user - 1).Text();
                 }
  
-                ParseDomain(Core::TextFragment(urlStr, atsign + 1, marker - atsign - 1));
+                ParseDomain(Core::TextFragment(urlStr, hostname + 1, marker - hostname - 1));
```
bramoosterhuis, line 280: **`ParseDomain` called with wrong offset**
`FindNext` was correctly finding the last `@` to handle passwords containing `@`, but then `ParseDomain` was being called with the first `@` offset instead. This meant URLs like `user:p@ss@example.com` would parse the domain as `ss@example.com`.

Source/websocket/URL.cpp
```diff
@@ -83,16 +84,17 @@ namespace Core
...
                 *destination++ = ' ';
                 srcLength--;
+                dstLength--;
             } else {
                 *destination++ = current;
                 srcLength--;
+                dstLength--;
             }
-
-            dstLength--;
```
bramoosterhuis, line 95: **`Decode()` length accounting on malformed input**
`dstLength--` ran unconditionally at the bottom of the loop regardless of whether anything was written. A malformed `%` sequence at end of input would decrement the accounting without producing output, corrupting the reported output length.

Source/websocket/URL.cpp
```diff
@@ -43,7 +43,7 @@ namespace Core
         uint16_t srcLength = sourceLength;
         uint16_t dstLength = destinationLength;
 
-        while ((*source != '\0') && (srcLength != 0) && (dstLength >= 3)) {
```
bramoosterhuis, line 46: **`Encode()` premature truncation**
The `dstLength >= 3` loop guard was protecting the worst-case `%XX` encoding, but also prevented writing single-byte characters when only 1-2 bytes remained in the destination, silently truncating valid output that would have fit.

Source/websocket/URL.cpp
bramoosterhuis, line 1: Buildroot with fortified sources enabled triggered a `-Werror=array-bounds` compiler error in URL::Text() triggered a broader audit of the file.

Source/websocket/URL.cpp
VeithMetro: Agreed, makes sense to move the `dstLength < 3` check into the else branch, and no to apply the "worst-case" space requirement to all characters

Source/websocket/URL.cpp
VeithMetro: Good catch, we shouldn't decrement when nothing was written…

Looking at the code got me thinking though - in the `current == '%'` branch for malformed input (where `source[0]` or `source[1]` is '\0'), neither `srcLength` nor `dstLength` is decremented, and source has already been incremented past the %. So if I understand correctly, this means the loop will re-read the character after % as a new current?

Source/websocket/URL.cpp
VeithMetro: Once again a great catch... it could have easily caused a stack buffer overflow, when index was near the end of the buffer

Source/websocket/URL.cpp
VeithMetro: Yeah, it looks correct as well, as the buffer was allocated as `alloca((url_len + 1) * sizeof(TCHAR))`, so `url_len` is valid

Source/websocket/URL.cpp
sebaszm: Agree, why not fix this already and bail out on malformed %xx? (But this need a better check, not against nul but with isxdigit to ensure crap like %$# is caught.)

Tests/unit/CMakeLists.txt
```diff
@@ -18,10 +18,13 @@
...
+add_compile_options(--coverage -fno-inline -fno-inline-small-functions -fno-default-inline)
+add_link_options(-fprofile-arcs)
```
sebaszm, line 22: I think profile arcs is implied by --coverage, but doesn't hurt I guess.

sebaszm (dismissed):
Looks good. Feel free to further impove Decode().

volkan-aslan:
Automated tests have been executed on this branch, and the results do not indicate any issues.

cmake/common/CompileSettingsStrict.cmake
bramoosterhuis: Good catch! I've added -Wpedantic and -Werror to the Clang flags so both compilers are now held to the same strictness level.

bramoosterhuis:
Disabling Core_ThreadPool.CheckThreadPool_CancelJob_WhileProcessing for now — the refactoring appears to have subtly changed the timing, exposing a race condition. I’ll address it in a follow-up branch to keep this PR focused.​​​​​​​​​​​​​​​​

Source/websocket/URL.cpp
bramoosterhuis: Good catches you two! I Added `srcLength >= 3` guard and `isxdigit` validation, bailing out on malformed input. 

Note that the caller currently cannot detect whether the input was fully consumed or bailed early, but that needs a refactor of the signature and that is beyond the PR's scope. 

something like: 
``` cpp
uint16_t Decode(const TCHAR* source, const uint16_t sourceLength, 
                TCHAR* destination, const uint16_t destinationLength,
                uint16_t* consumedSource = nullptr);
```
The caller can check if `consumedSource == sourceLength` to know full input was processed.

But these catches also highlights that we really need proper unit tests for the websocket code as well.

Tests/unit/CMakeLists.txt
bramoosterhuis: Fair point, it is implied. Keeping it explicit for clarity but happy to remove it if you prefer.

cmake/common/CompileSettingsDebug.cmake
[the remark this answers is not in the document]
bramoosterhuis: Good bot, accepting the suggestion. The behavior is the same but the explicit parentheses make the intent clear

cmake/common/CompileSettingsStrict.cmake
```diff
@@ -0,0 +1,73 @@
...
+        if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 12)
+            list(APPEND _STRICT_OPTS -Wrange-loop-construct)
+        endif()
+    elseif(CMAKE_COMPILER_IS_GNUCXX)
+        message(STATUS "Enabling strict compiler settings for GCC")
+        set(_STRICT_OPTS -Wall -Wextra -Wpedantic -Werror -Wnon-virtual-dtor -Wmisleading-indentation)
+        if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 12)
```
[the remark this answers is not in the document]
bramoosterhuis: Good bot. VERSION_GREATER excludes 12.0.0 itself which is unintentional, VERSION_GREATER_EQUAL is correct.

Source/websocket/URL.cpp
```diff
@@ -329,8 +338,9 @@ namespace Core
                     index = CopyFragment(url, url_len, index + 1, _password.Value());
```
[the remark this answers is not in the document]
bramoosterhuis: Valid observation, but at this point we're already truncating a URL that has hit the 2048 character limit — a : without a password is the least of the problems. The existing guard index < url_len prevents an out-of-bounds write, and CopyFragment handles the rest safely. Closing as won't fix, the truncation behavior at MaximumURLLength is a known limitation.

Source/websocket/URL.cpp
```diff
@@ -342,8 +352,9 @@ namespace Core
                 }
```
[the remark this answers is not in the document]
bramoosterhuis: Same as the previous : separator comment — truncation at 2048 characters is a known limitation and all writes are already bounds-guarded. Closing both as won't fix.

Source/websocket/URL.cpp
```diff
@@ -358,8 +369,10 @@ namespace Core
                 index = CopyFragment(url, url_len, index + 1, _ref.Value());
```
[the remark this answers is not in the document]
bramoosterhuis: Same as the previous : separator comment — truncation at 2048 characters is a known limitation and all writes are already bounds-guarded. Closing both as won't fix.

cmake/common/CompileSettingsStrict.cmake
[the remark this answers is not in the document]
bramoosterhuis: Good bot. VERSION_GREATER excludes 12.0.0 itself which is unintentional, VERSION_GREATER_EQUAL is correct.

VeithMetro (approved):

sebaszm (approved):

PR #2057, opened 2026-02-10
Fix RAM/swap reporting to use sysinfo mem_unit scaling

nanimatta:
Fix RAM/swap reporting to use sysinfo mem_unit scaling


karuna2git (approved):

MFransen69:
I'm not against this, but it might break backwards compatibly (e.g. DeviceInfo plugin exposes this info even to the outside and also the PerformanceMetrics plugin uses it as is). Let me look into this more closely

MFransen69:
Hi @nanimatta what is the reason behind this change? Do you need it urgently? I looked into it and I think it will not cause issues in places where we use the reported sizes, just the number will be different and perhaps more to what one would expect to be reported (and on embedded systems one should not expect an overflow as the reported memory should not be too high).
But as it breaks backwards compatibility I (and not even compile time but in behaviour) would only like to only include it in a Thunder major release so we can warn that this interface (behaviour) has changed.

nanimatta:
@MFransen69 - We identified an issue where the total RAM was being reported in pages instead of bytes. This occurs because the kernel sets the mem_unit value depending on whether the available RAM fits within ULONG_MAX.

On devices with 3GB DDR running a 32‑bit kernel, mem_unit is set to page_size[4096].
On devices with 2GB DDR, mem_unit is set to 1.

Because of this, the correct RAM size in bytes must be calculated using mem_unit as a multiplier. Our patch addresses this discrepancy to ensure the reported RAM is accurate across different memory configurations.
The issue is not urgent since I have already applied the fix as a patch. However, I believe it should also be incorporated into the master branch to keep it aligned and prevent future regressions. Therefore, I have raised a PR with this update.

MFransen69:
Hi @nanimatta 
Can you then tell me where the info returned by the systeminfo is used causing the issue? Is it in a plugin?
The patch is quite "dangerous" and might break code already using the info returned and as mentioned even more dangerous because it is a behavioral change  and not make the build fail. 
Who reviewed/approved that patch? I assume it then added new methods to get the new info instead of silently changing the existing ones? Where can I find the patch?

nanimatta:
@MFransen69 - Here's the patch https://github.com/rdkcentral/meta-rdk-video/pull/2761 , we triggered custom builds with patch and verified on multiple devices. Not seen any issues. 
cc: @karuna2git

MFransen69:
Thanks, @nanimatta 
And can you please let me know where exactly there was an issue before this change?

MFransen69 (approved):

PR #2067, opened 2026-03-02
Add safeguard against self-revocation of jobs

sivarajappan-siva:


MFransen69 (approved):
Approved as it fixes the issue when revoking from the joined thread. We'll fix the other issues in another PR

PR #2078, opened 2026-03-11
RDKDEV-1206 ERROR_ILLEGAL_STATE on quick rectivate

tomasz-karczewski-red:
In case of quick plugin deactivate/activate requests, it is currently possible for Thunder to respond with ERROR_ILLEGAL_STATE.
Seems that this is due to synchronisation implementation in Server::Service::Activate and Server::Service::Deactivate; while the execution is generally guarded with lock, these methods are partially allowed to run in parallel (there are parts of init/deinit code that run outside of critical section). And depending on the timing, the plugin state can still be DEACTIVATION when Activate starts to execute; in such case the request fails with ERROR_ILLEGAL_STATE. 
Exiting of critical section inside of Activate/Deactivate seems to be by design, to allow the plugins Initialize/Deinitialize methds run witout the Service admin lock, and would probably be dangerous to change (might result in deadlocks on initialization of some plugis, depending on plugins implementation, outside of Thunder core). Other solution would be to wait for some limited time to allow the ACTIVATION / DEACTIVATION action to finish.

This PR adds a mechanism to allow for up to 500ms for activation/deactivation to finish, before failing with ERROR_ILLEGAL_STATE.

Tested mostly on LG stb devices; also sanity-checked on linux/x86-64 with a script:

while true; do curl -X PUT http://127.0.0.1:9998/Service/Controller/Deactivate/OCDM; curl http://127.0.0.1:9998/Service/Controller | jq -r '.plugins[] | select(.callsign == "OCDM") | .state'; curl -X PUT http://127.0.0.1:9998/Service/Controller/Activate/OCDM; curl http://127.0.0.1:9998/Service/Controller | jq -r '.plugins[] | select(.callsign == "OCDM") | .state'; done 2>/dev/null

### DIFF
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -332,6 +332,11 @@ namespace PluginHost {
 
         IShell::state currentState(State());
 
+        if (currentState == IShell::DEACTIVATION) {
+            // give the plugin some time to finish with the deactivation to avoid ERROR_ILLEGAL_STATE
+            currentState = waitForInitializationStateChange(500);
+        }
+
         if (currentState == IShell::state::ACTIVATION) {
             Unlock();
             result = Core::ERROR_INPROGRESS;
@@ -404,6 +409,8 @@ namespace PluginHost {
 
                 _administrator.Initialize(callSign, this);
 
+                _initDeinitDone.ResetEvent();
+
                 State(ACTIVATION);
 
                 Unlock();
@@ -472,6 +479,7 @@ namespace PluginHost {
                     #endif
                     Notify(_T("statechange"), string(_T("{\"state\":\"activated\",\"reason\":\"")) + textReason.Data() + _T("\"}"));
                 }
+                _initDeinitDone.SetEvent();
             }
         } else {
             Unlock();
@@ -521,6 +529,11 @@ namespace PluginHost {
 
         IShell::state currentState(State());
 
+        if (currentState == IShell::ACTIVATION && why != IShell::INITIALIZATION_FAILED) {
+            // give the plugin some time to finish current activation (to avoid ERROR_ILLEGAL_STATE)
+            currentState = waitForInitializationStateChange(500);
+        }
+
         if (currentState == IShell::state::DEACTIVATION) {
             result = Core::ERROR_INPROGRESS;
             Unlock();
@@ -552,6 +565,7 @@ namespace PluginHost {
                 ASSERT(_handler != nullptr);
 
                 State(DEACTIVATION);
+                _initDeinitDone.ResetEvent();
 
                 SystemInfo& systeminfo = _administrator.SubSystemInfo();
 
@@ -591,6 +605,8 @@ namespace PluginHost {
                 REPORT_DURATION_WARNING( { _handler->Deinitialize(this); }, WarningReporting::TooLongPluginState, WarningReporting::TooLongPluginState::StateChange::DEACTIVATION, callSign.c_str());
 
                 Lock();
+
+                _initDeinitDone.SetEvent();
 
                 if (currentState != IShell::state::ACTIVATION) {
                     SYSLOG(Logging::Shutdown, (_T("Deactivated plugin [%s]:[%s]"), className.c_str(), callSign.c_str()));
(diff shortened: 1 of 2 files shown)


MFransen69:
Hi Tomasz,
I do understand the issue but can you tell me if this is a realistic scenario or something that came up during testing?
That might help us understanding this situation better.
My first response would be, before perhaps understanding your use case  fully,  is that seen from the system design perspective it is of course not expected that in parallel someone tries to activate a plugin while someone else tries to deactivate it. It is of course never guaranteed who will whin and what do you expect the one who looses to do, try again? Adding the 500ms (which in embedded world terms is quite long, and realize now worst case it blocks a workerpool thread for 500ms doing nothing but waiting) does not change this. So a system architect should be  making sure the plugin activation/deactivation is  done in controlled manor. 
Please give us more information on the use case then we can better understand the request.
(Activate, Deactivate and all plugin state requests, both jsonrpc and comrpc, are synchronous methods so cannot be in parallel if done from one component, if in the example above this sync behavior would be taken into account, wait for the http response, the requests would not go in parallel)
From that perspective the ILLEGAL_STATE is by design, while the plugin is being deactivated/activated it cannot be requested to change state (and not all plugins are written in a correct way so they initialize/deinitialize quickly, some even take more than 500ms :) ), one could listen to the plugin state notifications to know when the plugin has reached a deactivated/activated state. But of course when  state changes are done in parallel from multiple components then it can still happen (as there is not guarantee the state is still the same just after receiving  the state notification) and as mentioned above that would be very unexpected from an architectural system design). 
But please share more info on your use case so we can better analyze the proposed change.

tomasz-karczewski-red:
hello @MFransen69 ,

We caught that problem in a real scenario, where one of the pluins (ocdm) is shut down, and then another component in our stack reactivates it very quickly. OCDM is taking quite a while to deinitialize in our case, since this involves dbus communication via systemd, and so the caller gets ERROR_ILLEGAL_STATE. Seems that there is synchronization in 'lifetime actions' implementation (Activate/Deactivate etc.), but the critical section is exited/reentered for some time, and this opens some window for such situation. The workaround provided is to give the service some time to finish in this case, to alleviate the problems like this (quick deactivate/activate calls) ... The timeout should be changed/configured obviously, but that helped us handle that situation.

MFransen69:
Hi @tomasz-karczewski-red 

Thanks for explaining. 
As mentioned from a system design perspective the logic to handle activation/deactivation plugins should be in one location or at least be done in a controlled way as it makes no sense that one component is trying to deactivate a plugin while another is trying to activate it at the same time (that will always lead to unpredictable behaviour). Seen the description of your situation it is done from two places but in the one where you are restarting may I advice to listen to the plugin state notifications and only try to restart the plugin when you get the  deactivated state notification. This prevents it from being done too soon and might make it more efficient as now you will get notified when the plugin is deactivated and you can take action. It also makes that you do not need to guess the correct timeout value as you will get notified when the correct moment is there.
We cannot take this PR as the logic for this situation should not be in Thunder but be handled outside of it. As mentioned from that perspective the ILLEGAL_STATE reply is correct behaviour, the logic to prevent parallel activates/deactivates should not be in Thunder. It cannot do that correctly as it cannot know what the desired output state is when mutliple requests are received in parallel and it would unnecessarily block  workerpool threads form doing more useful stuff. This is perhaps more clear if you imagine the situation where while deactivating an activate as well as a new deactivate request is received. They would both be awoken at the same time leading to a similar unclear situation. Next to that there are more plugin states that would be needed to taken into account than just activate/deactivate.

So in summery:
The logic you are looking for should not be in Thunder and we think would be easy and efficient to do by registering for plugin state notifications and react to that in the component that restarts the Plugin.
This PR we cannot add to the Thunder code as it is not in agreement with the Thunder design: the waiting logic should not be in Thunder but in the external code, it will unnecessarily block WorkerPool threads (it will be impossible to set a corerct waiting value, short might be too short, too long and it will block workerpool threads "forever", and you can achieve a solution with the "correct" time by listing to the state notifications already), it will not work with multiple parallel requests
 (and the solution for this to keep a queue of request inside Thunder indicates it should not be in Thunder but outside Thunder). It will always lead to raceconditons activating/deactivating plugins which should be prevented, therefore the ILLEGAL_STATE is correct in this case.

tomasz-karczewski-red:
hello @MFransen69 ,

Thanks for the review; the current solution (as outlined in the PR) fixes our Thunder 4.2-based flow, but I understand your reasoning for not wanting Thunder to handle that case directly. We will have an opportunity to reevaluate the issue after RDK8 migration, so will try to keep the suggestion (to use plugin lifecycle notifications) in mind. Closing the PR for now.

PR #2079, opened 2026-03-11
Development/small cmake enhancements and code fixes

bramoosterhuis:
# Fix cross-platform build warnings and static library support

This PR addresses a collection of build failures surfaced by CI across different
platforms (32-bit, macOS/arm64, MinSizeRel) and configurations (static builds,
coverage builds).

### cmake: idempotent patch application (`ApplyPatch.cmake`)
`ApplyPatch` now uses a marker file keyed on an MD5 of the source dir and patch
path. Re-running CMake configure on an already-patched tree no longer fails or
double-applies patches.

### cmake/plugins: static library support for ThunderPlugins
`ThunderPlugins` was hardcoded as `SHARED`; removing that lets `BUILD_SHARED_LIBS`
control the output type. `Module.cpp` is updated to emit
`MODULE_NAME_ARCHIVE_DECLARATION` when built as a static archive, preventing
duplicate symbol collisions (`ModuleBuildRef`, `GetModuleServices`,
`SetModuleServices`) with the Thunder executable's own module definition.
`ThunderProxyStubs` remains `SHARED` and explicitly defines `BUILD_SHARED_LIBS`
to preserve its module declaration.

### core: fix sign-compare in `FrameType::Reader::UnlockBuffer` (`Frame.h`)
The template parameter `TYPENAME` can be a signed type; comparing it directly
against `uint32_t Length()` triggered `-Werror=sign-compare`. Fixed by guarding
with a non-negativity check before the cast.

### test: fix unused-result warning in `test_filesystem`
`system()` return value was discarded without acknowledgement; wrapped with
`PUSH_WARNING(DISABLE_WARNING_UNUSED_RESULT)` / `POP_WARNING()` following the
project convention.


Source/plugins/CMakeLists.txt
```diff
@@ -56,6 +56,10 @@ add_library(${TARGET_PROXYSTUBS} SHARED
         Module.cpp
         )
 
+# ProxyStubs should alwasy be a shared object, so define BUILD_SHARED_LIBS
```
VeithMetro: "alwasy" typo and some trailing white spaces in the line below 😄

Tests/unit/core/test_filesystem.cpp
```diff
@@ -117,7 +117,9 @@ namespace Core {
         EXPECT_TRUE(dirOne.IsDirectory());
         dirOne.Reset();
         EXPECT_TRUE(dirOne.Next());
+PUSH_WARNING(DISABLE_WARNING_UNUSED_RESULT)
         system("rm -rf home");
```
VeithMetro: Good enough, but perhaps we could use something like `std::filesystem::remove_all` and then we don't need to suppress the warning and don't need shell dependency in a unit test? If it is needed somewhere else anyway, let's just leave as is

cmake/common/ApplyPatch.cmake
```diff
@@ -1,58 +1,74 @@
...
+            message(STATUS "Patch applied successfully using patch.")
+        else()
+            message(FATAL_ERROR
+                "Failed to apply patch using patch -p1: ${PATCH_FILE}\n"
+                "Error:\n${PATCH_ERR}"
+            )
+        endif()
+    else()
+        get_filename_component(PATCH_NAME "${PATCH_FILE}" NAME)
+        message(STATUS "Skipping ${PATCH_NAME}, allready applied")
```
VeithMetro: Small typo here 😄

cmake/common/ApplyPatch.cmake
```diff
@@ -1,58 +1,74 @@
...
+    get_filename_component(PATCH_NAME "${PATCH_FILE}" NAME_WE)
+    set(${MARKER} "${CMAKE_CURRENT_BINARY_DIR}/.patch-${PATCH_NAME}-${HASH}" PARENT_SCOPE)
+endfunction()
+
 function(ApplyPatch SOURCE_DIR PATCH_FILE)
+    GetPatchMarker(PATCH_APPLIED ${SOURCE_DIR} ${PATCH_FILE})
+    
     # Find git
-    find_program(GIT_EXECUTABLE git)
+    if (NOT EXISTS ${PATCH_APPLIED})
```
VeithMetro: Not a CMake expert, but doesn't this `${PATCH_APPLIED}` need to be quoted to handle paths with spaces?

VeithMetro (dismissed):
Overall looks good, just some small remarks 😄

Source/plugins/CMakeLists.txt
bramoosterhuis: fixed

cmake/common/ApplyPatch.cmake
bramoosterhuis: fixed

Tests/unit/core/test_filesystem.cpp
bramoosterhuis: Fair point, updated to use `std::filesystem::remove_all("home")` instead.

cmake/common/ApplyPatch.cmake
bramoosterhuis: Sharp, should be "${PATCH_APPLIED}".

VeithMetro (approved):

PR #2084, opened 2026-03-17
[Tests] Add thunder_test_support static library for plugin integration testing

smanes0213:
Adds a static library (libthunder_test_support.a) that embeds the Thunder PluginHost::Server into a linkable archive. Test binaries link against this library to boot a real Thunder runtime without launching the standalone daemon, enabling integration testing of Thunder plugins using GTest/GMock.


mhughesacn:
Hi @smanes0213 : This is draft so not looking closely, but if you append:
```
Copyright 2024 RDK Management
Licensed under the Apache License, Version 2.0
```
to the end of NOTICE in top level, I will clear off blackduck.  Thank you.

smanes0213:
> Hi @smanes0213 : This is draft so not looking closely, but if you append:
> 
> ```
> Copyright 2024 RDK Management
> Licensed under the Apache License, Version 2.0
> ```
> 
> to the end of NOTICE in top level, I will clear off blackduck. Thank you.

Hi @mhughesacn: The changes are not finalised yet, I have removed the license header for now. Will add the header once everything is confirmed.

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,135 @@
...
+        // Boot the embedded Thunder server with the given plugins.
+        // Creates temp directories, generates config, parses it, and calls Server::Open().
+        // systemPath  — directory containing plugin .so files
+        // proxyStubPath — directory containing proxy stub .so files
+        // Returns Core::ERROR_NONE on success.
+        uint32_t Initialize(const std::vector<PluginConfig>& plugins, const string& systemPath = "", const string& proxyStubPath = "");
+
+        // Invoke a JSON-RPC method on a loaded plugin.
+        // Method format: "Callsign.Version.method" (e.g. "Counter.1.increment")
+        uint32_t InvokeJSONRPC(const string& callsign, const string& method,
```
bramoosterhuis: Was there a specific reason for having both `callsign` and `method` as separate parameters here?

As it stands, the callsign is already embedded in the fully-qualified method string (`"Counter.1.increment"`), so the caller ends up saying the same thing twice — and there's nothing stopping them from passing mismatched values, which would produce a confusing failure.

Would it make sense to drop the `callsign` parameter and derive it from `method` in the implementation instead?

see `Core::JSONRPC::Message` , `const string  callsign = Core::JSONRPC::Message::Callsign(method);`

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,135 @@
...
+        Core::ProxyType<PluginHost::IShell> GetShell(const string& callsign);
+
+        // Direct access to the underlying PluginHost::Server.
+        PluginHost::Server& Server();
+
+        // Returns the UNIX domain socket path used by the communicator.
+        string CommunicatorPath() const;
+
+        // Stop the server, release config, and clean up temp directories.
+        void Shutdown();
```
bramoosterhuis: Was there a reason for pairing `Initialize()` with `Shutdown()` rather than following one of Thunder's own naming conventions?

Thunder is consistent elsewhere — `IPlugin` uses `Initialize()`/`Deinitialize()`, and `PluginHost::Server` uses `Open()`/`Close()`. Using `Shutdown()` introduces a third scheme that fits neither, which feels slightly off for a library that lives inside the Thunder repo and targets Thunder plugin developers who are already familiar with those conventions.

Would `Initialize()`/`Deinitialize()` or `Open()`/`Close()` be a better fit here?

Tests/test_support/CMakeLists.txt
bramoosterhuis, line 1: The docs mention that consumers must link with `--whole-archive`, but there's nothing in the CMake target to enforce this. A consumer who forgets it will get silent failures from missing static initializers (`MODULE_NAME_DECLARATION`, `SERVICE_REGISTRATION`) with no clear diagnostic.

Would it make sense to enforce it on the target itself? Since Thunder requires CMake 3.15, `target_link_options` with `INTERFACE` and `$<TARGET_FILE:...>` scopes it to this archive only without affecting other libraries on the consumer's link line:

```cmake
target_link_options(thunder_test_support INTERFACE
    -Wl,--whole-archive $<TARGET_FILE:thunder_test_support> -Wl,--no-whole-archive
)
```

Tests/test_support/Module.cpp
```diff
@@ -0,0 +1,13 @@
...
+// This ensures all server sources and the MODULE_NAME_DECLARATION below
+// use the same symbol, and trace output shows "ThunderTestRuntime".
+
+#ifndef MODULE_NAME
+#define MODULE_NAME ThunderTestRuntime
+#endif
+
+#include <core/core.h>
+
+MODULE_NAME_DECLARATION(BUILD_REFERENCE)
```
bramoosterhuis: Thunder has a macro for static archives — `MODULE_NAME_ARCHIVE_DECLARATION`. Using it in `Module.cpp` instead of `MODULE_NAME_DECLARATION(BUILD_REFERENCE)` avoids conflicting definitions of `ModuleBuildRef()`, `GetModuleServices()` and `SetModuleServices()` with the consumer's own `MODULE_NAME_DECLARATION`. It also removes the need for the `#ifndef` guard here.

```cpp
#define MODULE_NAME ThunderTestRuntime
#include <core/core.h>
MODULE_NAME_ARCHIVE_DECLARATION
```

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,231 @@
...
+            Core::Directory(_tempDir.c_str()).Destroy();
+        }
+    }
+
+    // Build a minimal Thunder JSON config from the plugin list.
+    // Uses port 0 (OS-assigned) and binds to localhost only.
+    string ThunderTestRuntime::BuildConfigJSON(const std::vector<PluginConfig>& plugins,
+                                                const string& systemPath,
+                                                const string& proxyStubPath) const
+    {
```
bramoosterhuis, line 116: `BuildConfigJSON` is prone to JSON injection — if any `PluginConfig` field contains `"` or `\` the generated config will be malformed and parsing will fail with `ERROR_INCOMPLETE_CONFIG`.

Inside `BuildConfigJSON`, converting each field through `Core::JSON::String` before writing it handles escaping correctly without changing the public API:

```cpp
Core::JSON::String escaped;
escaped = p.callsign;
json << "\"callsign\":\"" << escaped.Value() << "\",";
```

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,119 @@
...
+//   TEST_F(MyTest, ComRpc) {
+//       auto* iface = _runtime.GetInterface<IMyInterface>("Callsign");
+//   }
+// ==========================================================================
+
+#ifndef MODULE_NAME
+#define MODULE_NAME Application
+#endif
+
+#include <plugins/plugins.h>
```
bramoosterhuis: Nit: `<plugins/plugins.h>` can be replaced with `<core/core.h>` and `<plugins/IShell.h>` — the header only needs `IShell` and `Core` this avoids pulling in the entire plugins umbrella header into every consumer TU.

bramoosterhuis (commented):
Maybe we want a minimal `main.cpp` example that boots the runtime and exercises a basic JSON-RPC call against the built-in Controller — no external plugin `.so` needed. This would serve as a smoke test to verify the library links and initialises correctly end-to-end.

cmake/common/CompileSettings.cmake
```diff
@@ -62,7 +62,8 @@ endif()
 # target_compile_definitions(CompileSettings INTERFACE "THUNDER_PLATFORM=${THUNDER_PLATFORM}")
 # message(STATUS "Selected platform ${THUNDER_PLATFORM}")
 
-target_compile_options(CompileSettings INTERFACE -Wno-psabi)
+# TEMPORARY: Force C++11 to work around SocketServer.h IteratorType bug under C++17.
+target_compile_options(CompileSettings INTERFACE -std=c++11 -Wno-psabi)
```
bramoosterhuis: If there is a real C++17 incompatibility in `SocketServer.h`, the right fix is `-DENABLE_CXX17=OFF` rather than hardcoding `-std=c++11` in `CompileSettings.cmake`. The hardcoded flag bypasses the existing CMake option and forces C++11 on the entire Thunder build regardless of what the consumer configured.

bramoosterhuis (changes requested):

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,249 @@
...
+        if (!_tempDir.empty()) {
+            Core::Directory(_tempDir.c_str()).Destroy();
+        }
+    }
+
+    // Build a minimal Thunder JSON config from the plugin list.
+    // Uses port 0 (OS-assigned) and binds to localhost only.
+    // Helper: JSON-escape a string value and return it quoted (e.g. "foo\"bar" → "\"foo\\\"bar\"")
+    // Uses Core::JSON::String serialization to handle escaping correctly.
+    static string JsonEscape(const string& value)
```
sebaszm: Suggest to use Core::JSON methods instead of creating JSON by hand. It will be cleaner, more flexible less error-prone and removes the need for manual escaping.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+        for (size_t i = 0; i < plugins.size(); ++i) {
+            const auto& p = plugins[i];
+            if (i > 0) json << ",";
+            json << "{"
+                 << "\"callsign\":" << JsonEscape(p.callsign) << ","
+                 << "\"locator\":" << JsonEscape(p.locator) << ","
+                 << "\"classname\":" << JsonEscape(p.classname) << ","
+                 << "\"startuporder\":" << p.startuporder << ","
+                 << "\"autostart\":" << (p.autostart ? "true" : "false");
```
sebaszm: There is no autostart in 5.x in anymore. Perhaps it's then better to actually reuse the plugin Configuration as defined in plugns/Configuration.h, that would've eliminate such issues.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,249 @@
...
+                                             const string& systemPath,
+                                             const string& proxyStubPath)
+    {
+        if (_server != nullptr) {
+            return Core::ERROR_ALREADY_CONNECTED;
+        }
+
+        // Create unique temp directory for this test run
+        char tempTemplate[] = "/tmp/thunder_test_XXXXXX";
+        char* tempResult = mkdtemp(tempTemplate);
```
sebaszm: For OS-abstraction use Thunder Core functionality (also to remove posix dependency).

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,249 @@
...
+            return Core::ERROR_GENERAL;
+        }
+        _tempDir = string(tempResult) + "/";
+
+        CreateDirectories();
+
+        // Determine system path for plugin .so files
+        string sysPath = systemPath;
+        if (sysPath.empty()) {
+            sysPath = "/usr/lib/wpeframework/plugins/";
```
sebaszm: Such path must not be hardcoded! On 5.x the path is different.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,249 @@
...
+            sysPath += '/';
+        }
+
+        // Determine proxy stub path
+        _proxyStubPath = proxyStubPath;
+        if (_proxyStubPath.empty()) {
+            // Default: look next to system path
+            _proxyStubPath = sysPath + "../proxystubs/";
+        }
+        if (_proxyStubPath.back() != '/') {
```
sebaszm: Use Thunder-provided functionality for normalizing paths!

mhughesacn:
Hi @smanes0213 : I have cleared off the blackduck issue in the tool but please add headers to the source code in this PR.  Thank you!

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+        _config = new PluginHost::Config(configFile, false, error);
+        configFile.Close();
+
+        if (error.IsSet()) {
+            delete _config;
+            _config = nullptr;
+            CleanupDirectories();
+            return Core::ERROR_INCOMPLETE_CONFIG;
+        }
+
```
sebaszm, line 220: I think this is missing messaging startup.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    // Invoke a JSON-RPC method synchronously via the in-process dispatcher.
+    // Bypasses HTTP/WebSocket — calls IDispatcher::Invoke() directly.
+    // Derives the callsign from the method string (text before the first '.').
+    uint32_t ThunderTestRuntime::InvokeJSONRPC(const string& method,
+                                                const string& params, string& response)
+    {
+        if (_server == nullptr) {
+            return Core::ERROR_ILLEGAL_STATE;
```
sebaszm: Multiple return points are not as per coding standards.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+        _server = new PluginHost::Server(*_config, false);
+        _server->Open();
+
+        return Core::ERROR_NONE;
+    }
+
+    // Invoke a JSON-RPC method synchronously via the in-process dispatcher.
+    // Bypasses HTTP/WebSocket — calls IDispatcher::Invoke() directly.
+    // Derives the callsign from the method string (text before the first '.').
+    uint32_t ThunderTestRuntime::InvokeJSONRPC(const string& method,
```
sebaszm: Perhaps let's just call it Invoke() so less changes to the existing unit tests?

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+        // systemPath  — directory containing plugin .so files
+        // proxyStubPath — directory containing proxy stub .so files
+        // Returns Core::ERROR_NONE on success.
+        uint32_t Initialize(const std::vector<PluginConfig>& plugins, const string& systemPath = "", const string& proxyStubPath = "");
+
+        // Invoke a JSON-RPC method on a loaded plugin.
+        // The callsign is derived from the method string (text before the first '.').
+        // Method format: "Callsign.Version.method" (e.g. "MyPlugin.1.doSomething")
+        uint32_t InvokeJSONRPC(const string& method, const string& params, string& response);
+
```
sebaszm, line 74: Missing Exists() method to check if a JSON-RPC endpoint is available.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+            }
+            if (!path.empty()) {
+                Core::Directory(path.c_str()).Destroy();
+            }
+        }
+    }
+
+    // Helper to escape a string for safe inclusion as a JSON string value.
+    // It escapes quotes, backslashes, and control characters (< 0x20).
+    static std::string JsonEscape(const std::string& input)
```
sebaszm: Use Thunder Core JSON for JSON handling, no need for manual escaping.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    string ThunderTestRuntime::BuildConfigJSON(const std::vector<PluginConfig>& plugins,
+                                                const string& systemPath,
+                                                const string& proxyStubPath) const
+    {
+                const string communicatorPath = _tempDir + "communicator|0777";
+
+        std::ostringstream json;
```
sebaszm: Use Thunder Core JSON for JSON manipulation!

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+    class Config;
+}
+
+namespace TestCore {
+
+    class ThunderTestRuntime {
+    public:
+        // Describes a plugin to be loaded by the test runtime.
+        // Maps directly to a Thunder plugin JSON config entry.
+        struct PluginConfig {
```
sebaszm: Can't Thunder Configuration be reused here, insted of crafting another one?

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,322 @@
...
+    {
+        if (_server == nullptr) {
+            return Core::ERROR_ILLEGAL_STATE;
+        }
+
+        size_t dot = method.find('.');
+        if (dot == string::npos) {
+            return Core::ERROR_INVALID_SIGNATURE;
+        }
+        string callsign = method.substr(0, dot);
```
sebaszm: There are Thunder Core JSON Message helper methods for this.

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+//
+//   static void SetUpTestSuite() {
+//       std::vector<PluginConfig> plugins = { ... };
+//       _runtime.Initialize(plugins, pluginPath, proxyStubPath);
+//   }
+//   static void TearDownTestSuite() { _runtime.Deinitialize(); }
+//
+//   TEST_F(MyTest, JsonRpc) {
+//       string resp;
+//       _runtime.InvokeJSONRPC("Callsign.1.method", params, resp);
```
sebaszm: Remove all .1 from JSON-RPC method designators, we don't endorse versioning.

sebaszm (changes requested):
See comments!

Additionally we need to think how to incorporate JSON-RPC event handlig into this solution and remove the need of knowing the callsign of the tested plugin (or at least repeating it with every JSON-RPC call).

Also please ensure this works with Thunder 5.3. (Autostart and 'wpeframework' in paths suggest it was checked with 4.4.) Thanks!

PR #2089, opened 2026-03-26
Development/coverity scan fixes

bramoosterhuis:
## Coverity static analysis fixes — batch 1

This PR addresses a set of real bugs and false positives identified
during a systematic triage of ~599 Coverity findings against Thunder
master (HEAD b5b4746).

---

### Real bug fixes

**BAD_SHIFT — PluginServer.h**
`1 << bitNr` with a signed int literal is undefined behaviour when
`bitNr >= 31`. Changed to `1u << bitNr`.

**Random.cpp — typo and sign extension**
`lsb1 < 16` was a boolean comparison (yields 0 or 1) instead of the
intended left shift `lsb1 << 16`, corrupting uint32/uint64 output.
`uint16_t` values were also shifted in an int32 context before
assignment to uint64_t, causing sign extension when bit 15 is set.

**NodeId.cpp — uninitialised sockaddr_nl**
`sockaddr_nl.nl_pad` was left uninitialised. Fixed with `memset` before
field assignment.

**SocketPort.cpp — source-specific multicast uninitialised field**
A 10-year-old TODO: `ip_mreq_source.imr_interface` was never set on
Linux in the source-specific `Join()`/`Leave()` overloads. Zero-
initialised the struct and set `imr_interface` to `INADDR_ANY`.

**SocketPort.cpp — TOCTOU and domain socket cleanup**
`access()+remove()` replaced with unconditional `unlink()` in
`ConstructSocket()` and `Closed()`. Cleanup extended from `LISTEN`-only
to all binding socket types (`!= STREAM`). The now-redundant second
`unlink()` block removed. 

**SocketPort.cpp — double-open hardening**
Calling `Open()` on an already-open socket triggered an `ASSERT` abort
in debug builds and silently leaked the socket in release. Added an
`else if (m_State != 0)` guard that returns `ERROR_ILLEGAL_STATE`
without destroying the still-registered socket.

**DoorBell.cpp — TOCTOU and dangling pointer**
Same `access()+remove()` TOCTOU race.

**Singleton.h — broken double-checked locking (DCLP)**
The outer `if (g_TypedSingleton == nullptr)` read a plain pointer
without a lock or memory barrier. On weakly-ordered CPUs (ARM) a thread
can observe the non-null pointer before the constructor has finished.
Fixed by declaring `g_TypedSingleton` as `std::atomic<SINGLETON*>` with
acquire/release semantics. `Dispose()` uses `exchange(nullptr, acq_rel)`
to eliminate the double-free window.

**PluginServer.cpp — null dereference in Server::Close()**
`ClassType<Controller>()` result was dereferenced without a null check.
Added ASSERT and null guard.

---

### False positive annotations

- `PW.INCLUDE_RECURSION` — IPlugin.h / IShell.h mutual include is
  intentional and guarded by include guards
- `RESOURCE_LEAK` — Administrator.h ownership transfers via Thunder
  ref-counting
- `PW.CONVERSION_FUNCTION_NOT_USABLE` — Proxy.h intentional implicit
  conversions in the proxy pattern
- `ATOMICITY` — PluginServer.cpp lock-straddling around blocking
  `Deactivate()` is intentional
- `MISSING_LOCK` — ProxyPoolType diagnostic accessors and
  ResourceMonitor::Count() intentionally return momentary snapshots
- `DEADCODE` — PluginServer.cpp Hibernate() and Controller.cpp
  BuildInfo() branches controlled by `#ifdef`
- `INTEGER_OVERFLOW` — WebSerializer.cpp `sizeof(uint8_t[64])` is 64,
  well within uint16_t range

---

### Tests added

- `test_singleton.cpp` — extended with lifetime, DCLP stress, and
  concurrent construction tests
- `test_socketport.cpp` — new file, 29 tests covering state machine,
  TOCTOU fixes, socket options, TTL, Broadcast, multicast, TCP stream,
  UDP loopback, double-open, remote close, and data flow verification
  (65.1% line coverage of SocketPort.cpp)

### DIFF
--- a/Source/Thunder/Controller.cpp
+++ b/Source/Thunder/Controller.cpp
@@ -1585,6 +1585,9 @@ namespace Plugin {
             extensions |= IMetadata::Data::BuildInfo::PROCESS_CONTAINERS;
         #endif
         
+        // coverity[DEADCODE] - extensions is set via #ifdef blocks above. If no optional
+        // features are enabled the guard is dead in that build configuration, but it is
+        // intentionally defensive for all other configurations.
         if (extensions != 0) {
             buildInfo.Extensions = static_cast<IMetadata::Data::BuildInfo::extensiontype>(extensions);
         }
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -1391,12 +1391,17 @@ namespace PluginHost {
     void Server::Close()
     {
         Plugin::Controller* destructor(_controller->ClassType<Plugin::Controller>());
-        destructor->AddRef();
-        _connections.Close(100);
-        destructor->Stopped();
-        _services.Close();
-        _dispatcher.Stop();
-        destructor->Release();
+
+        ASSERT(destructor != nullptr);
+
+        if (destructor != nullptr) {
+            destructor->AddRef();
+            _connections.Close(100);
+            destructor->Stopped();
+            _services.Close();
+            _dispatcher.Stop();
+            destructor->Release();
+        }
         _inputHandler.Deinitialize();
         _connections.Close(Core::infinite);
 
--- a/Source/Thunder/PluginServer.cpp
+++ b/Source/Thunder/PluginServer.cpp
@@ -764,6 +764,9 @@ namespace PluginHost {
                     local->Release();
                     result = Core::ERROR_NONE;
 #endif
+                    // coverity[DEADCODE] - On the non-HIBERNATE_ENABLED path result is always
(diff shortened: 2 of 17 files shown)


sebaszm:
Do we need to keep the false positive comments?

bramoosterhuis:
> Do we need to keep the false positive comments?

I thought leaving a bit of context could help a future developer help to understand that the annotations are placed intentionally, not to just silence the tool. I don't really have a preference, if you want them gone, I'll remove them, but they do not add any machine code either.

MFransen69 (approved):

PR #2095, opened 2026-04-17
[CMake] Add CXX_STANDARD propagation via the CompileSettings interface target

VeithMetro:
This is to make sure the macOS build (and any other build for that matter) properly uses C++17 instead of 11

### DIFF
--- a/cmake/common/CompileSettings.cmake
+++ b/cmake/common/CompileSettings.cmake
@@ -29,12 +29,6 @@ add_library(CompileSettings::CompileSettings ALIAS CompileSettings)
 #
 # Global options
 #
-if(CXX_STD)
-    set_target_properties(CompileSettings PROPERTIES
-        INTERFACE_COMPILE_FEATURES "cxx_std_${CXX_STD}")
-else()
-    message(FATAL_ERROR "CXX_STD is not set")
-endif()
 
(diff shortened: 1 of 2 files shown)


cmake/common/CompileSettings.cmake
```diff
@@ -29,6 +29,13 @@ add_library(CompileSettings::CompileSettings ALIAS CompileSettings)
 #
 # Global options
 #
+if(NOT CXX_STD)
+    set(CXX_STD 11)
+endif()
+
```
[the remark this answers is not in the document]
bramoosterhuis: bad bot... bad bot...

cmake/common/CompileSettings.cmake
```diff
@@ -29,6 +29,13 @@ add_library(CompileSettings::CompileSettings ALIAS CompileSettings)
 #
 # Global options
 #
+if(NOT CXX_STD)
```
bramoosterhuis: Copilot does have a point with the `NOT DEFINED` instead of the `NOT`

VeithMetro: Fair point, but I changed it even a bit more, since `CXX_STD` should always be set, perhaps it is better to explicitly get an error otherwise

PR #2100, opened 2026-04-22
RDKTV-40024: Fix proper cleanup service

SatheeshkumarG15:


sebaszm (approved):
Looks good as a quick and low risk change that solves a problem. We may want to revisit this though in the future to find a solution that fits the architecture better.

PR #2101, opened 2026-04-23
[Messaging] Make output configurable depending on message metadata

VeithMetro:


sebaszm (changes requested):
Rename enums 'plugin' to 'handler' and 'both' to 'all'. Simplify by making the output mode a control property as discussed.

sebaszm (approved):

PR #2103, opened 2026-04-27
[COM][RDKEMW-16318] Fix use-after-free SIGSEGV in UnknownProxy Release

sramani-metro:
Problem
-------
A SIGSEGV crash occurred in PluginInitializerService::PluginStarter:: NotifyInitiator() on a COM-RPC proxy (IActivationCallback) that had already been freed.

The race condition is:
1. DeleteChannel() (socket I/O thread) calls Invalidate() on a proxy that still has refCount==1.
2. Invalidate() calls _channel.Release() which decrements the internal ProxyType refCount.  Because no extra reference was taken first, the ProxyType destructor fires immediately, dropping the proxy's own refCount to 0 — the proxy is deleted while still registered in _danglingProxies.
3. Later, when a worker thread calls Release() → UnregisterUnknownProxy() it finds a recycled address in _danglingProxies, deletes a now-live proxy and corrupts heap state → SIGSEGV.

A second independent bug: UnregisterUnknownProxy() looked up the proxy in _channelProxyMap using proxy.LinkId(), but by the time Release() runs _channel may already be cleared (by a concurrent Invalidate()), so LinkId() returns 0 and the lookup always fails.  The proxy was also never looked up in _danglingProxies from the outer-else path.

Fix
---
IUnknown.h – Invalidate():
  Increment _refCount before releasing _channel so the proxy stays alive
  until UnregisterUnknownProxy() balances the count via its own Release().

IUnknown.h – Release():
  * Capture channelId = _channel->LinkId() before calling _channel.Release() so the ID is available for the UnregisterUnknownProxy() call even when DeleteChannel/Invalidate has already cleared _channel.
  * Remove the now-dead '_refCount == 0' path (unreachable since Invalidate now prevents the spurious zero).
  * Propagate ERROR_CONNECTION_CLOSED when refCount > 1 and channel is gone, so callers know the connection is dead.

Administrator.cpp / Administrator.h – UnregisterUnknownProxy():
  * Accept an explicit 'uintptr_t channelId' parameter rather than calling proxy.LinkId() (which may return 0 after _channel is cleared).
  * When the channel is not found in _channelProxyMap (DeleteChannel already ran), search _danglingProxies by pointer — the correct location for a proxy whose channel has been torn down.
  * Remove the misplaced dangling-search in the found-in-channelMap branch (a proxy cannot be in both lists simultaneously).

### DIFF
--- a/Source/com/IUnknown.h
+++ b/Source/com/IUnknown.h
@@ -160,6 +160,8 @@ namespace ProxyStub {
             ASSERT(_refCount > 0);
             _adminLock.Lock();
             if (_channel.IsValid() == true) {
+                _refCount++;    // Take a reference on behalf of the _danglingProxies entry so the proxy is not
+                                // destroyed while still tracked; balanced by UnregisterUnknownProxy() -> Release().
                 _channel.Release();
                 succeeded = true;
             }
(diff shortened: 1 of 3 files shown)


MFransen69 (approved):

Source/com/IUnknown.h
```diff
@@ -209,42 +211,51 @@ namespace ProxyStub {
...
-            else if( _refCount == 0 ) {
-                _adminLock.Unlock();
-                result = Core::ERROR_DESTRUCTION_SUCCEEDED;
-                ASSERT(_channel.IsValid() == false);
+                if (_channel.IsValid() == false) {
+                    // Channel is gone (Invalidate() was called) but other references still exist.
+                    // Signal the caller so it does not interpret ERROR_NONE as a live connection.
+                    result = Core::ERROR_CONNECTION_CLOSED;
+                }
             }
```
sebaszm, line 219: shouldn't this part still be within the lock?

sramani-metro: Valid Comment. _channel is being released in Invalidate() under a lock. After _adminLock.Unlock(), another thread can call Invalidate() concurrently — which acquires _adminLock, increments _refCount, then calls _channel.Release() — potentially clearing _channel while it is being read here. 
Reading _channel.IsValid() concurrently with Invalidate()'s _channel.Release() (a write) without any synchronization is a race, regardless of whether the logic might seem correct.

But, since this is a backport of the change in master, the fix for this needs to go in master first, tested for any regression. I will create an internal ticket to track this.

PR #2117, opened 2026-05-08
[R4_4] Update Thunder for generators backport

sebaszm:


Source/core/Services.h
```diff
@@ -246,6 +246,11 @@ namespace Core {
         mutable uint32_t _referenceCount;
     };
 
+    template <typename TYPE>
```
MFransen69, line 249: Haha, that is much smarter then adjusting all code :)

Source/plugins/JSONRPC.h
```diff
@@ -58,123 +67,273 @@ namespace PluginHost {
...
+                    _channelId = move._channelId;
+                    _designator = std::move(move._designator);
+                    _index = std::move(move._index);
+                    move._callback = nullptr;
+                    move._channelId = ~0;
+                    return (*this);
+                }
+                Destination& operator=(const Destination& copy)
+                {
+                    if (_callback != nullptr) {
```
MFransen69, line 148: Not a big issue (and it is a backport anyway but 'officialy' here a check should be done against self assignment, that would (possibly) fail because of the Release. And same for the move above of course

MFransen69 (approved):

VeithMetro (approved):

PR #2118, opened 2026-05-12
[Tests] Add thunder_test_support static library for plugin integration testing

smanes0213:
Adds a static library (libthunder_test_support.a) that embeds the Thunder PluginHost::Server into a linkable archive. Test binaries link against this library to boot a real Thunder runtime without launching the standalone daemon, enabling integration testing of Thunder plugins using GTest/GMock.


Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+
+        uint32_t Initialize(const std::vector<PluginConfig>& plugins,
+            const string& systemPath = "",
+            const string& proxyStubPath = "");
+
+        class JSONRPCLink* CreateJSONRPCLink(const string& callsign);
+        uint32_t Invoke(const string& method, const string& params, string& response);
+
+        template <typename INTERFACE>
+        INTERFACE* GetInterface(const string& callsign)
```
sebaszm: This is basically QueryInterfaceByCallsign... Perhaps it would make sense to name it as such?

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+            const string& Callsign() const { return _callsign; }
+
+            uint32_t AddRef() const override
+            {
+                return ++_refCount;
+            }
+            uint32_t Release() const override
+            {
+                const uint32_t result = --_refCount;
+                if (result == 0) {
```
sebaszm: This is racy...

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+            using EventHandler = std::function<void(const string& params)>;
+
+            JSONRPCLink(ThunderTestRuntime& runtime, const string& callsign);
+            ~JSONRPCLink() override;
+
+            JSONRPCLink(const JSONRPCLink&) = delete;
+            JSONRPCLink& operator=(const JSONRPCLink&) = delete;
+
+            uint32_t Invoke(const string& method, const string& params, string& response);
+            uint32_t Subscribe(const string& event, EventHandler handler);
```
sebaszm: I guess it's good for a start but eventually this need to support passing of index.

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,450 @@
...
+        const string& params,
+        string& response)
+    {
+        uint32_t result = Core::ERROR_UNAVAILABLE;
+        PluginHost::IDispatcher* dispatcher = Dispatcher();
+
+        if (dispatcher != nullptr) {
+            const string fullMethod = _callsign + '.' + method;
+
+            if (_runtime.MethodExists(dispatcher, _callsign, method) == false) {
```
sebaszm: Does this need to be checked here, dispatcher Invoke will do it.

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+        Core::ProxyType<PluginHost::IShell> GetShell(const string& callsign);
+        PluginHost::Server& Server();
+        string CommunicatorPath() const;
+        void Deinitialize();
+
+    private:
+        string BuildConfigJSON(const std::vector<PluginConfig>& plugins,
+            const string& systemPath,
+            const string& proxyStubPath) const;
+        bool MethodExists(PluginHost::IDispatcher* dispatcher,
```
sebaszm: This method needs to be public, it will be used from tests. (Please rename to Exists).

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,123 @@
...
+        public:
+            using EventHandler = std::function<void(const string& params)>;
+
+            JSONRPCLink(ThunderTestRuntime& runtime, const string& callsign);
+            ~JSONRPCLink() override;
+
+            JSONRPCLink(const JSONRPCLink&) = delete;
+            JSONRPCLink& operator=(const JSONRPCLink&) = delete;
+
+            uint32_t Invoke(const string& method, const string& params, string& response);
```
sebaszm, line 39: There coulde be an overload for Invoke to accept the JsonObject, to relieve the caller from tostring/fromstring...

Tests/test_support/ThunderTestRuntime.cpp
```diff
@@ -0,0 +1,450 @@
...
+                        dispatcher->Release();
+                    }
+                }
+            }
+        }
+
+        return result;
+    }
+
+    bool ThunderTestRuntime::MethodExists(PluginHost::IDispatcher* dispatcher,
```
sebaszm: I understand there is no simpler implementation because IDispatcher itself does not have a Exists method... Subject to change once it is added. :)

sebaszm (changes requested):
It's important to have the Exists method available. Otherwise, apart from eventual need for supporting index in events, it looks good.

sebaszm (changes requested):
I think there is a misunderstaning, perhaps of my fault. I meant that GetInterface (not CreateJSONRPCLink) resembles QueryInterfaceByCallsign. Would you please sort that out?

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,126 @@
...
+
+namespace TestCore {
+
+    class ThunderTestRuntime {
+    public:
+        using PluginConfig = Plugin::Config;
+
+        class JSONRPCLink : public PluginHost::IDispatcher::ICallback {
+        public:
+            using EventHandler = std::function<void(const string& params)>;
```
sebaszm: Can we already pass the the client id and the index in the EventHandler? Might not be needed today, but if we ever need to do it, we will have to modify this prototype and (if any) all existing test cases.

sebaszm (changes requested):

Tests/test_support/ThunderTestRuntime.h
smanes0213: Yes, it can be added.

smanes0213: I don't see clientId in the framework Event().
`virtual Core::hresult Event(const string& event, const string& designator, const string& index, const string& parameters /* @restrict:(4M-1) */) = 0;`
I have added designator and index in the prototype.
`using EventHandler = std::function<void(const string& designator, const string& index, const string& params)>;`

sebaszm (dismissed):
Good foundation for a test framework!

If IDispatcher was extended with Exists() method then the framework's own Exists could've been implemented more efficiently, but it is implementation detail and perhaps it's questionable to modify IDispatcher just for the sake of testing.

The main purpose of the framework is to do JSON-RPC calls, so perhaps the dispatcher could be cached, rather than look it up and release on each invoke.

Tests/test_support/ThunderTestRuntime.h
sebaszm: I think the designator actually is the client id :)

Tests/test_support/ThunderTestRuntime.h
```diff
@@ -0,0 +1,126 @@
...
+        ~ThunderTestRuntime();
+
+        ThunderTestRuntime(const ThunderTestRuntime&) = delete;
+        ThunderTestRuntime& operator=(const ThunderTestRuntime&) = delete;
+
+        uint32_t Initialize(const std::vector<PluginConfig>& plugins,
+            const string& systemPath = "",
+            const string& proxyStubPath = "");
+
+        class JSONRPCLink* CreateJSONRPCLink(const string& callsign);
```
bramoosterhuis: @sebaszm, Minor nit: would it make sense to return a `Core::ProxyType<JSONRPCLink>` here instead of a raw pointer? That feels a bit more idiomatic for Thunder-style lifetime management and removes the need for manual `Release()` handling by callers.

bramoosterhuis (approved):
Well done, look like a nice base!

sebaszm (approved):
Again, looks good!

PR #2137, opened 2026-05-27
Add new Common Folder to Thunder

nxtum:
https://github.com/WebPlatformForEmbedded/ThunderOnWindows/pull/16


Source/plugins/CMakeLists.txt
sebaszm, line 68: We need to make existing code that's depending on Plugins actually depend now on Common and Olugins. I suppose adding here a INTERFACE link to Common would do the trick (though it'll look "backwards").

sebaszm, line 77: If whe pull Common the WebSocket, CryptAlgo and Messaging dependency would not be needed anymore here

sebaszm (changes requested):
Optimally the solution must not require changes in plugin build makefiles and code.

sebaszm (approved):

PR #2140, opened 2026-05-29
Fix static analysis violations across the Thunder core directory

workkavint-ship-it:
## Summary

We ran a structured static analysis pass on Thunder core components using a prompt-driven workflow and worked through the reported findings one issue at a time. For each finding, the issue was reviewed, the likely root cause was analyzed, and a proposed fix was evaluated before making any code changes. The goal of this PR is to address the reported violations while preserving the existing behavior of the affected components.

The findings covered a range of issues across 13 files, including potential null pointer dereferences, memory and handle management problems, incorrect condition checks, arithmetic underflow, move semantics issues, buffer handling edge cases, and platform-specific initialization concerns.


Source/core/ASN1.h
sebaszm: Here is the same mistake!

sebaszm: buffer[3] and buffer[4] are swapped!

Source/core/ResourceMonitor.h
```diff
@@ -410,9 +414,11 @@ POP_WARNING()
                 // Resize the array to fit..
                 _descriptorArray = static_cast<::pollfd*>(::malloc(sizeof(::pollfd) * _descriptorArrayLength));
 
-                _descriptorArray[0].fd = _signalDescriptor;
-                _descriptorArray[0].events = POLLIN;
-                _descriptorArray[0].revents = 0;
+                if (_descriptorArray != nullptr) {
```
sebaszm: Would it silence the tool if we had ASSERT here instead the if-guard? It will crash anyway later, so I'd rather crash here.

Source/core/SerialPort.cpp
```diff
@@ -934,8 +934,10 @@ void SerialPort::Read(const uint16_t readBytes)
 
         uint8_t* allocatedMemory = static_cast<uint8_t*>(::calloc(_sendBufferSize + _receiveBufferSize, 1));
 
-        _sendBuffer = allocatedMemory;
-        _receiveBuffer = &(allocatedMemory[_sendBufferSize]);
+        if (allocatedMemory != nullptr) {
```
sebaszm: Same comment as above (would ASSERT suffice).

sebaszm (changes requested):
see comments

sebaszm (dismissed):

sebaszm (approved):

PR #2141, opened 2026-05-29
Add syslogs to plugin initialisation area

nxtum:


Source/Thunder/PluginServer.h
```diff
@@ -1512,13 +1515,24 @@ namespace PluginHost {
                                 result = moduleServiceMetadata();
                                 if (result != nullptr) {
                                     library = std::move(newLib);
+                                } else {
+                                    lastError = _T("GetModuleServices returned no service metadata");
                                 }
+                            } else {
+                                lastError = newLib.Error().empty() == false ? newLib.Error() : _T("GetModuleServices symbol missing");
                             }
+                        } else {
```
sebaszm, line 1524: Can we restructure this, so we know if any other error than file not found is stll logged for any locator? (The only error that is perhaps expected is file not found.)

Source/plugins/Shell.cpp
```diff
@@ -63,10 +66,22 @@ namespace PluginHost
...
                 }
                 all_paths->Release();
+
+                if (result == nullptr) {
+                    SYSLOG(Logging::Error, (_T("Root object [%s] for plugin [%s] could not be instantiated in process from locator [%s]. Last candidate [%s], error [%s]"), className.c_str(), Callsign().c_str(), locator.c_str(), lastPath.c_str(), lastError.c_str()));
+                }
+            }
+
+            if ((result == nullptr) && (locator.empty() == true)) {
+                SYSLOG(Logging::Error, (_T("Root object [%s] for plugin [%s] could not be instantiated in process from locator [%s]"), className.c_str(), Callsign().c_str(), locator.c_str()));
```
sebaszm: locator is empty so what to print?

Source/plugins/Shell.cpp
```diff
@@ -63,10 +66,22 @@ namespace PluginHost
                                 className.c_str(),
                                 version,
                                 interface);
-			}
+                        } else if (resource.IsLoaded() == false) {
+                            lastError = resource.Error();
+                        } else {
+                            lastError = _T("No service metadata found in library");
```
sebaszm, line 75: Can we restructure this, so we know if any other error than file not found is stll logged for any locator? (The only error that is perhaps expected is file not found.)

sebaszm (changes requested):

sebaszm (approved):

PR #2143, opened 2026-06-02
[R4_4] Allow Persist only when debug / persist flag is enabled

nxtum:


Source/WPEFramework/Controller.cpp
```diff
@@ -230,6 +232,7 @@ namespace Plugin {
     {
         ASSERT(_pluginServer != nullptr);
 
+#if defined(__DEBUG__) || defined(__ENABLE_PERSIST__)
```
sebaszm: we need a syslog also if the persist is enabled and will be conducted

Source/WPEFramework/Controller.cpp
```diff
@@ -627,11 +635,16 @@ namespace Plugin {
                     result->Message = _T("Discovery cycle initiated");
                 }
             } else if (index.Current() == _T("Persist")) {
-                
-                _pluginServer->Persist();
 
+                Persist();
```
sebaszm: I think there's no reason to call the stubbed Persist() if it's disabled?

Source/WPEFramework/PluginServer.h
```diff
@@ -4741,12 +4742,16 @@ POP_WARNING()
         void Open();
         void Close();
 
+#if defined(__DEBUG__) || defined(__ENABLE_PERSIST__)
```
sebaszm: Perhaps it's better to just make this Persist implementation conitional on flags rather in each of the callers? and elsewhere rely on return code if needed? then there will be less clutter of ifdefs.

Source/WPEFramework/Controller.cpp
```diff
@@ -319,6 +325,8 @@ namespace Plugin {
             // Normalise return code
             if (result != Core::ERROR_NONE) {
                 result = Core::ERROR_GENERAL;
+            } else {
```
sebaszm, line 320: Please make all syslogs a CC_SYSLOG instead, so it better stands out, except for this one in Configuration()

Source/WPEFramework/Controller.cpp
```diff
@@ -25,7 +25,9 @@
 #include "JsonData_Discovery.h"
 
 #include "JDiscovery.h"
+PUSH_WARNING(DISABLE_WARNING_DEPRECATED_USE)
```
sebaszm: this pragma shall not be needed here as already emitted by generator

sebaszm (changes requested):

Source/WPEFramework/Controller.cpp
nxtum: _pluginServer->Persist() has a syslog

Source/WPEFramework/Controller.cpp
nxtum: Not sure, if you move persist to the enabled part, and someone tries to call it when persist is disabled, they won't get a syslog, but rather just a message on the web?

nxtum: I could then add another syslog again, but thought it might make sense to just call the Persist function since it would already do it

sebaszm: Fair enough but then let's do everything in pluginServer::Persist(), so there is one point of condion logic

sebaszm (approved):
looking good!

PR #2149, opened 2026-06-09
Development/incremental test coverage expansion [Test Branch]

smanes0213:


Tests/unit/core/CMakeLists.txt
```diff
@@ -97,12 +97,22 @@ add_executable(${TEST_RUNNER_NAME}
...
+  test_url.cpp
+  test_websocket_protocol.cpp
+  test_tls.cpp
+  test_assertionunit.cpp
 )
 
 target_include_directories(${TEST_RUNNER_NAME} PRIVATE ${CMAKE_SOURCE_DIR}/Source/addons)
 target_include_directories(${TEST_RUNNER_NAME} PRIVATE ${CMAKE_SOURCE_DIR}/Source/Thunder)
 
+find_package(OpenSSL QUIET)
```
bramoosterhuis: Suggest making CMake the single source of truth for whether TLS tests
build at all: gate `test_tls.cpp` under `SECURE_SOCKET` in
target_sources()/target_link_libraries() (see below), and drop the
`#if defined(SECURESOCKETS_ENABLED)` guard inside the file. Two independent
mechanisms encoding the same condition (a CMake option and a preprocessor
macro) can drift out of sync; letting CMake alone decide also means a
misconfiguration fails loudly at compile time instead of silently
compiling an empty test file.

``` cmake
if(SECURE_SOCKET)
    target_sources(${TEST_RUNNER_NAME} PRIVATE test_tls.cpp)
    target_link_libraries(${TEST_RUNNER_NAME}
        PRIVATE
            OpenSSL::SSL
            OpenSSL::Crypto)
endif()
```

Tests/unit/core/test_assertionunit.cpp
```diff
@@ -0,0 +1,144 @@
...
+        proxy.AssertionEvent(metadata, message, ::Thunder::Core::Messaging::OutputMode::HANDLER);
+
+        EXPECT_EQ(handler.EventCount(), 1);
+
+        // Route another
+        proxy.AssertionEvent(metadata, message, ::Thunder::Core::Messaging::OutputMode::DIRECT);
+
+        EXPECT_EQ(handler.EventCount(), 2);
+
+        proxy.Handle(nullptr);
```
bramoosterhuis: Just a suggestion, if someone ever swaps EXPECT_EQ for the stricter ASSERT_EQ here, a failing
assertion returns early and skips proxy.Handle(nullptr), leaving a dangling
pointer registered in the singleton. Suggest an RAII guard so unregistration
doesn't depend on reaching the last line of the test function:

``` c++
struct HandlerGuard {
    explicit HandlerGuard(Assertion::IAssertionUnit* h) { proxy.Handle(h); }
    ~HandlerGuard() { proxy.Handle(nullptr); }
};
```

Tests/unit/core/test_comrpc.cpp
```diff
@@ -850,6 +850,113 @@ namespace COMRPC {
...
+    TEST(COMRPC_Gap, INotification_Construction)
+    {
+        ConnectionNotificationTracker tracker;
+
+        EXPECT_EQ(tracker.ActivatedCount(), 0);
+        EXPECT_EQ(tracker.DeactivatedCount(), 0);
+    }
+
+    // Verify Communicator server can register and unregister notifications
+    TEST(COMRPC_Gap, Communicator_NotificationRegistration)
```
bramoosterhuis, line 918: This test contains no assertions at all: `Register()`/`Unregister()` both return
void, and nothing is checked afterwards, so it passes unconditionally unless
it crashes. In release builds even the internal ASSERTs in Register()/
Unregister() are compiled out.

On top of that, `GTEST_SKIP()` on a failed `Open()` means the test can silently
disable itself: if opening the socket structurally fails in CI (sandbox, no
unix sockets in /tmp, leftover socket file), this reports "skipped" forever
and nobody notices the coverage is gone. For a test added specifically to
close a coverage gap, failing is more useful than skipping.

What is observable here is the refcount contract: `Register()` does
`sink->AddRef()`, `Unregister()` does `sink->Release()`. Suggest exposing the
tracker's refcount and asserting 1 -> 2 -> 1 around the calls, so the test
verifies the registration contract rather than just absence of a crash

Tests/unit/core/test_controller.cpp
```diff
@@ -0,0 +1,289 @@
...
+namespace Tests {
+
+    // =========================================================================
+    // TEST FILE: test_controller.cpp
+    //
+    // Purpose:
+    //   Tests Controller plugin operations (Gap 4) and startup/shutdown
+    //   sequence (Gap 5) using the ThunderTestRuntime.
+    //
+    // Coverage:
```
bramoosterhuis, line 37: Went through each line of this list against the tests below:

- "Controller.status — full response validation": StatusQuery_ReturnsValidJSON
  checks non-empty and that it parses as JSON. No field is inspected.
  StatusQuery_JsonObject checks only the return code, never the response object.
- "Controller.subsystems — subsystem flags": checks non-empty only, doesn't
  even parse the JSON. No flag is read.
- "Controller.activate/deactivate — lifecycle transitions": both tests use
  NonExistentPlugin_12345 and assert an error. No plugin is activated or
  deactivated, so no transition is exercised.
- "Controller.configuration — config read": grep finds no test invoking
  Controller.configuration anywhere in this file.
- "Shutdown deinitialize — clean teardown": DeinitWithoutInit_NoCrash has no
  assertions at all. Nothing verifies teardown is clean.
- "Multiple init/deinit cycles — no resource leaks": three cycles, asserts
  Initialize() succeeds. Nothing measures leaks.

The test names themselves are accurate (NoCrash, ReturnsValidJSON), it's only
this summary block that overstates. Suggest trimming it to match, so it stays
usable as a map of what's actually covered rather than what was intended.

Tests/unit/core/test_nodeid.cpp
```diff
@@ -48,6 +46,303 @@ namespace Core {
...
+
+    TEST(Core_NodeId, IPv4_HostAddress)
+    {
+        ::Thunder::Core::NodeId node("10.0.0.1", 443, ::Thunder::Core::NodeId::TYPE_IPV4);
+
+        EXPECT_FALSE(node.HostAddress().empty());
+        EXPECT_FALSE(node.QualifiedName().empty());
+    }
+
+    TEST(Core_NodeId, IPv4_DefaultMask)
```
bramoosterhuis: Two things here.

The little-endian expectations match current behaviour, but only because all
three addresses end in .1. On little-endian, bit31/bit30 are read from the
last octet (0x01), so DefaultMask() returns 24 for any address in this test.
The three EXPECT_EQs can't distinguish class A from B or C on the platform CI
actually runs on, and the test would pass unchanged with completely different
addresses.

The big-endian branch doesn't match the code at all:

  address        expected here   DefaultMask() returns
  10.0.0.1             8                  24
  172.16.0.1          16                  16
  192.168.1.1         24                   8

Class A and C are swapped in the production code (bit31==0 returns 24 where
classful says 8), and DefaultMask() is missing the ntohl() that IsMulticast()
and IsLocalInterface() do apply a few lines above it in NodeId.cpp. So this
#else branch encodes correct classful behaviour that the code doesn't
implement, and would fail on a big-endian target.

I'd prefer the test to assert what the current code is clearly trying to do
(RFC 791 classful masks), rather than pin the broken output: let it fail,
file an issue, and mark it DISABLED_ with a comment pointing at that issue.
That way the test documents the defect instead of hiding it, and turns green
again once the bug is fixed. Asserting current behaviour keeps a defect that
has been there for years sitting in plain sight while CI stays green.

Worth raising separately: classful addressing has been obsolete since
RFC 1519 (CIDR), so a caller that needs a mask should really be given one
explicitly rather than having it guessed from the first octet. If DefaultMask()
is still used anywhere that matters, deriving it classfully is a fallback at
best. Fixing the A/C swap and the missing ntohl() is the immediate concern,
but it may be worth checking who still depends on this at all.

Tests/unit/core/test_nodeid.cpp
```diff
@@ -48,6 +46,303 @@ namespace Core {
...
+        ::Thunder::Core::NodeId original("192.168.1.1", 80, ::Thunder::Core::NodeId::TYPE_IPV4);
+        ::Thunder::Core::NodeId copy(original);
+
+        EXPECT_TRUE(copy.IsValid());
+        EXPECT_EQ(copy.Type(), original.Type());
+        EXPECT_EQ(copy.PortNumber(), original.PortNumber());
+        EXPECT_EQ(copy, original);
+    }
+
+    TEST(Core_NodeId, MoveConstructor)
```
bramoosterhuis, line 224: This test can't tell a move from a copy. m_structInfo is a POD union, so
std::move on it is a plain copy, and the test only inspects `moved`, never
`original`. It would pass unchanged if the move constructor were removed and
the copy constructor used instead, so no move semantics are pinned here.

What is worth pinning is the asymmetry in NodeId::NodeId(NodeId&&)
(NodeId.cpp:451): m_hostName is genuinely moved out while m_structInfo is
left intact. A moved-from NodeId therefore still reports IsValid() == true
with the same address but an empty hostname. Asserting the state of
`original` after the move would document whether that is intended.

Tests/unit/core/test_nodeid.cpp
```diff
@@ -48,6 +46,303 @@ namespace Core {
...
+    {
+        ::Thunder::Core::NodeId node("192.168.1.1", 80, ::Thunder::Core::NodeId::TYPE_IPV4);
+        ::Thunder::Core::NodeId any = node.AnyInterface();
+
+        EXPECT_TRUE(any.IsValid());
+        EXPECT_TRUE(any.IsAnyInterface());
+        EXPECT_EQ(any.Type(), ::Thunder::Core::NodeId::TYPE_IPV4);
+    }
+
+    TEST(Core_NodeId, Origin_ReturnsValid)
```
bramoosterhuis, line 274: This only asserts `IsValid()`, which `Origin()` would satisfy without doing its
actual job. Zeroing the address is the whole point of the call, so
`EXPECT_TRUE(origin.IsAnyInterface())` or a check on `HostAddress()` would make
this verify something.


More relevant: the IPv6 branch of Origin() looks broken and this test doesn't
reach it. NodeId.cpp:881-883 sets

    response.m_structInfo.IPV6Socket.sin6_family = AF_INET;

on an IPv6 node, and the line below writes through IPV4Socket.sin_addr rather
than IPV6Socket.sin6_addr, which lands at a different offset in the union and
only partially clears the IPv6 address. An IPv6 case here would catch both.

Tests/unit/core/test_nodeid.cpp
```diff
@@ -48,6 +46,303 @@ namespace Core {
...
+        EXPECT_EQ(classC.DefaultMask(), expectedClassC);
+    }
+
+    // =========================================================================
+    // IPv6 — Construction and Edge Cases (Gap 12)
+    // =========================================================================
+
+    TEST(Core_NodeId, IPv6_LoopbackAddress)
+    {
+        if (!::Thunder::Core::NodeId::IsIPV6Enabled()) {
```
bramoosterhuis: IsIPV6Enabled() isn't a host capability check. It returns a static flag that
initialises to true (NodeId.cpp:162) and is only cleared by
ClearIPV6Enabled(), whose sole caller is PluginHost.cpp:750-753 in this project, where the
daemon acts on the IPv6 setting from its config ("Forcing the network to IPv4
only"). It reflects an operator's configuration choice, not whether the host
has an IPv6 stack.

The unit test binary never loads that config, so the flag stays true, this
guard never fires and the IPv6 tests always run. That's fine for coverage,
but the guard reads as "skip when the host has no IPv6" and gives false
reassurance: on a runner without an IPv6 stack the tests run anyway, and a
failure there gives no signal about whether the code or the environment is at
fault. Either drop it as dead code, or use a real capability probe if
environment awareness was the intent.

Tests/unit/core/test_workerpool.cpp
```diff
@@ -1379,6 +1380,11 @@ namespace Core {
         ~Trigger() {
             Stop();
             _decoupledJob.Revoke();
+            // Revoke() skips the wait when the job is actively dispatching (proxy is invalid).
+            // Spin until dispatch completes so Job is not destroyed while still running.
+            while (_decoupledJob.IsIdle() == false) {
```
bramoosterhuis: The race this fixes is real: Revoke() (WorkerPool.h:86-94) only acts when the
proxy is valid, so it returns immediately without waiting when the job is
mid-dispatch, and Job could be destroyed while still running.

The loop itself is unbounded though, and it sits in a destructor. If IsIdle()
never turns true (blocked job, deadlock, or a regression in the pool) the test
run hangs instead of failing, with no result and no stack trace. Every other
wait loop in this PR is bounded (waited < maxWaitTimeMs). Suggest the same
here, with an ASSERT or a logged warning on timeout so a hang surfaces as a
failure.

Tests/unit/core/test_socketport.cpp
bramoosterhuis, line 1: Two things in the same diff.

The scoping fix is right and for the right reason: the socket must be
destroyed before Singleton::Dispose() so it is deregistered from the
ResourceMonitor, which runs its own thread and is not torn down by the
socket's Close().

Separately, the timeouts go 8s -> 16s and maxRetries 4 -> 10 (and 1 -> 10 in
is_suspended_after_remote_close). Doubling timeouts and a tenfold retry bump
usually means a flake was papered over rather than diagnosed. If slow CI
runners are the cause that is fine, but it belongs in the commit message,
otherwise the next person hitting a flake here has no idea whether these
numbers are meaningful or just what made it pass.

Also, both handshakes in remote_unexpected_close_triggers_statechange use the
same initHandshakeValue (0): the parent signals 0, then waits on 0, while the
child waits on 0, closes, then signals 0. Can the parent's own signal satisfy
its own wait here, or does IPTestAdministrator scope signals per direction?
If it doesn't, distinct handshake values would make the intent clearer.

Tests/unit/core/test_message_dispatcher.cpp
bramoosterhuis, line 1: Solid file overall: real assertions, data is read back and compared per byte,
FIFO ordering and variable lengths are both verified.

Two things:

The section is titled "Cyclic Buffer Stress", but nothing here makes the
buffer cycle. DATA_SIZE is 9 KB and HighVolume_RapidPushPop pushes 4 bytes
and pops them immediately, 100 times, so occupancy never exceeds a few bytes.
Wrap-around and buffer-full are the behaviours that distinguish a cyclic
buffer from a plain queue and neither is reached. Pushing past 9 KB without
popping, or running enough iterations to force a wrap, would cover what the
title claims.

PopFromEmptyBuffer: "ERROR_READ_ERROR or similar" is in the source rather than
a guess. MessageDispatcher.h:256 initialises result to Core::ERROR_READ_ERROR
and leaves it there when Read() returns nothing, so EXPECT_EQ on that code
pins the contract. EXPECT_NE(result, ERROR_NONE) would accept any future
error code without noticing the change.

Tests/unit/core/test_jsonrpc_http.cpp
bramoosterhuis, line 1: UnknownMethodReturnsJSONRPCError goes through _handler.Invoke(), so it covers
Core::JSONRPC::Handler's error path. That one is real.

EmptyBodyReturnsError and NonPostVerbReturnsMethodNotAllowed assert the 400/405
branching that this file's own JSONRPCHTTPServer implements at lines 201-206.
The Web::Request/Web::Response round-trip over the socket is genuinely
exercised, but the named behaviour belongs to the mock, not to Thunder, so
those two can't fail on a Thunder regression. Worth renaming or re-scoping so
they don't read as coverage of production request validation.

The response handler now pushes to _responseQueue unconditionally, including
an empty string when there is no body. That's needed for the error responses
here, but it changes shared test infrastructure: existing tests that would
have timed out now receive "" instead. Worth calling out explicitly.

SleepMs(maxInitTime) is doing the real synchronisation work here. The child
opens the server and then waits, so the parent's 4s sleep is the only thing
ensuring the listener is up before the client connects, and the test fails on
a runner slower than that. Having the child signal after Open() and the parent
wait on it would be deterministic and remove 12s of wall clock across these
three tests.

Ports 12355-12357 are hardcoded and bound to 0.0.0.0. Parallel CI jobs on the
same host will collide, and TIME_WAIT from a previous run can block a rebind.
127.0.0.1 would be enough and avoids exposing the listener off-box.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_TRUE(url.IsValid());
+        EXPECT_TRUE(url.UserName().IsSet());
+        EXPECT_STREQ(url.UserName().Value().c_str(), "user");
+        EXPECT_TRUE(url.Password().IsSet());
+        EXPECT_STREQ(url.Password().Value().c_str(), "pass");
+        EXPECT_TRUE(url.Host().IsSet());
+        EXPECT_STREQ(url.Host().Value().c_str(), "host.com");
+    }
+
+    TEST(URL_Parse, UserInfoWithoutPassword)
```
bramoosterhuis: The NOTE isn't accurate. Walking URL.cpp:268-283 with this input, the '@' is
found and ParseDomain() only receives the part after it, so _host is correctly
"ftp.example.com". The username is dropped because line 275 requires a ':'
before the '@' (if (user < hostname)), not because host parsing is polluted.
Asserting Host()'s value here would have shown that.

RFC 3986 makes the password optional in userinfo, so anonymous@host should
yield UserName "anonymous". As written this pins the deviation as intended
behaviour rather than flagging it.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+    // =========================================================================
+    // URL Encoding / Decoding
+    // =========================================================================
+
+    TEST(URL_Encode, SpaceEncodedAsPlus)
+    {
+        const TCHAR source[] = "hello world";
+        TCHAR dest[64] = {};
+
+        uint16_t len = ::Thunder::Core::URL::Encode(source, static_cast<uint16_t>(strlen(source)), dest, sizeof(dest));
```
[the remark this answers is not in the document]
bramoosterhuis: Building on the strlen/_tcslen point: the suggestion is correct as far as it
goes, but it leaves two related problems on the same lines.

sizeof(dest) is passed as destinationLength. URL.cpp:44-60 decrements
dstLength once per TCHAR written, so that parameter counts elements, not
bytes. With char TCHAR the two coincide, which is why this passes today. With
wchar_t, sizeof gives 256 for a 64-element buffer and Encode() writes four
times past the end of a stack array.

Also, const TCHAR source[] = "hello world" won't compile under _UNICODE at
all, since TCHAR is wchar_t there and the literal is char[]. So the suggested
fix corrects the length calculation for a build that already fails on the
line above it.

Full form would be:

    const TCHAR source[] = _T("hello world");
    TCHAR dest[64] = {};

    uint16_t len = ::Thunder::Core::URL::Encode(
        source,
        static_cast<uint16_t>(_tcslen(source)),
        dest,
        sizeof(dest) / sizeof(dest[0]));

Applies to all six call sites (544, 556, 567, 581, 592, 604) and to the four
Base64Encode calls further down, which use the same TCHAR destination buffer.

That said: does anyone actually build Thunder with _UNICODE, in CI or
downstream? If not, all of this is a consistency argument rather than a live
bug, and it's worth deciding whether TCHAR is still carrying its weight here.
If someone does, this file doesn't compile for them today.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        string encoded(dest, len);
+        EXPECT_NE(encoded.find('+'), string::npos);
+    }
+
+    TEST(URL_Encode, AlphanumericPassThrough)
+    {
+        const TCHAR source[] = "abc123XYZ";
+        TCHAR dest[64] = {};
+
+        uint16_t len = ::Thunder::Core::URL::Encode(source, static_cast<uint16_t>(strlen(source)), dest, sizeof(dest));
```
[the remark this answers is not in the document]
bramoosterhuis: Same as line 544: Copilot's suggestion is close but incomplete, see my reply
there for the full fix (sizeof element count and _T() on the literal).

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+
+        EXPECT_STREQ(encoded.c_str(), "abc123XYZ");
+    }
+
+    TEST(URL_Encode, SpecialCharsEncoded)
+    {
+        const TCHAR source[] = "a=b&c";
+        TCHAR dest[64] = {};
+
+        uint16_t len = ::Thunder::Core::URL::Encode(source, static_cast<uint16_t>(strlen(source)), dest, sizeof(dest));
```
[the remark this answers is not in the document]
bramoosterhuis: Same as line 544: Copilot's suggestion is close but incomplete, see my reply
there for the full fix (sizeof element count and _T() on the literal).

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_EQ(encoded.find('='), string::npos);
+        EXPECT_EQ(encoded.find('&'), string::npos);
+    }
+
+    TEST(URL_Decode, DecodePercentEncoded)
+    {
+        const TCHAR source[] = "hello%20world";
+        TCHAR dest[64] = {};
+
+        uint16_t len = ::Thunder::Core::URL::Decode(source, static_cast<uint16_t>(strlen(source)), dest, sizeof(dest));
```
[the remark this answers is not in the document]
bramoosterhuis: Same as line 544: Copilot's suggestion is close but incomplete, see my reply
there for the full fix (sizeof element count and _T() on the literal).

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+
+        EXPECT_STREQ(decoded.c_str(), "hello world");
+    }
+
+    TEST(URL_Decode, DecodePlusAsSpace)
+    {
+        const TCHAR source[] = "hello+world";
+        TCHAR dest[64] = {};
+
+        uint16_t len = ::Thunder::Core::URL::Decode(source, static_cast<uint16_t>(strlen(source)), dest, sizeof(dest));
```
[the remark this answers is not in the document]
bramoosterhuis: Same as line 544: Copilot's suggestion is close but incomplete, see my reply
there for the full fix (sizeof element count and _T() on the literal).

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_STREQ(decoded.c_str(), "hello world");
+    }
+
+    TEST(URL_Encode, RoundTrip)
+    {
+        const TCHAR source[] = "key=hello world&other=a/b";
+        TCHAR encoded[128] = {};
+        TCHAR decoded[128] = {};
+
+        uint16_t encLen = ::Thunder::Core::URL::Encode(source, static_cast<uint16_t>(strlen(source)), encoded, sizeof(encoded));
```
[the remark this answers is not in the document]
bramoosterhuis: Same as line 544: Copilot's suggestion is close but incomplete, see my reply
there for the full fix (sizeof element count and _T() on the literal).

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+    // URL-safe Base64 Encoding / Decoding
+    // =========================================================================
+
+    TEST(URL_Base64, EncodeDecodeRoundTrip)
+    {
+        const uint8_t data[] = { 0x00, 0x01, 0x02, 0xFF, 0xFE, 0x80, 0x7F };
+        TCHAR encoded[64] = {};
+        uint8_t decoded[64] = {};
+
+        uint16_t encLen = ::Thunder::Core::URL::Base64Encode(data, sizeof(data), encoded, sizeof(encoded));
```
bramoosterhuis: Same TCHAR destination issue as the Encode/Decode call sites, see my reply on
line 544. encoded is TCHAR[], so sizeof counts bytes while destinationLength
counts elements. The Base64Decode calls are fine, their destination is
uint8_t[] where the two coincide.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_EQ(memcmp(data, decoded, sizeof(data)), 0);
+    }
+
+    TEST(URL_Base64, URLSafeCharsUsed)
+    {
+        // Use data that would produce + and / in standard Base64
+        const uint8_t data[] = { 0xFB, 0xFF, 0xFE };
+        TCHAR encoded[64] = {};
+
+        uint16_t encLen = ::Thunder::Core::URL::Base64Encode(data, sizeof(data), encoded, sizeof(encoded));
```
bramoosterhuis: Same TCHAR destination issue as the Encode/Decode call sites, see my reply on
line 544. encoded is TCHAR[], so sizeof counts bytes while destinationLength
counts elements. The Base64Decode calls are fine, their destination is
uint8_t[] where the two coincide.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+
+        // URL-safe base64 uses '-' and '_' instead of '+' and '/'
+        EXPECT_EQ(encodedStr.find('+'), string::npos);
+        EXPECT_EQ(encodedStr.find('/'), string::npos);
+    }
+
+    TEST(URL_Base64, EmptyInput)
+    {
+        TCHAR encoded[64] = {};
+        uint16_t encLen = ::Thunder::Core::URL::Base64Encode(nullptr, 0, encoded, sizeof(encoded));
```
bramoosterhuis: Same TCHAR destination issue as the Encode/Decode call sites, see my reply on
line 544. encoded is TCHAR[], so sizeof counts bytes while destinationLength
counts elements. The Base64Decode calls are fine, their destination is
uint8_t[] where the two coincide.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_EQ(encLen, 0u);
+    }
+
+    TEST(URL_Base64, WithPadding)
+    {
+        const uint8_t data[] = { 0x41 }; // 1 byte -> 2 base64 chars + padding
+        TCHAR encoded[64] = {};
+        uint8_t decoded[64] = {};
+
+        uint16_t encLen = ::Thunder::Core::URL::Base64Encode(data, sizeof(data), encoded, sizeof(encoded), true);
```
bramoosterhuis: Same TCHAR destination issue as the Encode/Decode call sites, see my reply on
line 544. encoded is TCHAR[], so sizeof counts bytes while destinationLength
counts elements. The Base64Decode calls are fine, their destination is
uint8_t[] where the two coincide.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        // 32 bytes (like a SHA256 hash)
+        uint8_t data[32];
+        for (int i = 0; i < 32; i++) {
+            data[i] = static_cast<uint8_t>(i * 7 + 3);
+        }
+
+        TCHAR encoded[128] = {};
+        uint8_t decoded[64] = {};
+
+        uint16_t encLen = ::Thunder::Core::URL::Base64Encode(data, sizeof(data), encoded, sizeof(encoded));
```
bramoosterhuis: Same TCHAR destination issue as the Encode/Decode call sites, see my reply on
line 544. encoded is TCHAR[], so sizeof counts bytes while destinationLength
counts elements. The Base64Decode calls are fine, their destination is
uint8_t[] where the two coincide.

Tests/unit/core/test_url.cpp
```diff
@@ -0,0 +1,814 @@
...
+        EXPECT_TRUE(url.IsValid());
+        EXPECT_TRUE(url.Path().IsSet());
+        EXPECT_STREQ(url.Path().Value().c_str(), "a/b/c/d/e");
+    }
+
+    // =========================================================================
+    // Text() Serialization / Round-trip
+    // =========================================================================
+
+    TEST(URL_Text, HTTPRoundTrip)
```
bramoosterhuis, line 308: .Value() is called on OptionalType without checking IsSet() first. If the
reparse fails, which is exactly what this test should catch, it crashes
instead of failing with a usable message.

Tests/unit/core/test_websocket_protocol.cpp
bramoosterhuis, line 1: Best-structured file in the PR: real content assertions on the echo
round-trip, and UniqueSocketPath() with pid plus an atomic counter avoids the
collisions the other test files can hit. Teardown order is right too, join()
before Singleton::Dispose().

ASSERT_EQ(server.Open(...)) inside the server thread is the one thing I'd
change. On failure it returns from the lambda before serverReady is set and
before notify_one(), so the main thread waits the full 10 seconds and then
reports "Server did not become ready in time", hiding the actual cause. With
seven tests on this helper that's up to 70 seconds of CI time for a failure
that was known immediately. EXPECT_EQ plus an explicit failure signal would
surface the real error and skip the wait.

CloseIsClean only asserts IsOpen(); the Close() assertion lives in the helper
and runs for all seven tests, so this one adds nothing beyond its name.

<functional> is missing for std::function in RunTextServer's signature (per
Copilot's comment), currently relying on a transitive include.

Tests/unit/core/test_tls.cpp
bramoosterhuis, line 1: Handshake_SelfSigned is the only test here that exercises TLS itself, and it
passes regardless of outcome: success asserts IsOpen(), failure is explicitly
accepted with "the important thing is no crash", and a server that doesn't
start gives GTEST_SKIP. Three paths, one assertion, and which path runs
depends on the environment. A broken TLS implementation would not fail this.

If handshake failure genuinely can't be relied on in CI, that's worth stating
as an open gap rather than encoding as an accepted outcome. If it can, the
else branch should be a failure.

There's also no negative validation case: AcceptAll is right for a self-signed
cert, but nothing tests that a rejecting IValidate actually aborts the
handshake, which is where a security regression would show up.

CertificateStore_AddCert claims "should accept the cert" but has no
assertion, and Certificate_ValidPeriod comments that validTill should be ~365
days out while only asserting IsValid() on both values, so the 365 passed to
GenerateTestCert is never checked.

Port derivation 14000 + (getpid() % 1000) isn't collision-free (pids 1234 and
2234 map to the same port), and a collision results in a skip rather than a
failure. test_websocket_protocol.cpp in this same PR gets this right with pid
plus an atomic counter; worth reusing that approach here. 127.0.0.1 would
also be enough instead of 0.0.0.0.

Tests/unit/Module.cpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2020 Metrological
```
bramoosterhuis: Copyright says 2020 on a new file; other new files in this PR use 2026.

bramoosterhuis (changes requested):
Went through all 13 files, and cross-checked them against the v4.0 gap
analysis report.

There's real work here, and some of it is good. test_message_dispatcher.cpp
and test_websocket_protocol.cpp verify actual content rather than absence of
crashes. UniqueSocketPath() in the latter (pid plus atomic counter) is the
right way to avoid CI collisions and worth reusing elsewhere. The workerpool
Revoke() fix addresses a genuine destroy-while-dispatching race, and the
socketport scoping fix around Singleton::Dispose() is correct and for the
right reason. Gap 7 in the report is exemplary: marked ⚠️ rather than ✅,
explains why thread exhaustion isn't testable, names the existing tests that
do cover the behaviour, and admits the earlier CI hangs. That's the standard
the rest should meet.

Three patterns run through the rest that I'd want addressed before this
lands, because they work against what the PR is for.

Tests that cannot fail. Handshake_SelfSigned treats handshake failure as an
acceptable outcome, so the only test that exercises TLS would not catch a
broken TLS implementation. Communicator_NotificationRegistration has no
assertions at all. Several tests reach GTEST_SKIP on environment conditions
that never fire (IsIPV6Enabled) or that silently disable coverage when they
do (port collisions, socket open failures).

Coverage claims that outrun the tests. test_controller.cpp lists six items,
one of which has no test at all (Controller.configuration) and two of which
describe behaviour that isn't exercised. The report has the same issue and
contradicts itself on it: it defines ✅ as "assertion-tested", marks OOP
crash detection ✅ and Gap 3 RESOLVED, while its own Limitations section
states the tests "do not fork a child process and kill it". Gap 11 claims
the OOP error scenarios "are assertion-tested" where the cited tests contain
no assertions. Marking those 🔶 and promoting the Limitations text into the
status would make the report accurate without touching a line of code, and
would leave a usable record of what's genuinely still open.

Known defects pinned as intended behaviour. IPv4_DefaultMask and
UserInfoWithoutPassword both document a bug in a comment and then assert the
broken output. DefaultMask() is missing the ntohl() that IsMulticast() and
IsLocalInterface() apply a few lines above it, and has class A and C swapped;
the test's big-endian branch encodes correct classful behaviour the code
doesn't implement, so it would fail on a big-endian target while passing on
x86 for the wrong reason (all three addresses end in .1, so the function
returns 24 regardless). My preference is to assert correct behaviour, file an
issue, and mark the test DISABLED_ pointing at it. That way the test
documents the defect and turns green when it's fixed, instead of keeping a
years-old bug green and in plain sight.

Along the way this review surfaced four apparent production defects worth
issues of their own, independent of the tests: NodeId::DefaultMask() (missing
ntohl, A/C swapped), NodeId::Origin()'s IPv6 branch (sets sin6_family to
AF_INET and clears via IPV4Socket.sin_addr, which lands on sin6_flowinfo, so
the address is never zeroed and the result silently degrades to IPv4
0.0.0.0), NodeId::IsLocalInterface() comparing s6_addr[15] to ASCII '1', and
URL parsing dropping the username when there's no ':' before the '@', which
RFC 3986 permits.

Smaller items are in the per-file comments. Two general ones: test_tls.cpp
should be gated in CMake under SECURE_SOCKET rather than by a preprocessor
guard, so a misconfiguration fails loudly instead of registering zero tests
(the report notes this behaviour explicitly on the conditional compilation
table); and hardcoded ports bound to 0.0.0.0 should be 127.0.0.1 with a
collision-free derivation.

Last thing, and this is the one I'd most like an answer to: was a coverage
run done before and after this branch? A delta would make this objective and
settle whether these tests move the numbers or mainly document intent. Happy
to re-review once the assertion-free tests and the pinned-defect tests are
sorted.

Tests/unit/core/test_url.cpp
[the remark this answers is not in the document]
smanes0213: No CMake file, CI workflow, or build script in this repo defines _UNICODE. It's purely a Windows MSVC-era guard. On Linux (the only platform CI runs), every #ifdef _UNICODE block is dead code and #ifndef _UNICODE is always active — so sizeof(dest) vs sizeof(dest)/sizeof(dest[0]) coincide today and this never manifests as a live bug.

Tests/unit/core/test_controller.cpp
smanes0213: "Controller.activate/deactivate — lifecycle transitions": both tests use
NonExistentPlugin_12345 and assert an error. No plugin is activated or
deactivated, so no transition is exercised.


disabled the above tests for time being until we find a better solution to get a plugin activated and deactivated in the tests. Tried bringing in nanoservices but that would again cuase a major time increase in the CI runs.

PR #2171, opened 2026-07-10
Fix shutdown ordering for OOP IPlugin Release

nxtum:
Fix shutdown of plugins where IPlugin itself is hosted by ThunderPlugin OOP
 
The failing release was the host releasing the plugin interface itself
 
ReleaseInterfaces() could terminate _connection before releasing currentIF
ThunderPlugin could call WorkerPool:Stop() from the ServiceAdministrator callback when the final IPlugin object was destroyed
 
That allowed the ThunderPlugin process to close while the host was still waiting for the remote currentIF->Release() to complete...


VeithMetro (commented):
@nxtum I get the idea, and I think it looks good in general, but I am just wondering about one corner case:

Since `_parent. Stop()` is no longer called in `Dispatch()` for a fully OOP plugins, what if the connection was no longer valid for some reason, and then we would not get to `_connection->Terminate()` in `ReleaseInterfaces()`?

nxtum:
> @nxtum I get the idea, and I think it looks good in general, but I am just wondering about one corner case:
> 
> Since `_parent. Stop()` is no longer called in `Dispatch()` for a fully OOP plugins, what if the connection was no longer valid for some reason, and then we would not get to `_connection->Terminate()` in `ReleaseInterfaces()`?

Well, there are 2 diff cleanup paths

1) _parent.Stop() from Dispatch(), which is on the thunder plugin side. It stops the workerpool when the last obj is released

2) _connection->Terminate() on the Thunder side

When Thunder first creates this OOP plugin, it stores the remote connection

https://github.com/rdkcentral/Thunder/blob/master/Source/Thunder/PluginServer.h#L1343-L1354

https://github.com/rdkcentral/Thunder/blob/master/Source/com/Communicator.h#L1359

This returns a ref counted connection obj, which is owned until its released by ReleaseInterfaces

So from my understanding, during the deactive part, when you terminate this _connection at the end, it will still exist as its still a stored connection even if the channel was closed

nxtum:
For a "traditional" OOP plugin, Stop() is still called. Only when you run a plugin that was not designed for OOP you cannot Stop() because the remote object is Iplugin itself, thunder could still be releasing IPlugin proxy. So if you were to stop there, there is a race, which is why you get:

[Error]: Could not remote release the Proxy for Interface [0x30]

VeithMetro (commented):
Let's add an ASSERT like we discussed, and please do a smoke test to see it does not fire 😄

nxtum:
added an ASSERT that would fire if: 
- plugin is OOP
- there is a plugin interface to release
- but there is no stored connection

VeithMetro (approved):

VeithMetro:
And just to make sure, you did smoke test with this ASSERT and it does not fire? 😄

nxtum:
> And just to make sure, you did smoke test with this ASSERT and it does not fire? 😄

We will find out in production 😄 

I tested spamming activate/deactivate on multiple in process and oop plugins and didn't fire, so should be good to go

PR #2177, opened 2026-07-22
[CMake] Prevent plugin libraries from being retained after dlclose

VeithMetro:
This is a fix for an issue first observed in [this](https://jira.rdkcentral.com/jira/browse/METROL-1073) Jira ticket.

GCC emits `STB_GNU_UNIQUE` symbols for some header-defined static template members. On glibc, this causes plugin .so to remain mapped after dlclose().

Propagate `-fno-gnu-unique` from the Core target to unloadable consumers, allowing deactivated plugin libraries to be fully unmapped. This applies generically to all affected static template members, not only warning-reporting code mentioned in the ticket.

### DIFF
--- a/Source/core/CMakeLists.txt
+++ b/Source/core/CMakeLists.txt
@@ -159,6 +159,13 @@ set(PUBLIC_HEADERS
 
 target_compile_options (${TARGET} PRIVATE -Wno-psabi)
 
+# GNU unique symbols make glibc retain a DSO after dlclose(). Header-defined
+# static template members can be instantiated by plugin DSOs, so this option
+# must propagate to every Core consumer that can be unloaded.
+target_compile_options(${TARGET} INTERFACE
+    $<$<CXX_COMPILER_ID:GNU>:-fno-gnu-unique>
+    )
+
 # non-buildable interface target to carry the definitions
 # All build flags that are applicable to the whole system should be set here in the 
 # core since the core is included everywhere so if they are set here, they are 


sebaszm:
Please note that this will have impact on memory usage (increased by duplicate data structures and decreased by actually allowing to unload modules) and startup performace (increased by no need to lookup symbols, and perhaps decreased by cache needing to track multiple same objects). This is most likely beneficial but I think this needs some performance tests.

bramoosterhuis:
I spent some time reproducing this and tracing where the NODELETE actually comes from, because @sebaszm's point about the wider impact worried me too.

With a breakpoint on glibc's `do_lookup_unique()`, the only symbol that acquires NODELETE on a plugin DSO is `Core::CallsignTLS::CallsignAccess<&Plugin_X>::_moduleName` and its guard variable, one pair per plugin. The `WarningReportingBoundsCategory` bounds are emitted as `STB_GNU_UNIQUE` as well, but they never resolve through that path, so they pin nothing. Their presence in the dynamic symbol table is harmless.

The subtle part is that `_moduleName` is already marked `EXTERNAL_HIDDEN` at its definition in `CallsignTLS.h`, and that attribute does nothing. `CallsignTLS` itself is marked `EXTERNAL`, which applies default visibility to the whole class, and that takes precedence over an attribute on the out of class definition of a nested member. GCC accepts it without a diagnostic. Placing `EXTERNAL_HIDDEN` on the nested template itself does take effect. I confirmed the mechanism with a standalone reproducer: no attribute gives a unique symbol, the attribute on the definition still gives one, the attribute on the nested template does not.

On my build (Fedora 44, x86-64, gcc 16.1, glibc 2.43) that one word takes `/proc/<pid>/maps` from 12 plugin mappings to 0 with every plugin set to Deactivated.

If that holds up on other targets it should also settle @sebaszm's question: with a single symbol addressed there is no duplication of vague linkage statics across DSOs, so there is nothing to measure. Happy to be wrong about that, but I would rather rule it out before we take the flag on the INTERFACE of Core, where it lands on every consumer including code where cross DSO unique binding is intentional.

```diff
--- a/Source/core/CallsignTLS.h
+++ b/Source/core/CallsignTLS.h
@@ -71,7 +71,7 @@ namespace Core {
         };
 
         template <const TCHAR** MODULENAME>
-        struct CallsignAccess {
+        struct EXTERNAL_HIDDEN CallsignAccess {
             static const TCHAR* Callsign() {
                 if (_moduleName.empty()) {
```

The full patch is on [`development/callsign-unique-symbol`](https://github.com/rdkcentral/Thunder/tree/development/callsign-unique-symbol), commit [`fe36e27`](https://github.com/rdkcentral/Thunder/commit/fe36e270681d023b4434f874d01ac17c88966e36). Cherry-pick it if you want it in this PR, or let me know if you'd rather I open a separate one.

bramoosterhuis (commented):
See my findings in my previous comment.

VeithMetro:
@bramoosterhuis That's why I wanted you to review it, I knew that our CMake guru would find a better way to solve that 😄

I fully agree that this is the correct way to approach it, and I also prefer the targeted fix over the broad `-fno-gnu-unique` propagation, as it limits the change to the actual `NODELETE` trigger.

Feel free to make a PR from your branch, and I'll close this one

PR #2185, opened 2026-07-24
[Core] Add ToDecString/FromDecString serialization helpers

sebaszm:
to facilitate JsonGenerator @encode:ip

### DIFF
--- a/Source/core/Serialization.cpp
+++ b/Source/core/Serialization.cpp
@@ -148,7 +148,7 @@ POP_WARNING()
 
     static const TCHAR hex_chars[] = "0123456789abcdef";
 
-    void EXTERNAL ToHexString(const uint8_t object[], const uint32_t length, string& result, const TCHAR delimiter)
+    void ToHexString(const uint8_t object[], const uint32_t length, string& result, const TCHAR delimiter)
     {
         ASSERT(object != nullptr);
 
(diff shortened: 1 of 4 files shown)


Source/core/Serialization.cpp
```diff
@@ -220,6 +220,137 @@ POP_WARNING()
         return (FromHexString(hexString, object.data(), static_cast<uint32_t>(object.size()), delimiter));
     }
 
+    void ToDecString(const uint8_t object[], const uint32_t length, string& result, const TCHAR delimiter)
+    {
+        ASSERT(object != nullptr);
+        ASSERT(delimiter != '\0');
```
VeithMetro, line 227: Should we make it idiot-proof and add an ASSERT if someone passes a decimal digit as an delimiter?

Source/core/Serialization.cpp
```diff
@@ -220,6 +220,137 @@ POP_WARNING()
...
+                }
+            }
+            else if (c == delimiter) {
+                if ((haveValue == true) && (count < maxLength)) {
+                    object[count++] = static_cast<uint8_t>(current);
+                    current = 0;
+                    haveValue = false;
+                }
+                else {
+                    failure = true;
```
VeithMetro, line 301: In `FromHexString`, we still return the number of values that fit even when `maxLength` is reached, shouldn't it be like this here too?

sebaszm: with broken heart, I agree...

Source/core/Serialization.cpp
sebaszm: I guess there are more characters that make no sense as a delimiter, but only nul breaks the function.

VeithMetro (dismissed):

VeithMetro (approved):

PR #2186, opened 2026-07-26
Add JSON::ArrayType Insert and Remove functions with tests and API documentation updates

SatheeshkumarG15:
This PR adds index-based mutation support to JSON::ArrayType by introducing Insert(index, element) and Remove(index) while preserving existing Add(), operator[], Get(), Length(), and Clear() behavior.

- Unit tests for insert/remove at head, middle, and tail, plus mixed mutation and serialization scenarios.
- Documentation updates in json.md covering:
- behavior and complexity (Insert/Remove are (O(n)), append remains (O(1)))
- index bounds and thread-safety expectations
- examples for front/middle/end insert and remove
- clarification that serialization includes only elements with IsSet() == true, with guidance for wrapper-type insertion.

### DIFF
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -3345,55 +3345,48 @@ namespace Core {
             {
                 ASSERT(index <= Length());
 
-                if (index == Length()) {
-                    _data.push_back(ELEMENT());
-                    return (_data.back());
-                }
-                else {
-                    uint32_t skip = index;
-                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+                typename std::list<ELEMENT>::iterator locator = _data.begin();
+                uint32_t skip = index;
 
-                    while (skip != 0) {
-                        locator++;
-                        skip--;
-                    }
-
-                    ASSERT(locator != _data.end());
-                    return (*(_data.insert(locator, ELEMENT())));
+                while (skip != 0) {
+                    ++locator;
+                    --skip;
                 }
+
+                locator = _data.insert(locator, ELEMENT());
+
+                return (*locator);
+
             }
 
             inline ELEMENT& Insert(const uint32_t index, const ELEMENT& element)
             {
                 ASSERT(index <= Length());
 
-                if (index == Length()) {
-                    _data.push_back(element);
-                    return (_data.back());
-                } else {
-                    uint32_t skip = index;
-                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+                typename std::list<ELEMENT>::iterator locator = _data.begin();
+                uint32_t skip = index;
 
-                    while (skip != 0) {
-                        locator++;
-                        skip--;
-                    }
-
-                    ASSERT(locator != _data.end());
-                    return (*(_data.insert(locator, element)));
+                while (skip != 0) {
+                    ++locator;
+                    --skip;
                 }
+
+                locator = _data.insert(locator, element);
+
+                return (*locator);
+
             }
 
             inline ELEMENT* Remove(const uint32_t index)
             {
                 ASSERT(index < Length());
 
-                uint32_t skip = index;
                 typename std::list<ELEMENT>::iterator locator = _data.begin();
+                uint32_t skip = index;
 
                 while (skip != 0) {
-                    locator++;
-                    skip--;
+                    ++locator;
+                    --skip;
                 }
 
                 ASSERT(locator != _data.end());
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -3377,6 +3377,23 @@ namespace Core {
 
             }
 
+            inline ELEMENT& Insert(const uint32_t index, ELEMENT&& element)
+            {
+                ASSERT(index <= Length());
+
+                typename std::list<ELEMENT>::iterator locator = _data.begin();
+                uint32_t skip = index;
+
+                while (skip != 0) {
+                    ++locator;
+                    --skip;
+                }
+
+                locator = _data.insert(locator, std::move(element));
+
+                return (*locator);
+            }
+
             inline ELEMENT* Remove(const uint32_t index)
             {
                 ASSERT(index < Length());
--- a/Source/core/JSON.h
+++ b/Source/core/JSON.h
@@ -3341,6 +3341,43 @@ namespace Core {
                 return (_data.back());
             }
 
+            void Insert(uint32_t index, const ELEMENT& element)
+            {
+                ASSERT(index <= Length());
+
+                if (index == Length()) {
+                    Add(element);
+                }
+                else {
+                    uint32_t skip = index;
+                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+
+                    while (skip != 0) {
+                        locator++;
+                        skip--;
+                    }
+
+                    ASSERT(locator != _data.end());
+                    _data.insert(locator, element);
+                }
+            }
+
+            void Remove(uint32_t index)
+            {
+                ASSERT(index < Length());
+
+                uint32_t skip = index;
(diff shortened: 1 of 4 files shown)


docs/utils/json.md
```diff
@@ -96,8 +96,68 @@ The following JSON data type representations are available:
...
+
+// Insert in middle:
+values.Insert(2, c);
+// values == ["A", "B", "C", "D"]
+
+// Insert at end (index == Length()):
+values.Insert(values.Length(), e);
+// values == ["A", "B", "C", "D", "E"]
+```, "E"]
+```
```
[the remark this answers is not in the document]
SatheeshkumarG15: Valid and addressed in 9edad81332cab16f3927822a846d010ab9a1e215

Tests/unit/core/test_jsonarray.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    using JsonArray = ::Thunder::Core::JSON::ArrayType<::Thunder::Core::JSON::String>;
+
+    TEST(JSONARRAY, InsertAtHead)
+    {
+        JsonArray array;
+
+        array.Add() = _T("middle");
+        array.Add() = _T("tail");
+        array.Insert(0, ::Thunder::Core::JSON::String(_T("head")));
```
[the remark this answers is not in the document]
SatheeshkumarG15: Valid and addressed in 9edad81332cab16f3927822a846d010ab9a1e215

Tests/unit/core/test_jsonarray.cpp
```diff
@@ -0,0 +1,251 @@
...
+        JsonArray array;
+        std::string serialized;
+
+        array.Add() = _T("first");
+        array.Add() = _T("second");
+        array.Add() = _T("third");
+
+        ::Thunder::Core::JSON::String moved = array[0];
+        array.Remove(0);
+        array.Insert(1,moved);
```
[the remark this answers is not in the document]
SatheeshkumarG15: Valid and addressed in 9edad81332cab16f3927822a846d010ab9a1e215

Source/core/JSON.h
```diff
@@ -3341,6 +3341,43 @@ namespace Core {
                 return (_data.back());
             }
 
+            void Insert(uint32_t index, const ELEMENT& element)
```
sramani-metro: There are 2 different Add(), 
1. one which adds an empty element and returns the reference.
2. One which adds the passed in Element at the back of the array and returns the reference of the Element.

I think Insert should also follow similar pattern.

Source/core/JSON.h
```diff
@@ -3341,6 +3341,43 @@ namespace Core {
...
+                        locator++;
+                        skip--;
+                    }
+
+                    ASSERT(locator != _data.end());
+                    _data.insert(locator, element);
+                }
+            }
+
+            void Remove(uint32_t index)
```
sramani-metro: For remove, I am thinking should we return the reference of the next item after the one which is erased?

sramani-metro: Also, please make sure use of const appropriately.

sramani-metro (changes requested):

Source/core/JSON.h
SatheeshkumarG15: Overloaded `Insert()` similar to `Add()` in 36278b82f4523c31785cd70fe4abdb489b6cf39a. Please review the PR.

Source/core/JSON.h
SatheeshkumarG15: Modified `Remove()' to return the reference of the next item. If its the last element we return nullptr.

SatheeshkumarG15: Made the parameter as const.

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
                 return (_data.back());
             }
 
+            inline ELEMENT& Insert(const uint32_t index)
+            {
+                ASSERT(index <= Length());
+
+                if (index == Length()) {
+                    _data.push_back(ELEMENT());
+                    return (_data.back());
```
MFransen69: In principle in Thunder coding guidelines we only have one return per method.
I can see why that is inconvenient here. You could use an iterator (think the optimizer might make it as efficient ) but would be a nice test. Or perhaps use an pointer (with an ASSERT to check it is not nullptr) before returning?

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
...
+                }
+                else {
+                    uint32_t skip = index;
+                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+
+                    while (skip != 0) {
+                        locator++;
+                        skip--;
```
MFransen69: --skip is a (very little :) ) more performant, as well as ++locator where the difference is a little bigger

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
...
+                if (index == Length()) {
+                    _data.push_back(ELEMENT());
+                    return (_data.back());
+                }
+                else {
+                    uint32_t skip = index;
+                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+
+                    while (skip != 0) {
+                        locator++;
```
MFransen69: doing ++ on an iterator that is already on end is undefined behaviour so the assert below might now work correclty. Of course there is already an assert in the beginning so it should never happen. 
Perhaps move the assert inside the loop (I can't care about the small perfomance impact in debug), when assigning skip make sure it cannot be bigger or perhaps just remove the assert as it is already covered with the assert in the beginning?

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
...
+                ASSERT(index <= Length());
+
+                if (index == Length()) {
+                    _data.push_back(element);
+                    return (_data.back());
+                } else {
+                    uint32_t skip = index;
+                    typename std::list<ELEMENT>::iterator locator = _data.begin();
+
+                    while (skip != 0) {
```
MFransen69: see previous method remarks

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
...
+                        locator++;
+                        skip--;
+                    }
+
+                    ASSERT(locator != _data.end());
+                    return (*(_data.insert(locator, ELEMENT())));
+                }
+            }
+
+            inline ELEMENT& Insert(const uint32_t index, const ELEMENT& element)
```
MFransen69, line 3362: Don't know if we have rvalue versions (move capable) already somewhere else otherwise we should add these as well.

Source/core/JSON.h
```diff
@@ -3341,6 +3341,67 @@ namespace Core {
...
+            }
+
+            inline ELEMENT* Remove(const uint32_t index)
+            {
+                ASSERT(index < Length());
+
+                uint32_t skip = index;
+                typename std::list<ELEMENT>::iterator locator = _data.begin();
+
+                while (skip != 0) {
```
MFransen69, line 3404: see first method remarks

MFransen69 (changes requested):

Source/core/JSON.h
SatheeshkumarG15: Overloaded Insert() to accept rvalue versions in 90379685ca088f3749205bbecb8df188135272d2

Source/core/JSON.h
SatheeshkumarG15: Addressed in 9564fa55b8b531fb1ed65fef1bdef3cc99a25b1e

MFransen69 (approved):

sramani-metro (approved):

PR #2189, opened 2026-07-27
Migrate Thunder JSON-RPC and COM-RPC propagation to traceparent model

tabbas651:
This change moves Thunder distributed tracing to a traceparent-first propagation model across JSON-RPC and COM-RPC paths. It removes dependency on shared-memory key lookup for Thunder RPC parent-child correlation and aligns runtime behavior with the current design.

**Changes**
**1. JSON-RPC invoke path (PluginServer.h, DistributedTracing.cpp)**
Added traceparent-aware invoke begin/end flow gated behind #ifdef THUNDER_DISTRIBUTED_TRACING.
OnInvokeBegin() reads the current traceparent from rdk_otlp TLS and starts a child span only when one is active; returns bool to track span state.
OnInvokeEnd() finishes the child span and records result code, skipped safely when no span was started.

**2. COM-RPC acquire path (PluginServer.h, DistributedTracing.cpp)**
Migrated acquire tracing to traceparent-based child creation via OnCOMRPCAcquireBegin()/OnCOMRPCAcquireEnd().
Removed legacy shared-memory context-key lookup from acquire begin flow.
Returns bool for consistent span lifecycle tracking.

**3. COM-RPC message header extension (Messages.h)**
Extended Data::Input header to carry a 56-byte traceparent field at bytes [21..76].
Added SetTraceParent() and TraceParent() accessors with bounds checks to prevent out-of-bounds access on legacy frames.
Added LEGACY_HEADER_SIZE (13 bytes) constant; IsValid() validates against legacy size for backward compatibility.
Writer()/Reader() use dynamic header offset — full header (77) when present, legacy header (13) otherwise — to avoid corrupting payload decoding on legacy frames.

**4. COM-RPC proxy side (IUnknown.cpp)**
Stamps traceparent into COM-RPC message header on the sender side via SetTraceParent(), gated behind #ifdef THUNDER_DISTRIBUTED_TRACING.

**5. COM-RPC stub side (Administrator.h, Administrator.cpp)**
Added ICOMRPCStubTraceCallbacks struct with onBegin/onEnd function pointers for stub-side trace callbacks.
g_stubTraceCallbacks is std::atomic with memory_order_acquire/memory_order_release to prevent data races during init/shutdown.
Invoke() loads the callback pointer once into a local variable and checks both function pointers before calling, ensuring no null dereference and consistent begin/end pairing.

**6. Tracing API (DistributedTracing.h, DistributedTracing.cpp)**
All tracing methods return bool (not uint64_t span IDs) — lightweight span-active tracking.
No parentSpanId parameter; traceparent propagation is handled entirely by the rdk_otlp library.
OnCOMRPCStubBegin() receives the traceparent extracted from the COM-RPC message header.
Single _enabled member (std::atomic<bool>); no SpanMetadata or duration tracking.

**7. Build system (CMakeLists.txt)**
DISTRIBUTED_TRACING CMake option gates find_library(rdk_otlp), source compilation, and THUNDER_DISTRIBUTED_TRACING compile definition.
Fatal error message aligned with actual library search term (librdk_otlp).

**Impact**
No ThunderTools generator changes required.
No plugin method signature changes required.
Backward-compatible: legacy COM-RPC frames (13-byte header) are accepted and decoded correctly; traceparent accessors return empty when the field is absent.
All new code is gated behind #ifdef THUNDER_DISTRIBUTED_TRACING; builds without the flag are unaffected.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -49,6 +49,8 @@ option(DISABLE_USE_COMPLEMENTARY_CODE_SET
         "Disable the complementary code set" OFF)
 option(ENABLE_CONFIG_OVERRIDE
         "Enable persistent configuration overrides." OFF)
+option(DISTRIBUTED_TRACING
+    "Enable distributed tracing via rdk_otlp_instrumentation" OFF)
 
 if(HIDE_NON_EXTERNAL_SYMBOLS)
     set(CMAKE_CXX_VISIBILITY_PRESET hidden)
--- a/Source/WPEFramework/CMakeLists.txt
+++ b/Source/WPEFramework/CMakeLists.txt
@@ -78,7 +78,7 @@ endif()
 if(DISTRIBUTED_TRACING)
     find_library(RDK_OTEL_LIB rdk_otlp)
     if(NOT RDK_OTEL_LIB)
-        message(FATAL_ERROR "DISTRIBUTED_TRACING enabled but librdk_otel_instrumentation not found")
+        message(FATAL_ERROR "DISTRIBUTED_TRACING enabled but librdk_otlp not found")
     endif()
 
     target_sources(${TARGET} PRIVATE DistributedTracing.cpp)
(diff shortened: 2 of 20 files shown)


AravindanNC:
Closing this PR, the last set of working changes are captured in https://github.com/rdkcentral/Thunder/pull/2195

PR #2216, opened 2026-09-01
Runtime Configurable Attributes

nxtum:


Source/Thunder/Controller.cpp
```diff
@@ -361,6 +361,79 @@ namespace Plugin {
...
+        Core::hresult result = Core::ERROR_UNKNOWN_KEY;
+
+        ASSERT(_pluginServer != nullptr);
+
+        const PluginHost::Config& configuration(_pluginServer->Configuration());
+
+        if (attribute.IsSet() == false) {
+            Core::JSON::VariantContainer attributes;
+            attributes.Set(_T("prefix"), Core::JSON::Variant(configuration.Prefix()));
+            attributes.Set(_T("idletime"), Core::JSON::Variant(static_cast<uint32_t>(configuration.IdleTime())));
```
sebaszm: use EnumToString to ensure the enum string is always the same as in the interface

sebaszm (changes requested):

Source/Thunder/Controller.cpp
```diff
@@ -361,6 +361,79 @@ namespace Plugin {
...
+
+        ASSERT(_pluginServer != nullptr);
+
+        const PluginHost::Config& configuration(_pluginServer->Configuration());
+
+        if (attribute.IsSet() == false) {
+            Core::JSON::VariantContainer attributes;
+            attributes.Set(Core::EnumToCString(IConfiguration::PREFIX), Core::JSON::Variant(configuration.Prefix()));
+            attributes.Set(Core::EnumToCString(IConfiguration::IDLETIME), Core::JSON::Variant(static_cast<uint32_t>(configuration.IdleTime())));
+            attributes.ToString(value);
```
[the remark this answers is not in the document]
nxtum: this is not true at all, it uses the correct one from the text tag

Source/plugins/IController.h
```diff
@@ -86,6 +86,11 @@ namespace Controller {
     struct EXTERNAL IConfiguration : virtual public Core::IUnknown {
         enum { ID = RPC::ID_CONTROLLER_CONFIGURATION };
 
+        enum attribute : uint8_t {
+            PREFIX /* @text:prefix */,
+            IDLETIME /* @text:idletime */
+        };
```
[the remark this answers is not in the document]
sebaszm: this is not a concern

Source/Thunder/Config.h
```diff
@@ -927,12 +935,17 @@ namespace PluginHost {
...
         {
             Core::SafeSyncType<Core::CriticalSection> scopedLock(_configLock);
-            return (_prefix);
+            return (_attributes.Prefix);
         }
         inline void SetPrefix(const string& newValue) {
             Core::SafeSyncType<Core::CriticalSection> scopedLock(_configLock);
-            _prefix = newValue;
-            _webPrefix = '/' + _prefix;
+            _attributes.Prefix = newValue;
```
sebaszm, line 942: Prefix and idletime cannot be applied at runtime, so on set you always set pending and on get you always get the original. Setting the value different than original means request_submitted, otherwise error_none. On store you always store pending. On Load you populate the original and pending. This way the logic is simpler, you don't need the SetPendingX methods either.

Source/plugins/IController.h
```diff
@@ -105,6 +110,18 @@ namespace Controller {
...
         virtual Core::hresult Configuration(const string& callsign /* @index */, const string& configuration /* @opaque */) = 0;
+
+        // @property
+        // @brief Framework attributes
+        // @details Changes are not persisted automatically. Call Persist("PluginHost") to retain them.
+        // @param attribute Attribute name (omit to retrieve all)
+        // @retval ERROR_NONE Attribute was retrieved or is unchanged
+        // @retval ERROR_UNKNOWN_KEY Attribute does not exist
+        // @retval ERROR_BAD_REQUEST Attribute value is invalid
+        // @retval ERROR_UNSUPPORTED Attribute cannot be changed in current framework configuration
```
sebaszm, line 121: I think it's ERROR_NOT_SUPPORTED

sebaszm (changes requested):

sebaszm:
> ### 🔵 Needs a closer look
> Correct all-config persistence/restoration handling and regenerate the Controller documentation.
> 
> Review details
> ### Suppressed comments (3)
> **Source/Thunder/PluginServer.h:2009**
> 
> * This changes the existing no-callsign behavior: `Persist()` used to include the PluginHost override in the save-all path, but this condition now skips `SavePluginHostConfig()`. As a result, runtime `Attribute` changes are silently not persisted when callers use the documented all-config operation; retain `saveAll` here (or explicitly revise the API contract and its callers).
> 
> ```
>                 if (isPluginHost == true) {
> ```
> 
> **Source/Thunder/PluginServer.h:2056**
> 
> * `Restore()` without a callsign is documented as restoring all configuration, and previously removed the PluginHost override as part of `destroyAll`. With this condition, `PluginHost.json` is left behind, so a later restart re-applies the framework attributes instead of restoring their defaults. Include `destroyAll` in this condition as before.
> 
> ```
>                 if (isPluginHost == true) {
> ```
> 
> **Source/plugins/IController.h:119**
> 
> * The checked-in generated `Source/Thunder/doc/ControllerPlugin.md` still omits this new `Attribute` API (including the Controller Configuration method list and `exists` documentation), so the public Controller contract is undocumented after this change. Please regenerate and commit the Controller plugin documentation with the new property and its return/value semantics.
> 
> ```
>         // @property
>         // @brief Framework attributes
>         // @details Changes are not persisted automatically. Call Persist("PluginHost") to retain them.
>         // @param attribute Attribute name (omit to retrieve all)
>         // @retval ERROR_NONE Attribute was retrieved or matches the active value
>         // @retval ERROR_UNKNOWN_KEY Attribute does not exist
> ```
> 
> * **Files reviewed:** 6/6 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite

This is a concious decision not to inlcude Thunder configuration if the param is omitted (it is not a plugin).

sebaszm (approved):

